PR #77, opened 2026-02-11
feat: add RTL support for ordered/unordered lists and blockquote

y0u-0:
### What/Why?
Adds RTL (right-to-left) support for ordered lists, unordered lists, and blockquotes across both native renderers
  and HTML generation.

 Native rendering (Android):
  - Uses the x and dir parameters from drawLeadingMargin to correctly position list markers and blockquote borders in
   both LTR and RTL layouts
  - Fixes drawRect to use minOf/maxOf so border rects are valid regardless of direction

 Native rendering (iOS):
  - Detects app-level layout direction via UIApplication.sharedApplication.userInterfaceLayoutDirection
  - Mirrors blockquote borders and list markers to the trailing edge in RTL

 HTML generation (both platforms):
  - Replaces physical CSS properties (border-left, padding-left, border-radius) with logical equivalents
  (border-inline-start, padding-inline-start, border-start-end-radius, border-end-end-radius)

### Testing
  - Tested on iOS and Android with RTL locale
  - Verified ordered lists, unordered lists, and blockquotes (including nested) render markers/borders on the correct
   side
  - Verified LTR rendering is unchanged

 #### Screenshots

  **iOS — Before**

   

  **iOS — After**

   

  **Android — Before**

   

  **Android — After**

   

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/spans/TaskListSpan.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/spans/TaskListSpan.kt
@@ -71,7 +71,7 @@ class TaskListSpan(
     val fontMetrics = paint.fontMetrics
     val capHeight = -fontMetrics.ascent * CAP_HEIGHT_RATIO
     val centerY = baseline - capHeight / HALF_DIVISOR
-    val centerX = (depth * marginLeft + checkboxSize / HALF_DIVISOR) * dir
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (commented):
Hi @y0u-0 , can you please solve the conflicts for you PR?

PR #115, opened 2026-02-27
fix(android): preserve nested renderer block style context

longseespace:
### What/Why?

**Problem:**
- Certain complex markdown structures on Android (notably nested blockquotes and list items containing blockquotes followed by sibling items) could render as a blank view.

**Root cause:**
- Nested block renderers were calling clearBlockStyle() in finally blocks, which discarded parent block style state.
- Subsequent inline rendering could run without a valid block style and fail, triggering fallback paths that clear rendered output.

**Solution:**
- Added BlockStyleContext snapshot/restore APIs to preserve and reinstate parent block style state.
- Updated BlockquoteRenderer, HeadingRenderer, and CodeBlockRenderer to restore previous block style in finally blocks instead of clearing global state.

**Notes/Caveats:**
- Change is Android renderer-context only; no parser/AST behavior was modified.
- Semantics for top-level style reset remain unchanged (clearBlockStyle still exists for true resets).

### Testing
I added a new sample markdown document with nested blockquotes etc. Before this fix it was blank. After the fix it renders correctly.

| Before | After |
| - | - |
|  |  |

### PR Checklist


longseespace:
Ah sorry I forgot the changes I made to the example. You can find the markdown here https://github.com/software-mansion-labs/react-native-enriched-markdown/pull/115/changes/28cd283c1367218fe86f635d07c4f9828e1e971d#diff-98e881f756c02d7d5dcb83a0c311dfd4864e5f6d956eb1e1894fcaa729e06b14

hryhoriiK97 (approved):

PR #151, opened 2026-03-12
fix: unwrap consecutive latex/math blocks merged into a single paragraph by md4c

hryhoriiK97:
### What/Why?
Consecutive `$$...$$` blocks on adjacent lines (no blank line between them) were rendered as plain text instead of math. `md4c` merges them into a single paragraph with LineBreak nodes between them, and the existing unwrap logic only handled the single-block case. Extended it to unwrap all display math blocks from such paragraphs.

#### Screenshots

| Before | After |
| - | - |
|  |  |

### PR Checklist

### DIFF
--- a/cpp/parser/MD4CParser.cpp
+++ b/cpp/parser/MD4CParser.cpp
@@ -395,16 +395,43 @@ std::shared_ptr<MarkdownASTNode> MD4CParser::parse(const std::string &markdown, 
 
   impl_->flushText();
 
-  // md4c emits $$...$$ as an inline span inside a Paragraph — promote to block-level
-  // so the segment splitter treats it as a standalone math segment.
+  // md4c wraps certain block-level constructs as inline spans inside a Paragraph.
+  // When they appear on consecutive lines without a blank separator, md4c merges
+  // them into a single Paragraph with LineBreak / whitespace Text nodes between them.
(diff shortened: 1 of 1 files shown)


ppaanngggg:
hi, I use 0.4.1, this issue seems still exists.

```markdown
Based on the image provided, we can solve the quadratic equation as follows:

### 1. Identify the coefficients
For a quadratic equation in the form $ax^2 + bx + c = 0$, we have:
- $a = 1$
- $b = -3$
- $c = 2$

### 2. The Quadratic Formula
We use the standard formula for finding the roots:
$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}$$

### 3. Substitution & Calculation
Substitute the values into the formula:
$$x = \frac{-(-3) \pm \sqrt{(-3)^2 - 4(1)(2)}}{2(1)}$$
$$x = \frac{3 \pm \sqrt{9 - 8}}{2}$$
$$x = \frac{3 \pm \sqrt{1}}{2}$$

### 4. Final Results
Simplified, we get two possible values for $x$:
- $x_1 = \frac{3 + 1}{2} = 2$
- $x_2 = \frac{3 - 1}{2} = 1$

**Solution set: $\{1, 2\}$**
```

as this example, block latex formulas in 2 and 3 don't show.

PR #226, opened 2026-04-08
feat: add ordered and unordered lists to the input

yahya-lafhal:
### What/Why?
Adds the support for ordered (numbered) and unordered (bullet) lists in the input provided. I followed the logic of the inline styles that are currently implemented. I made `toggle` functions along with an `isActive` property in the StyleState. I also handled list continuation when returning to line. This should work for both iOS and Android. I made sure to include the feature in the `InputScreen.tsx` and added a small description in the documentation.

### Testing
I tested thoroughly on Android. It works as expected. I unfortunately can't test on iOS, so I would appreciate it if someone could try the example app on iOS and try to :
- toggle an unordered list 
- return to line and have the list continue
- toggle again to stop the list
- toggle an ordered list
- return to line and have the list continue (with 2.)
- toggle again to stop the list
- verify that states (color on the buttons) are coherent even when returning the cursor later to a list.

https://github.com/user-attachments/assets/893523b7-70bb-4442-9c31-cd9b04f27cc9

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownInputManager.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownInputManager.kt
@@ -360,6 +360,14 @@ class EnrichedMarkdownInputManager :
     view?.removeLinkAtCursor()
   }
 
+  override fun toggleUnorderedList(view: EnrichedMarkdownInputView?) {
+    view?.toggleUnorderedList()
+  }
+
+  override fun toggleOrderedList(view: EnrichedMarkdownInputView?) {
+    view?.toggleOrderedList()
+  }
+
   override fun requestMarkdown(
     view: EnrichedMarkdownInputView?,
     requestId: Int,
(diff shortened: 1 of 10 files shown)


hryhoriiK97 (commented):
@yahya-lafhal thanks for the PR, but lists are block-level elements and our input formatting pipeline is entirely inline-based (ENRMFormattingStore, ENRMStyleHandler, ENRMInputFormatter, ENRMMarkdownSerializer, ENRMInputParser). This PR bypasses all of those layers and does raw string manipulation directly on the text view, which breaks formatting range tracking and causes state drift on subsequent edits.

Lists are on our radar and we plan to support them, but they require a new block-level formatting abstraction designed across all three layers (JS, iOS, Android). This isn't something that can be patched incrementally on top of the current approach. We'd recommend closing this PR and opening a feature request instead so we can track it properly and design the right solution.

yahya-lafhal:
Got it, that makes sense. Thanks for the detailed explanation!

PR #242, opened 2026-04-13
fix(android): preserve block markdown in InputParser plain text

eduardoborges:
## Summary

`InputParser.parseToPlainTextAndRanges` on Android now reconstructs GFM structure in plain text from the AST: headings (`#`), list markers, blockquotes (`>`), block separation with double newlines, and trims the trailing injected paragraph separator.

## Motivation

Without this, plain text dropped block structure and the rich input diverged from expected markdown (we had been applying the same change via a local patch in our app).

## Notes

- Kotlin only (`InputParser.kt`).
- Android lib tests / ktlint were not run in this environment; you can validate with `./gradlew :react-native-enriched-markdown:ktlintCheck` from the example app if needed.

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
@@ -25,7 +25,8 @@ object InputParser {
     val plainText = StringBuilder()
     val ranges = mutableListOf<FormattingRange>()
 
-    walkNode(ast, plainText, ranges, ArrayDeque())
+    walkNode(ast, plainText, ranges, ArrayDeque(), null)
+    trimTrailingInjectedParagraphSeparator(plainText)
 
     return ParseResult(plainText.toString(), ranges)
   }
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (commented):
Hi @eduardoborges! Thanks for the PR — we appreciate the contribution.

Unfortunately, we won't be able to proceed with this approach. We have full support for block-level markdown elements (headings, lists, blockquotes, etc.) on our roadmap, but the plan is to render them as rich text rather than reconstructing raw markdown syntax in the plain text. So this would conflict with the direction we're heading.

eduardoborges:
> Hi @eduardoborges! Thanks for the PR — we appreciate the contribution.
> 
> Unfortunately, we won't be able to proceed with this approach. We have full support for block-level markdown elements (headings, lists, blockquotes, etc.) on our roadmap, but the plan is to render them as rich text rather than reconstructing raw markdown syntax in the plain text. So this would conflict with the direction we're heading.

Its makes sense! Maybe i can help on that.

hryhoriiK97:
> > Hi @eduardoborges! Thanks for the PR — we appreciate the contribution.
> > Unfortunately, we won't be able to proceed with this approach. We have full support for block-level markdown elements (headings, lists, blockquotes, etc.) on our roadmap, but the plan is to render them as rich text rather than reconstructing raw markdown syntax in the plain text. So this would conflict with the direction we're heading.
> 
> Its makes sense! Maybe i can help on that.

@eduardoborges Actually, I think we'll want to handle this on our own - we should think about support for block elements in EnrichedMarkdownInput. Right now it only supports inline styles, so this will require architectural changes in the current implementation.

PR #248, opened 2026-04-17
feat: add list.markerMinWidth to align UL/OL/task lists

mozharovsky:
### What/Why?

Today each list kind reserves its own natural marker column width: `bulletSize / 2` for unordered, the width of `"99."` at the marker font for ordered, and the checkbox size for tasks. Mixed content therefore looks ragged — bulleted rows sit far to the left of numbered rows, task rows to the left of both.

This PR adds one optional style, `list.markerWidth`, that acts as a **minimum reserved marker column width** applied to all three list kinds. Each list's effective column becomes `max(markerWidth, natural)`, so consumers can line up mixed lists without shrinking ordered lists or resizing the bullet/checkbox glyphs. Values below the natural width are ignored; leaving it unset preserves existing behavior.

```tsx
<EnrichedMarkdownText
  markdownStyle={{
    list: {
      bulletSize: 6,
      markerWidth: 20, // widens UL and task rows to line up with OL at 16pt
    },
  }}
  ...
/>
```

### Implementation

- **Public API**: `ListStyle.markerWidth?: number` (type-only; JSDoc on the internal type captures the sentinel contract).
- **Internal codegen prop**: concrete `Float`; negative = auto. Default is set in `normalizeMarkdownStyle` so consumers never touch the sentinel.
- **iOS**: `StyleConfig` gets a new ivar + getter/setter; `effectiveListMarginLeftForBullet` / `effectiveListMarginLeftForNumber` floor by `markerWidth`; new `effectiveListMarginLeftForTask` handles task rows. Wired through `StylePropsUtils` and included in `MeasurementCache` so cached sizes invalidate when the prop changes. `ListItemRenderer.m` routes the task case through the new accessor.
- **Android**: `ListStyle` carries the field and exposes `effectiveMarkerWidth(natural)`. `UnorderedListSpan`, `OrderedListSpan`, and `TaskListSpan` all apply the floor. Bullets and checkboxes now position themselves relative to the reserved column's right edge (matching iOS), which stays pixel-identical when the column width equals the natural glyph width.
- **Docs**: new row in `docs/STYLES.md`.
- **Example app**: `apps/example/src/markdownStyles.ts` sets `markerWidth: 20` so the feature is visible against the existing mixed-list sample content.

### Testing

- [x] iOS example app built and launched on iPhone 17 Pro simulator.
- [x] Android example app built and launched on Pixel 3a (API 34) emulator.
- [x] Verified before/after visually on the existing `sampleMarkdown` which has mixed UL, OL, and task lists on both platforms.
- [x] Behavior unchanged when `markerWidth` is undefined (floor is below natural).

#### Screenshots

Same mixed-list sample content (UL + OL + task rows from `sampleMarkdown.ts`) rendered with and without `markerWidth: 20`.

**iOS (iPhone 17 Pro simulator)**

| Default | `markerWidth: 20` |
| - | - |
| Default | markerWidth: 20 |

**Android (Pixel 3a API 34 emulator)**

| Default | `markerWidth: 20` |
| - | - |
| Default Android | markerWidth: 20 Android |

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
@@ -12,16 +12,13 @@ data class ListStyle(
   override val lineHeight: Float,
   val bulletColor: Int,
   val bulletSize: Float,
-  val markerWidth: Float,
+  val markerMinWidth: Float,
   val markerColor: Int,
   val markerFontWeight: String,
   val gapWidth: Float,
   val marginLeft: Float,
 ) : BaseBlockStyle {
-
-  /** Floor `naturalWidth` by the consumer-configured `markerWidth`. */
-  fun effectiveMarkerWidth(naturalWidth: Float): Float =
-    if (markerWidth > naturalWidth) markerWidth else naturalWidth
+  fun effectiveMarkerWidth(naturalWidth: Float): Float = naturalWidth.coerceAtLeast(markerMinWidth)
 
   companion object {
     fun fromReadableMap(
@@ -38,13 +35,7 @@ data class ListStyle(
       val lineHeight = parser.toPixelFromSP(lineHeightRaw)
       val bulletColor = parser.parseColor(map, "bulletColor")
       val bulletSize = parser.toPixelFromDIP(map.getDouble("bulletSize").toFloat())
-      val markerWidthRaw = map.getDouble("markerWidth").toFloat()
-      val markerWidth =
-        if (markerWidthRaw < 0f) {
-          -1f
-        } else {
-          parser.toPixelFromDIP(markerWidthRaw)
-        }
+      val markerMinWidth = parser.toPixelFromDIP(map.getDouble("markerMinWidth").toFloat().coerceAtLeast(0f))
       val markerColor = parser.parseColor(map, "markerColor")
       val markerFontWeight = parser.parseString(map, "markerFontWeight", "normal")
       val gapWidth = parser.toPixelFromDIP(map.getDouble("gapWidth").toFloat())
@@ -60,7 +51,7 @@ data class ListStyle(
         lineHeight,
         bulletColor,
         bulletSize,
-        markerWidth,
+        markerMinWidth,
         markerColor,
         markerFontWeight,
         gapWidth,
--- a/android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
@@ -12,11 +12,17 @@ data class ListStyle(
   override val lineHeight: Float,
   val bulletColor: Int,
   val bulletSize: Float,
+  val markerWidth: Float,
   val markerColor: Int,
   val markerFontWeight: String,
   val gapWidth: Float,
   val marginLeft: Float,
 ) : BaseBlockStyle {
+
+  /** Floor `naturalWidth` by the consumer-configured `markerWidth`. */
+  fun effectiveMarkerWidth(naturalWidth: Float): Float =
+    if (markerWidth > naturalWidth) markerWidth else naturalWidth
+
   companion object {
     fun fromReadableMap(
       map: ReadableMap,
@@ -32,6 +38,13 @@ data class ListStyle(
       val lineHeight = parser.toPixelFromSP(lineHeightRaw)
       val bulletColor = parser.parseColor(map, "bulletColor")
       val bulletSize = parser.toPixelFromDIP(map.getDouble("bulletSize").toFloat())
+      val markerWidthRaw = map.getDouble("markerWidth").toFloat()
+      val markerWidth =
+        if (markerWidthRaw < 0f) {
+          -1f
+        } else {
+          parser.toPixelFromDIP(markerWidthRaw)
+        }
       val markerColor = parser.parseColor(map, "markerColor")
       val markerFontWeight = parser.parseString(map, "markerFontWeight", "normal")
       val gapWidth = parser.toPixelFromDIP(map.getDouble("gapWidth").toFloat())
@@ -47,6 +60,7 @@ data class ListStyle(
         lineHeight,
         bulletColor,
         bulletSize,
+        markerWidth,
         markerColor,
         markerFontWeight,
         gapWidth,
--- a/ios/styles/StyleConfig.mm
+++ b/ios/styles/StyleConfig.mm
@@ -151,6 +151,7 @@ static inline NSString *normalizedFontWeight(NSString *fontWeight)
   CGFloat _listStyleLineHeight;
   RCTUIColor *_listStyleBulletColor;
   CGFloat _listStyleBulletSize;
+  CGFloat _listStyleMarkerWidth;
   RCTUIColor *_listStyleMarkerColor;
   NSString *_listStyleMarkerFontWeight;
   CGFloat _listStyleGapWidth;
@@ -434,6 +435,7 @@ static inline NSString *normalizedFontWeight(NSString *fontWeight)
   copy->_listStyleLineHeight = _listStyleLineHeight;
   copy->_listStyleBulletColor = [_listStyleBulletColor copy];
   copy->_listStyleBulletSize = _listStyleBulletSize;
+  copy->_listStyleMarkerWidth = _listStyleMarkerWidth;
   copy->_listStyleMarkerColor = [_listStyleMarkerColor copy];
   copy->_listStyleMarkerFontWeight = [_listStyleMarkerFontWeight copy];
   copy->_listStyleGapWidth = _listStyleGapWidth;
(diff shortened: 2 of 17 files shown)


ios/styles/StyleConfig.mm
```diff
@@ -1786,16 +1798,30 @@ - (CGFloat)effectiveListGapWidth
...
+  return _listStyleMarkerWidth > natural ? _listStyleMarkerWidth : natural;
+}
+
+- (CGFloat)effectiveListMarginLeftForTask
+{
+  // Task checkbox reserves at least its own size. `markerWidth` can widen
+  // the column so task rows align with adjacent UL/OL rows. The checkbox
+  // glyph itself is still drawn at its configured size.
+  CGFloat natural = [self taskListCheckboxSize];
+  return _listStyleMarkerWidth > natural ? _listStyleMarkerWidth : natural;
```
hryhoriiK97: The three methods all do  `return _listStyleMarkerWidth > natural ? _listStyleMarkerWidth : natural;`

We can introduce this helper:

```
static inline CGFloat floorByMarkerWidth(CGFloat natural, CGFloat markerWidth) {
   return markerWidth > natural ? markerWidth : natural;
 }

```

src/normalizeMarkdownStyle.ts
```diff
@@ -112,6 +112,7 @@ const DEFAULT_NORMALIZED_STYLE = Object.freeze({
     marginBottom: 16,
     bulletColor: normalizeColor('#6B7280')!,
     bulletSize: 6,
+    markerWidth: -1,
```
hryhoriiK97: Is there a reason for -1 as the "auto" marker rather than 0? 🤔 Looking at effectiveMarkerWidth(natural)'s max(markerWidth, natural) semantic, 0 would yield the same behavior (max(0, natural) === natural) without needing a magic value. That would let you drop the three < 0 checks (TS normalizeMarkdownStyle, Android ListStyle.fromReadableMap, iOS effectiveListMarginLeftFor*)

src/types/MarkdownStyle.ts
```diff
@@ -28,6 +28,7 @@ interface BlockquoteStyle extends BaseBlockStyle {
 interface ListStyle extends BaseBlockStyle {
   bulletColor?: string;
   bulletSize?: number;
+  markerWidth?: number;
```
hryhoriiK97: Please include JSDoc for the public types. For example:

```
/**
 * Minimum reserved marker column width applied uniformly to UL/OL/task lists.
 * `0` (the default) means no minimum — each list uses its natural marker width.
 */
```

android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
```diff
@@ -32,6 +38,13 @@ data class ListStyle(
       val lineHeight = parser.toPixelFromSP(lineHeightRaw)
       val bulletColor = parser.parseColor(map, "bulletColor")
       val bulletSize = parser.toPixelFromDIP(map.getDouble("bulletSize").toFloat())
+      val markerWidthRaw = map.getDouble("markerWidth").toFloat()
+      val markerWidth =
+        if (markerWidthRaw < 0f) {
+          -1f
+        } else {
+          parser.toPixelFromDIP(markerWidthRaw)
+        }
```
hryhoriiK97: With the 0-as-auto approach (see [comment](https://github.com/software-mansion-labs/react-native-enriched-markdown/pull/248/changes#r3104939892)), this whole block collapses to:

```
val markerWidth = parser.toPixelFromDIP(map.getDouble("markerWidth").toFloat().coerceAtLeast(0f))
```

android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
```diff
@@ -12,11 +12,17 @@ data class ListStyle(
...
+  val markerWidth: Float,
   val markerColor: Int,
   val markerFontWeight: String,
   val gapWidth: Float,
   val marginLeft: Float,
 ) : BaseBlockStyle {
+
+  /** Floor `naturalWidth` by the consumer-configured `markerWidth`. */
+  fun effectiveMarkerWidth(naturalWidth: Float): Float =
+    if (markerWidth > naturalWidth) markerWidth else naturalWidth
```
hryhoriiK97: Could we rewrite this as coerceAtLeast? Behavior is identical (coerceAtLeast(min) is if (this < min) min else this), and the call carries the floor semantic into every call site without needing the doc to explain it.

```
 fun effectiveMarkerWidth(naturalWidth: Float): Float = naturalWidth.coerceAtLeast(markerWidth)
```

Also, we can drop KDoc for this - the function name effectiveMarkerWidth says what it returns.

src/types/MarkdownStyleInternal.ts
```diff
@@ -34,6 +34,10 @@ interface BlockquoteStyleInternal extends BaseBlockStyleInternal {
 interface ListStyleInternal extends BaseBlockStyleInternal {
   bulletColor: string;
   bulletSize: number;
+  // Minimum reserved marker column width. Negative = "auto" → use each
+  // list's natural marker width (UL: bulletSize/2, OL: width of "99.",
+  // task: checkbox size).
```
hryhoriiK97: Please delete this comment. Having JSDoc in the public MarkdownStyle is enough.

src/types/MarkdownStyleInternal.ts
```diff
@@ -34,6 +34,10 @@ interface BlockquoteStyleInternal extends BaseBlockStyleInternal {
...
+  markerWidth: number;
```
hryhoriiK97: Maybe rename this to markerMinWidth for clarity (across all instances)?

hryhoriiK97 (changes requested):
@mozharovsky, thank you for the PR! It's a great improvement for the lists 💪 I've left a few comments. 🙂

ios/styles/StyleConfig.mm
mozharovsky: Good call — with the `0`-as-default switch (see below), the three methods reduce to one-line `MAX(...)` calls each, so the helper isn't needed anymore. Done in the follow-up commit.

src/normalizeMarkdownStyle.ts
mozharovsky: Agreed — switched to `0` as the default. Floor semantics make it work naturally (`max(0, natural) === natural`), and the three guards in normalize/Android parse/iOS effective accessors are gone. Much cleaner, thanks for the nudge.

src/types/MarkdownStyle.ts
mozharovsky: Added — used your suggested wording verbatim, since it fits exactly (and matches the one-line row in `STYLES.md`).

android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
mozharovsky: Collapsed to one line, thanks.

android/src/main/java/com/swmansion/enriched/markdown/styles/ListStyle.kt
mozharovsky: Done — `effectiveMarkerWidth` is now `naturalWidth.coerceAtLeast(markerMinWidth)` and the KDoc is gone.

src/types/MarkdownStyleInternal.ts
mozharovsky: Removed — the JSDoc now lives on the public `MarkdownStyle.ListStyle` field instead.

src/types/MarkdownStyleInternal.ts
mozharovsky: Renamed across the board — public type, internal spec, iOS ivar/getter/setter, Android field, style normalizer (both native and web), docs, and the example app. `markerMinWidth` reads cleanly as "at least this wide" now.

mozharovsky:
Thanks for the thorough review, @hryhoriiK97 — all seven points accepted in [d505e61](https://github.com/software-mansion-labs/react-native-enriched-markdown/pull/248/commits/d505e61). Summary:

- **Rename** `list.markerWidth` → `list.markerMinWidth` across public/internal types, codegen specs, iOS ivar + getter/setter, Android field, style normalizer (native + web), docs, and the example.
- **Default** is now `0` instead of a `-1` sentinel. Floor semantics (`max(0, natural) === natural`) made the sentinel unnecessary; the three `< 0` guards are gone.
- **iOS** `effectiveListMarginLeftForBullet` / `…ForNumber` / `…ForTask` are now single-line `MAX(...)` calls — no helper needed.
- **Android** `effectiveMarkerWidth` uses idiomatic `coerceAtLeast`; `fromReadableMap` parsing is one line; KDoc dropped (the name speaks for itself).
- **JSDoc** moved onto the public `MarkdownStyle.ListStyle` field (where consumers hover it); dropped from `MarkdownStyleInternal`.

Net diff from this commit: `-48 / +30`. Ready for another look whenever you have a minute.

hryhoriiK97 (approved):
LGTM! Thank you for the contribution. @mozharovsky, markerMinWidth will be included in tomorrow’s nightly release.

Issue #246, opened 2026-04-15
Support for citation chips


xindixu:
First of all, thank you for this amazing library! ❤️ 

**Is your feature request related to a problem? Please describe.**
Partially. I'm switching from react-native-markdown-display so that we can have native text selection support on mobile. I'm slightly frustrated that this library doesn't support custom renderers, esp for inline elements (link, mention, etc)

**Describe the solution you'd like**
<img width="858" height="88" alt="Image" src="https://github.com/user-attachments/assets/41c9539f-c38e-408c-a42d-cc4a776cb7c9" />

Would be great if this library to support citation chips. Something like this: 
1. Show a superscript of the source number
2. On hover, show a tooltip with any content
3. On click, it opens the link to the cited site
4. A way to distinguish different links and their rendering logic. Examples:
    - `[some library](https://google.com)` → normal link
    - `{@}[Casper](de047f59-b2e4-43d7-bf85-786087c98a72)` → mentioning users
    - `{+}[table.csv](Generated/table.csv)` → mentioning files
    - `[people.com](https://people.com?citation=1)` → citation chips

Nice to have:
1. Allow users define: 
    - how to render the display text
    - how to render the tooltip
    - how to distinguish different links (regex-based)
2. Users can highlight text around citation chips and mention (i.e. they won't cut the paragraph flow)

**Describe alternatives you've considered**
I've been manually constructing the citation chips and mention link. I was able to achieve this by:
1. Manually parsing the markdown content
2. If the markdown contains citation chips or mentions, I'll use custom rendering logic to compose the block instead of using this library. (ex: mention chip, bolded text, normal text, code, citation)

FYI, I've been applying similar logic to block content like code blocks, tables, latex math. Block content is much easier to deal with, so custom rendering logic works nicely over there. 

Example screenshots

<img width="871" height="411" alt="Image" src="https://github.com/user-attachments/assets/bd2fd7fd-10c5-49f8-b74a-12955d9dabb9" />

<img width="692" height="491" alt="Image" src="https://github.com/user-attachments/assets/1ad474b1-6607-4a44-8758-482e069b7df6" />

hryhoriiK97:
Hi @xindixu , thank you for creating this request!

As discussed in #54, inline custom renderers are very challenging in our current architecture - everything is rendered natively via AttributedStrings (iOS) and Spannables (Android), so injecting custom renderers from the JS side would introduce significant complexity, maintenance overhead, and would be prone to bugs.

[...]

PR #249, opened 2026-04-17
feat: mention and citation links with custom event handler and styling support (display only)

xindixu:
### What/Why?

Add mention and citation links to the `<EnrichedMarkdownText />` component only. Supports custom event handler and styling.

### Testing

Manually performed these tests
- Mention and citation with custom styling background, border, color, and padding
- Copy as text and copy as markdown
- Styling looks correct when they are rendered inside: paragraph, block quotes, tables, lists
- Clicking/pressing the mention and citation links work as expected

#### Screenshots 
| iOS | Android | Web
| - | - | - | 
| ios citation   mention| android citation   mention | web citation   mention

### PR Checklist


hryhoriiK97 (commented):
@xindixu thanks a lot for the PR! Really appreciate it.

A few pieces of feedback before we move forward:

### The PR is too large to land as-is
~3k lines across ~70 files, bundling at least four independent concerns:

- BlockquoteSpan rework + new SPAN_FLAGS_CONTAINER_BACKGROUND priority flag — useful on its own, affects every blockquote consumer.
- Build/env fixes (podspec HEADER_SEARCH_PATHS, CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES, Gemfile nkf).
- Citations.
- Mentions.

### Mentions

We have a parser-level approach in flight: https://github.com/hryhoriiK97/md4c/pull/1 adds a real `MD_SPAN_MENTION` extension to the md4c fork we already use. Syntax: `@[display](target)` / `#[display](target)`. The indicator and free-form target are surfaced via `MD_SPAN_MENTION_DETAIL`, and the application interprets the target.

Why do we prefer that over URL-scheme encoding on Link:

- **First-class AST node** — fits the existing "md4c → AST → per-platform renderer" pipeline (same shape as spoiler, math, underline). The current PR forces parallel `if (url.startsWith("mention://"))` branches into LinkRenderer on iOS / Android / web, plus `MarkdownExtractor`, `LinkTapUtils`, `PasteboardUtils`, `SelectionActionMode`, and the table tap path. That branching grows linearly with each new "kind of inline."
- **Editor support becomes possible.** A real AST node lets `ENRMInputParser` / `InputParser.kt` recognise mentions atomically in `EnrichedMarkdownInput` (select-as-one, delete-as-one, @-typing autocomplete). URL-scheme encoding is display-only on `EnrichedMarkdownText` with no clean path to editor support.
- **No collisions** — any existing `[text](mention://...)` a consumer authored before enabling the feature would silently turn into a mention. With distinct syntax, regular link behaviour is unchanged for everyone who doesn't enable the flag.
- **Cleaner event payload** — `{ indicator, target, displayText }` instead of consumers re-parsing `mention://uuid?type=user`.

### Citations

Worth discussing separately in an issue — they're link-shaped (clickable URL reference), so the design choice (separate md4c extension vs. styled-link convention vs. not shipping) is independent from mentions.

### Suggested next steps

- **Mentions**: land via the md4c [PR](https://github.com/hryhoriiK97/md4c/pull/1) + AST integration once it merges. Happy to share the integration plan then.
- **Citations** — let's discuss in an issue to settle on the design before writing any code.
- **BlockquoteSpan + SPAN_FLAGS_CONTAINER_BACKGROUND** — could land as its own small PR, unblocks any future inline-chip work. Could you please open a separate PR containing only this change? 🙏
- **Podspec / Gemfile fixes** — could you share the exact build errors each one resolves, along with the RN/Xcode/Ruby versions where they reproduce? If you can, please open a separate issue with a reproduction. We don't see those problems on our current setup.

Thanks for your patience and understanding! We want to build a generic solution that works across various use cases and remains easier to maintain and scale - it just takes a bit more time and effort. 🙂

xindixu:
@hryhoriiK97 Thank you for the advice! I'll work on these next week!

xindixu:
@hryhoriiK97 I've pulled out blockquote related fixes here: https://github.com/software-mansion-labs/react-native-enriched-markdown/pull/253

eszlamczyk:
@xindixu Closing this due to https://github.com/software-mansion/react-native-enriched-markdown/issues/246#issuecomment-4981625843

PR #250, opened 2026-04-18
feat: custom selection color

xindixu:
### What/Why?

Add support for custom selection color to the `<EnrichedMarkdownText />` component. Added 2 props: `selectionColor` and `selectionHandleColor`. See React Native doc:
https://reactnative.dev/docs/textinput#selectioncolor and https://reactnative.dev/docs/textinput#selectioncolor for more detail

### Testing

Test with selecting text in both ios and android.

#### Screenshots 

| iOS | Android | web
| - | - | - |
| ios custom selection color | android custom selection color | web custom selection color

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -18,6 +18,7 @@ import com.swmansion.enriched.markdown.utils.common.FeatureFlags
 import com.swmansion.enriched.markdown.utils.common.MarkdownSegmentRenderer
 import com.swmansion.enriched.markdown.utils.common.RenderedSegment
 import com.swmansion.enriched.markdown.utils.common.splitASTIntoSegments
+import com.swmansion.enriched.markdown.utils.text.view.applyMarkdownSelectionColors
 import com.swmansion.enriched.markdown.utils.text.view.emitLinkLongPressEvent
 import com.swmansion.enriched.markdown.utils.text.view.emitLinkPressEvent
 import com.swmansion.enriched.markdown.views.BlockSegmentView
@@ -54,6 +55,8 @@ class EnrichedMarkdown
     private var maxFontSizeMultiplier: Float = 0f
     private var allowTrailingMargin: Boolean = false
     private var selectable: Boolean = true
+    private var propSelectionColor: Int? = null
+    private var propSelectionHandleColor: Int? = null
 
     private var onLinkPressCallback: ((String) -> Unit)? = null
     private var onLinkLongPressCallback: ((String) -> Unit)? = null
@@ -125,6 +128,22 @@ class EnrichedMarkdown
       selectable = value
       segmentViews.filterIsInstance<EnrichedMarkdownInternalText>().forEach {
         it.setIsSelectable(value)
+      }
+    }
+
+    fun setSelectionColorFromProps(color: Int?) {
+      propSelectionColor = color
+      applySelectionColorsToSegments()
+    }
+
+    fun setSelectionHandleColorFromProps(color: Int?) {
+      propSelectionHandleColor = color
+      applySelectionColorsToSegments()
+    }
+
+    private fun applySelectionColorsToSegments() {
+      segmentViews.filterIsInstance<EnrichedMarkdownInternalText>().forEach {
+        it.applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
       }
     }
 
@@ -234,6 +253,8 @@ class EnrichedMarkdown
         if (contextMenuItemTexts.isNotEmpty()) {
           setContextMenuItems(contextMenuItemTexts, ::forwardContextMenuItemPress)
         }
+
+        applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
       }
 
     private fun createTableView(
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -18,7 +18,7 @@ import com.swmansion.enriched.markdown.utils.common.FeatureFlags
 import com.swmansion.enriched.markdown.utils.common.MarkdownSegmentRenderer
 import com.swmansion.enriched.markdown.utils.common.RenderedSegment
 import com.swmansion.enriched.markdown.utils.common.splitASTIntoSegments
-import com.swmansion.enriched.markdown.utils.text.view.applyMarkdownSelectionColors
+import com.swmansion.enriched.markdown.utils.text.view.applySelectionColors
 import com.swmansion.enriched.markdown.utils.text.view.emitLinkLongPressEvent
 import com.swmansion.enriched.markdown.utils.text.view.emitLinkPressEvent
 import com.swmansion.enriched.markdown.views.BlockSegmentView
@@ -55,8 +55,8 @@ class EnrichedMarkdown
     private var maxFontSizeMultiplier: Float = 0f
     private var allowTrailingMargin: Boolean = false
     private var selectable: Boolean = true
-    private var propSelectionColor: Int? = null
-    private var propSelectionHandleColor: Int? = null
+    private var selectionColor: Int? = null
+    private var selectionHandleColor: Int? = null
 
     private var onLinkPressCallback: ((String) -> Unit)? = null
     private var onLinkLongPressCallback: ((String) -> Unit)? = null
@@ -131,19 +131,19 @@ class EnrichedMarkdown
       }
     }
 
-    fun setSelectionColorFromProps(color: Int?) {
-      propSelectionColor = color
+    fun setSelectionColor(color: Int?) {
+      selectionColor = color
       applySelectionColorsToSegments()
     }
 
-    fun setSelectionHandleColorFromProps(color: Int?) {
-      propSelectionHandleColor = color
+    fun setSelectionHandleColor(color: Int?) {
+      selectionHandleColor = color
       applySelectionColorsToSegments()
     }
 
     private fun applySelectionColorsToSegments() {
       segmentViews.filterIsInstance<EnrichedMarkdownInternalText>().forEach {
-        it.applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+        it.applySelectionColors(selectionColor, selectionHandleColor)
       }
     }
 
@@ -254,7 +254,7 @@ class EnrichedMarkdown
           setContextMenuItems(contextMenuItemTexts, ::forwardContextMenuItemPress)
         }
 
-        applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+        applySelectionColors(selectionColor, selectionHandleColor)
       }
 
     private fun createTableView(
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
@@ -80,6 +80,20 @@ class EnrichedMarkdownManager :
     view?.setIsSelectable(selectable)
   }
 
+  override fun setSelectionColor(
+    view: EnrichedMarkdown?,
+    value: Int?,
+  ) {
+    view?.setSelectionColorFromProps(value)
+  }
+
+  override fun setSelectionHandleColor(
+    view: EnrichedMarkdown?,
+    value: Int?,
+  ) {
+    view?.setSelectionHandleColorFromProps(value)
+  }
+
   @ReactProp(name = "md4cFlags")
   override fun setMd4cFlags(
     view: EnrichedMarkdown?,
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
@@ -84,14 +84,14 @@ class EnrichedMarkdownManager :
     view: EnrichedMarkdown?,
     value: Int?,
   ) {
-    view?.setSelectionColorFromProps(value)
+    view?.setSelectionColor(value)
   }
 
   override fun setSelectionHandleColor(
     view: EnrichedMarkdown?,
     value: Int?,
   ) {
-    view?.setSelectionHandleColorFromProps(value)
+    view?.setSelectionHandleColor(value)
   }
 
   @ReactProp(name = "md4cFlags")
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
@@ -22,6 +22,7 @@ import com.swmansion.enriched.markdown.styles.StyleConfig
 import com.swmansion.enriched.markdown.utils.text.TailFadeInAnimator
 import com.swmansion.enriched.markdown.utils.text.interaction.CheckboxTouchHelper
 import com.swmansion.enriched.markdown.utils.text.view.LinkLongPressMovementMethod
+import com.swmansion.enriched.markdown.utils.text.view.applyMarkdownSelectionColors
 import com.swmansion.enriched.markdown.utils.text.view.applySelectableState
 import com.swmansion.enriched.markdown.utils.text.view.cancelJSTouchForCheckboxTap
 import com.swmansion.enriched.markdown.utils.text.view.cancelJSTouchForLinkTap
@@ -80,6 +81,9 @@ class EnrichedMarkdownText
     override var spoilerOverlayDrawer: SpoilerOverlayDrawer? = null
       private set
     var spoilerOverlay: SpoilerOverlay = SpoilerOverlay.PARTICLES
+
+    private var propSelectionColor: Int? = null
+    private var propSelectionHandleColor: Int? = null
 
     init {
       setupAsMarkdownTextView()
@@ -248,6 +252,8 @@ class EnrichedMarkdownText
         fadeAnimator?.animate(tailStart, styledText.length)
         previousTextLength = styledText.length
       }
+
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
     }
 
     fun setContextMenuItems(items: List<String>) {
@@ -256,6 +262,16 @@ class EnrichedMarkdownText
 
     fun setIsSelectable(selectable: Boolean) {
       applySelectableState(selectable)
+    }
+
+    fun setSelectionColorFromProps(color: Int?) {
+      propSelectionColor = color
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+    }
+
+    fun setSelectionHandleColorFromProps(color: Int?) {
+      propSelectionHandleColor = color
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
     }
 
     fun emitOnLinkPress(url: String) {
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
@@ -22,8 +22,8 @@ import com.swmansion.enriched.markdown.styles.StyleConfig
 import com.swmansion.enriched.markdown.utils.text.TailFadeInAnimator
 import com.swmansion.enriched.markdown.utils.text.interaction.CheckboxTouchHelper
 import com.swmansion.enriched.markdown.utils.text.view.LinkLongPressMovementMethod
-import com.swmansion.enriched.markdown.utils.text.view.applyMarkdownSelectionColors
 import com.swmansion.enriched.markdown.utils.text.view.applySelectableState
+import com.swmansion.enriched.markdown.utils.text.view.applySelectionColors
 import com.swmansion.enriched.markdown.utils.text.view.cancelJSTouchForCheckboxTap
 import com.swmansion.enriched.markdown.utils.text.view.cancelJSTouchForLinkTap
 import com.swmansion.enriched.markdown.utils.text.view.createSelectionActionModeCallback
@@ -82,8 +82,8 @@ class EnrichedMarkdownText
       private set
     var spoilerOverlay: SpoilerOverlay = SpoilerOverlay.PARTICLES
 
-    private var propSelectionColor: Int? = null
-    private var propSelectionHandleColor: Int? = null
+    private var selectionColor: Int? = null
+    private var selectionHandleColor: Int? = null
 
     init {
       setupAsMarkdownTextView()
@@ -253,7 +253,7 @@ class EnrichedMarkdownText
         previousTextLength = styledText.length
       }
 
-      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+      applySelectionColors(selectionColor, selectionHandleColor)
     }
 
     fun setContextMenuItems(items: List<String>) {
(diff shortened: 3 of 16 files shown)


hryhoriiK97 (changes requested):
Thanks for the PR! It's a great addition to enriched-markdown 🙌 ❤️

Before going further on implementation details, I'd like to propose simplifying the API.

[react-native-enriched ](https://github.com/software-mansion/react-native-enriched) solves the same problem with a single selectionColor prop, explicitly aligning with React Native's convention (TextInput also exposes only selectionColor — no selectionHandleColor). 

We'd like to match that pattern here. I recommend checking their implementation for reference.

xindixu:
@hryhoriiK97 Thank you for the review! ❤️ 

> (TextInput also exposes only selectionColor — no selectionHandleColor).

I'm pretty new to React Native, so I could be wrong. The React Native doc does mention the `selectionHandleColor` prop for Android: https://reactnative.dev/docs/textinput#selectionhandlecolor-android

In Android, the handler color matches the selection color precisely, but the selection color needs to be much lighter for the underlying text to be shown properly. iOS has the computation done automatically, but Android doesn't. Basically, I'm trying to make the experience look similar in iOS and Android.

Set `selectionHandleColor` to the same value as `selectionColor`
same color

Set `selectionHandleColor` to a darker value than `selectionColor`

different color

hryhoriiK97:
> @hryhoriiK97 Thank you for the review! ❤️
> 
> > (TextInput also exposes only selectionColor — no selectionHandleColor).
> 
> I'm pretty new to React Native, so I could be wrong. The React Native doc does mention the `selectionHandleColor` prop for Android: https://reactnative.dev/docs/textinput#selectionhandlecolor-android
> 
> In Android, the handler color matches the selection color precisely, but the selection color needs to be much lighter for the underlying text to be shown properly. iOS has the computation done automatically, but Android doesn't. Basically, I'm trying to make the experience look similar in iOS and Android.
> 
> Set `selectionHandleColor` to the same value as `selectionColor` same color
> 
> Set `selectionHandleColor` to a darker value than `selectionColor`
> 
> different color

@xindixu thank you for catching this! I'd totally missed this prop. In that case, let's go with this:
- selectionColor — iOS + Android + web.
- selectionHandleColor — Android only (mark as @platform android in the JSDoc). Ignored on iOS entirely.
- Web: drop the accent-color implementation. It doesn't tint selection handles and accidentally tints task list checkboxes. Only selectionColor should affect ::selection.

xindixu:
@hryhoriiK97 Updated based on your review! PTAL. Thanks!

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -54,6 +55,8 @@ class EnrichedMarkdown
     private var maxFontSizeMultiplier: Float = 0f
     private var allowTrailingMargin: Boolean = false
     private var selectable: Boolean = true
+    private var propSelectionColor: Int? = null
+    private var propSelectionHandleColor: Int? = null
```
hryhoriiK97, line 60: Please don't use prop/FromProps convention — no other field/setter in the codebase uses it.

`propSelectionColor` => `selectionColor`
`propSelectionHandleColor` => `selectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -128,6 +131,22 @@ class EnrichedMarkdown
...
 
+    fun setSelectionColorFromProps(color: Int?) {
+      propSelectionColor = color
+      applySelectionColorsToSegments()
+    }
+
+    fun setSelectionHandleColorFromProps(color: Int?) {
+      propSelectionHandleColor = color
+      applySelectionColorsToSegments()
+    }
```
hryhoriiK97, line 142: The same here:
`setSelectionColorFromProps` => `setSelectionColor`
`setSelectionHandleColorFromProps` => `setSelectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
```diff
@@ -81,6 +82,9 @@ class EnrichedMarkdownText
       private set
     var spoilerOverlay: SpoilerOverlay = SpoilerOverlay.PARTICLES
 
+    private var propSelectionColor: Int? = null
+    private var propSelectionHandleColor: Int? = null
```
hryhoriiK97: `propSelectionColor` => `selectionColor`
`propSelectionHandleColor` => `selectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
```diff
@@ -258,6 +264,16 @@ class EnrichedMarkdownText
...
 
+    fun setSelectionColorFromProps(color: Int?) {
+      propSelectionColor = color
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+    }
+
+    fun setSelectionHandleColorFromProps(color: Int?) {
+      propSelectionHandleColor = color
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
+    }
```
hryhoriiK97, line 275: `setSelectionColorFromProps` => `setSelectionColor`
`setSelectionHandleColorFromProps` => `setSelectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
```diff
@@ -80,6 +80,20 @@ class EnrichedMarkdownManager :
     view?.setIsSelectable(selectable)
   }
 
+  override fun setSelectionColor(
+    view: EnrichedMarkdown?,
+    value: Int?,
+  ) {
+    view?.setSelectionColorFromProps(value)
```
hryhoriiK97: `setSelectionColorFromProps` => `setSelectionColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
```diff
@@ -80,6 +80,20 @@ class EnrichedMarkdownManager :
...
+  }
+
+  override fun setSelectionHandleColor(
+    view: EnrichedMarkdown?,
+    value: Int?,
+  ) {
+    view?.setSelectionHandleColorFromProps(value)
```
hryhoriiK97: `setSelectionHandleColorFromProps => setSelectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownTextManager.kt
```diff
@@ -100,6 +100,20 @@ class EnrichedMarkdownTextManager :
     view?.setIsSelectable(selectable)
   }
 
+  override fun setSelectionColor(
+    view: EnrichedMarkdownText?,
+    value: Int?,
+  ) {
+    view?.setSelectionColorFromProps(value)
```
hryhoriiK97: `setSelectionColorFromProps` => `setSelectionColor`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownTextManager.kt
```diff
@@ -100,6 +100,20 @@ class EnrichedMarkdownTextManager :
...
+  }
+
+  override fun setSelectionHandleColor(
+    view: EnrichedMarkdownText?,
+    value: Int?,
+  ) {
+    view?.setSelectionHandleColorFromProps(value)
```
hryhoriiK97: `setSelectionHandleColorFromProps` => `setSelectionHandleColor`

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.core.graphics.drawable.DrawableCompat
+
+/**
+ * Applies selection highlight and (where supported) handle tinting to a [TextView].
+ *
+ * Handle drawables are only tinted on API 29+ where the framework exposes getters;
+ * on older versions the handle theme defaults remain unchanged.
+ */
+fun TextView.applyMarkdownSelectionColors(
+  selectionColor: Int?,
```
hryhoriiK97, line 22: Please let's rename `applyMarkdownSelectionColors` to `applySelectionColors` and update the two call sites in EnrichedMarkdown and EnrichedMarkdownText accordingly.

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -128,6 +131,22 @@ class EnrichedMarkdown
...
+
+    private fun applySelectionColorsToSegments() {
+      segmentViews.filterIsInstance<EnrichedMarkdownInternalText>().forEach {
+        it.applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
```
hryhoriiK97: `applyMarkdownSelectionColors` => `applySelectionColors`

android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
```diff
@@ -248,6 +252,8 @@ class EnrichedMarkdownText
         fadeAnimator?.animate(tailStart, styledText.length)
         previousTextLength = styledText.length
       }
+
+      applyMarkdownSelectionColors(propSelectionColor, propSelectionHandleColor)
```
hryhoriiK97: `applyMarkdownSelectionColors` => `applySelectionColors`

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
```diff
@@ -0,0 +1,47 @@
...
+    tintHandle(textSelectHandleLeft, color)?.let { setTextSelectHandleLeft(it) }
+    tintHandle(textSelectHandle, color)?.let { setTextSelectHandle(it) }
+    tintHandle(textSelectHandleRight, color)?.let { setTextSelectHandleRight(it) }
+  } catch (_: Exception) {
+    // Defensive: OEM TextView variants may not support all handle accessors.
+  }
+}
+
+private fun tintHandle(
+  drawable: android.graphics.drawable.Drawable?,
```
hryhoriiK97: Please move `android.graphics.drawable.Drawable?` to the top of the file, like every other file in this package

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
```diff
@@ -0,0 +1,47 @@
...
+  @ColorInt color: Int,
+): android.graphics.drawable.Drawable? {
```
hryhoriiK97: The same as above 🔝

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
```diff
@@ -0,0 +1,47 @@
...
+    return
+  }
+  try {
+    tintHandle(textSelectHandleLeft, color)?.let { setTextSelectHandleLeft(it) }
+    tintHandle(textSelectHandle, color)?.let { setTextSelectHandle(it) }
+    tintHandle(textSelectHandleRight, color)?.let { setTextSelectHandleRight(it) }
+  } catch (_: Exception) {
+    // Defensive: OEM TextView variants may not support all handle accessors.
+  }
+}
```
hryhoriiK97: Regarding the implementation of applySelectionHandleTint—I recommend going with something like this:

```
private fun TextView.applySelectionHandleTint(@ColorInt color: Int) {
  if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) return

  val handles = listOf(
    ::getTextSelectHandleLeft to ::setTextSelectHandleLeft,
    ::getTextSelectHandle to ::setTextSelectHandle,
    ::getTextSelectHandleRight to ::setTextSelectHandleRight,
  )

  handles.forEach { (getter, setter) ->
    try {
      getter()?.mutate()?.also { DrawableCompat.setTint(it, color) }?.let(setter)
    } catch (e: LinkageError) {
      // Defensive: OEM TextView variants may strip individual handle accessors.
      Log.w(TAG, "Selection handle tint skipped: ${e.message}")
    }
  }
}
```

ios/EnrichedMarkdown.mm
```diff
@@ -880,4 +896,13 @@ - (NSInteger)indexOfAccessibilityElement:(id)element
...
 #endif
 
+- (void)applySelectionTintFromProps:(const EnrichedMarkdownProps &)props toTextView:(ENRMPlatformTextView *)textView
+{
+#if !TARGET_OS_OSX
+  if (isColorMeaningful(props.selectionColor)) {
+    ENRMSetSelectionColor(textView, RCTUIColorFromSharedColor(props.selectionColor));
+  }
+#endif
+}
```
hryhoriiK97, line 908: Missing reset path. The diff in updateProps correctly fires on (set → unset), but this helper only acts when isColorMeaningful(...) is true — so clearing the prop from JS leaves tintColor stuck at the previous value forever. Consumers will hit this with patterns like `selectionColor={isDarkMode ? '#fff' : undefined}`.


```
- (void)applySelectionTintFromProps:(const EnrichedMarkdownProps &)props toTextView:(ENRMPlatformTextView *)textView
 {
 #if !TARGET_OS_OSX
   if (isColorMeaningful(props.selectionColor)) {
     ENRMSetSelectionColor(textView, RCTUIColorFromSharedColor(props.selectionColor));
   } else {
     ENRMSetSelectionColor(textView, nil); // resets to system tint
   }
 #endif
 }

```

ios/EnrichedMarkdownText.mm
```diff
@@ -411,6 +412,14 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
     _textView.selectable = newViewProps.selectable;
   }
 
+  if (newViewProps.selectionColor != oldViewProps.selectionColor) {
+#if !TARGET_OS_OSX
+    if (isColorMeaningful(newViewProps.selectionColor)) {
+      ENRMSetSelectionColor(_textView, RCTUIColorFromSharedColor(newViewProps.selectionColor));
+    }
+#endif
+  }
```
hryhoriiK97, line 423: Same missing reset path as in EnrichedMarkdown.mm - needs the explicit else branch.

ios/EnrichedMarkdown.mm
hryhoriiK97: Let's also rename applySelectionTintFromProps to applySelectionColor 🙏

src/web/EnrichedMarkdownText.tsx
```diff
@@ -95,21 +102,43 @@ export const EnrichedMarkdownText = ({
...
   );
 
-  const wrapperStyle = useMemo<CSSProperties>(
-    () => ({
+  const wrapperStyle = useMemo<CSSProperties>(() => {
+    const selectionBgVar =
+      selectionColor != null && selectionColor !== undefined
+        ? String(selectionColor)
+        : undefined;
+
```
hryhoriiK97, line 110: selectionColor != null && selectionColor !== undefined is redundant. != null already covers both null and undefined. We can use !!selectionColor && ....

hryhoriiK97: Regarding `String(selectionColor)` - the codebase already has the right helper  - `normalizeColor`.

src/web/EnrichedMarkdownText.tsx
```diff
@@ -95,21 +102,43 @@ export const EnrichedMarkdownText = ({
...
-    [containerStyle, selectable]
-  );
+      ...(selectionBgVar != null
+        ? ({ ['--enrm-selection-bg']: selectionBgVar } as CSSProperties)
+        : null),
+    };
+  }, [containerStyle, selectable, selectionColor]);
+
+  const selectionStyle =
+    selectionColor != null && selectionColor !== undefined ? (
```
hryhoriiK97: !!selectionColor && ...

hryhoriiK97 (changes requested):
@xindixu I’ve done an initial review and left a few comments 🙂

src/EnrichedMarkdownNativeComponent.ts
```diff
@@ -272,6 +272,25 @@ export interface NativeProps extends ViewProps {
...
+  selectionColor?: ColorValue;
+  /**
+   * Color of the selection handles (drag anchors).
+   *
+   * - **Android**: tints the left, mid, and right handle drawables (API 29+;
+   *   older API levels leave handles at the default theme color).
+   * @platform android
+   */
+  selectionHandleColor?: ColorValue;
   /**
```
hryhoriiK97, line 292: The same as above 🔝

src/EnrichedMarkdownTextNativeComponent.ts
```diff
@@ -272,6 +272,25 @@ export interface NativeProps extends ViewProps {
...
+   */
+  selectionColor?: ColorValue;
+  /**
+   * Color of the selection handles (drag anchors).
+   *
+   * - **Android**: tints the left, mid, and right handle drawables (API 29+;
+   *   older API levels leave handles at the default theme color).
+   * @platform android
+   */
+  selectionHandleColor?: ColorValue;
```
hryhoriiK97, line 291: The same as above 🔝

src/types/MarkdownTextProps.ts
```diff
@@ -91,6 +91,26 @@ export interface EnrichedMarkdownTextProps extends Omit<ViewProps, 'style'> {
...
+  selectionColor?: ColorValue;
+  /**
+   * Color of the selection handles (drag anchors).
+   *
+   * - **Android**: tints the left, mid, and right handle drawables (API 29+;
+   *   older API levels leave handles at the default theme color).
+   *
+   * @platform android
+   */
+  selectionHandleColor?: ColorValue;
```
hryhoriiK97, line 109: Let's simplify the JS docs for selectionColor and selectionHandleColor:

```
 /**
 * Color of the text selection highlight.
 *
 * On iOS, this also affects the caret and selection handle colors
 * (they share a single tint).
 *
 * @platform ios, android, web
 */
 selectionColor?: ColorValue;

 /**
  * Color of the selection handles (drag anchors).
  * No-op on API levels below 29.
  *
  * @platform android
  */
 selectionHandleColor?: ColorValue;
```

src/types/MarkdownTextProps.web.ts
```diff
@@ -65,6 +66,11 @@ export interface EnrichedMarkdownTextProps
    * @platform ios, android, web
    */
   selectable?: boolean;
+  /**
+   * Color of the text selection highlight (`::selection` background).
+   * @platform web
+   */
+  selectionColor?: ColorValue;
```
hryhoriiK97, line 73:
```
 /**
  * Color of the text selection highlight.
  * @platform web
  */
selectionColor?: ColorValue;
```

hryhoriiK97 (changes requested):

ios/EnrichedMarkdown.mm
xindixu: Applied the fix. Thank you!

xindixu:
@hryhoriiK97 thank you for the detailed review! I've addressed all your feedback. 

New tests:

iOS:

https://github.com/user-attachments/assets/6e5b1c64-a269-4042-afa9-eaf5a407d9c9


Android:

https://github.com/user-attachments/assets/b4e28f4f-77ba-4cb1-a814-149a72e92e98

src/web/EnrichedMarkdownText.tsx
```diff
@@ -95,21 +103,41 @@ export const EnrichedMarkdownText = ({
     [lastChildStyle]
   );
 
-  const wrapperStyle = useMemo<CSSProperties>(
-    () => ({
+  const wrapperStyle = useMemo<CSSProperties>(() => {
+    const selectionBgVar = selectionColor
```
hryhoriiK97: Please let's rename selectionBgVar to selectionColorCss 🙏

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
```diff
@@ -0,0 +1,51 @@
...
+import android.os.Build
+import android.util.Log
+import android.widget.TextView
+import androidx.annotation.ColorInt
+import androidx.core.graphics.drawable.DrawableCompat
+
+private const val TAG = "TextSelectionColors"
+
+private typealias HandleGetter = (TextView) -> Drawable?
+private typealias HandleSetter = (TextView, Drawable) -> Unit
```
hryhoriiK97: These are each used in exactly one place — the List<Pair<HandleGetter, HandleSetter>> annotation below — and the annotation itself is redundant since Kotlin can infer the list type from the listOf(... to ...) content.

The setters all match the (TextView, Drawable) -> Unit shape, so both sides of each pair can be plain method references — no lambda wrappers needed:

```
private fun TextView.applySelectionHandleTint(@ColorInt color: Int) {
  if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) return

  val handles = listOf(
    TextView::getTextSelectHandleLeft to TextView::setTextSelectHandleLeft,
    TextView::getTextSelectHandle to TextView::setTextSelectHandle,
    TextView::getTextSelectHandleRight to TextView::setTextSelectHandleRight,
  )

  handles.forEach { (getter, setter) ->
    try {
      getter(this)?.mutate()?.also { DrawableCompat.setTint(it, color) }?.let { setter(this, it) }
    } catch (e: LinkageError) {
      Log.w(TAG, "Selection handle tint skipped: ${e.message}")
    }
  }
}
```

ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -292,6 +292,8 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
   if (newViewProps.selectionColor != oldViewProps.selectionColor) {
     if (isColorMeaningful(newViewProps.selectionColor)) {
       ENRMSetSelectionColor(_textView, RCTUIColorFromSharedColor(newViewProps.selectionColor));
+    } else {
+      ENRMSetSelectionColor(_textView, nil); // resets to inherited / system tint
```
hryhoriiK97: Let's drop this comment

ios/EnrichedMarkdown.mm
```diff
@@ -880,4 +896,15 @@ - (NSInteger)indexOfAccessibilityElement:(id)element
 }
 #endif
 
+- (void)applySelectionColor:(const EnrichedMarkdownProps &)props toTextView:(ENRMPlatformTextView *)textView
+{
+#if !TARGET_OS_OSX
+  if (isColorMeaningful(props.selectionColor)) {
+    ENRMSetSelectionColor(textView, RCTUIColorFromSharedColor(props.selectionColor));
+  } else {
+    ENRMSetSelectionColor(textView, nil); // resets to inherited / system tint
```
hryhoriiK97: The same as above 🔝

ios/EnrichedMarkdownText.mm
```diff
@@ -411,6 +412,16 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
     _textView.selectable = newViewProps.selectable;
   }
 
+  if (newViewProps.selectionColor != oldViewProps.selectionColor) {
+#if !TARGET_OS_OSX
+    if (isColorMeaningful(newViewProps.selectionColor)) {
+      ENRMSetSelectionColor(_textView, RCTUIColorFromSharedColor(newViewProps.selectionColor));
+    } else {
+      ENRMSetSelectionColor(_textView, nil); // resets to inherited / system tint
```
hryhoriiK97: The same as above 🔝

hryhoriiK97 (changes requested):
Thank you for addressing all the comments! A few additional ones from me 🙏

android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextSelectionColors.kt
xindixu: I tried applying your suggestion but the code failed to build. Feedback from Cursor:

> On TextView, setTextSelectHandle, setTextSelectHandleLeft, and setTextSelectHandleRight each have two overloads: Drawable and Int.
>
> Using TextView::setTextSelectHandleLeft (and the others) as a bound reference does not pick an overload, so the compiler reports overload resolution ambiguity and then fails to infer the Pair / forEach types.

xindixu:
@hryhoriiK97 Addressed all your feedback. PTAL!

Built android app again and verified that it works as expected

Screenshot 2026-04-22 at 12 43 41 PM

hryhoriiK97 (approved):
Thanks for the contribution! 🚀 A few things still worth improving 0 I'll handle those in follow-up PRs to keep this PR's scope focused. selectionColor will be available in tomorrow's nightly release.

PR #253, opened 2026-04-20
fix(android): span with background should show up inside blockquotes

xindixu:
### What/Why?

Text spans with background inside blockquotes don't have their background showing properly in Android.

- Add SPAN_FLAGS_CONTAINER_BACKGROUND (max priority) so blockquote draws before lower-priority LineBackgroundSpans on the same line.
- Implement LineBackgroundSpan on BlockquoteSpan and move fill from drawLeadingMargin to drawBackground; keep accent bars in drawLeadingMargin.
- Unblocks inline chip/pill backgrounds rendering above blockquote fill.

### Testing

Render code with background color inside blockquotes.

#### Screenshots

| Before | After |
| - | - |
| before fix | after fix | 

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/renderer/BlockquoteRenderer.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/renderer/BlockquoteRenderer.kt
@@ -46,11 +46,9 @@ class BlockquoteRenderer(
         .map { builder.getSpanStart(it) to builder.getSpanEnd(it) }
         .sortedBy { it.first }
 
-    // The Accent Bar Span covers the full range for visual continuity.
-    // Use high-priority flags so BlockquoteSpan's LineBackgroundSpan pass
-    // runs FIRST on each line — the blockquote fill is painted first, then
-    // inline chip/pill backgrounds (mention pills etc.) draw on top of it
-    // instead of being covered by it.
+    // The accent bar span covers the full range for visual continuity.
+    // SPAN_FLAGS_CONTAINER_BACKGROUND keeps the blockquote fill under any
+    // inline chip/pill backgrounds on the same line.
     builder.setSpan(
       BlockquoteSpan(style, depth, factory.context, factory.styleCache),
       start,
--- a/android/src/main/java/com/swmansion/enriched/markdown/renderer/BlockquoteRenderer.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/renderer/BlockquoteRenderer.kt
@@ -3,6 +3,7 @@ package com.swmansion.enriched.markdown.renderer
 import android.text.SpannableStringBuilder
 import com.swmansion.enriched.markdown.parser.MarkdownASTNode
 import com.swmansion.enriched.markdown.spans.BlockquoteSpan
+import com.swmansion.enriched.markdown.utils.text.span.SPAN_FLAGS_CONTAINER_BACKGROUND
 import com.swmansion.enriched.markdown.utils.text.span.SPAN_FLAGS_EXCLUSIVE_EXCLUSIVE
 import com.swmansion.enriched.markdown.utils.text.span.applyMarginBottom
 import com.swmansion.enriched.markdown.utils.text.span.applyMarginTop
@@ -45,12 +46,16 @@ class BlockquoteRenderer(
         .map { builder.getSpanStart(it) to builder.getSpanEnd(it) }
         .sortedBy { it.first }
 
-    // The Accent Bar Span covers the full range for visual continuity
+    // The Accent Bar Span covers the full range for visual continuity.
+    // Use high-priority flags so BlockquoteSpan's LineBackgroundSpan pass
+    // runs FIRST on each line — the blockquote fill is painted first, then
+    // inline chip/pill backgrounds (mention pills etc.) draw on top of it
+    // instead of being covered by it.
     builder.setSpan(
       BlockquoteSpan(style, depth, factory.context, factory.styleCache),
       start,
       end,
-      SPAN_FLAGS_EXCLUSIVE_EXCLUSIVE,
+      SPAN_FLAGS_CONTAINER_BACKGROUND,
     )
 
     // Apply styling only to segments that are NOT nested quotes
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/swmansion/enriched/markdown/utils/text/span/SpanFlags.kt
```diff
@@ -1,5 +1,17 @@
...
+/**
+ * EXCLUSIVE_EXCLUSIVE flags with the maximum span priority bit set. Higher
+ * priority means the span is iterated FIRST during the text view's draw
+ * passes (e.g. `Layout.drawBackground`), which means it's painted FIRST — so
+ * lower-priority spans drawn afterwards end up on top visually. Use this for
+ * full-width container backgrounds (like blockquote) that must sit UNDER any
+ * inline chip / pill backgrounds on the same line.
+ */
+const val SPAN_FLAGS_CONTAINER_BACKGROUND =
+  SpannableString.SPAN_EXCLUSIVE_EXCLUSIVE or ((0xFF) shl Spanned.SPAN_PRIORITY_SHIFT)
```
hryhoriiK97: Spanned.SPAN_PRIORITY is already the pre-shifted full-priority mask (0x0000FF00), which is exactly 0xFF shl 8. Using it removes the magic number and reads as intent:

```
const val SPAN_FLAGS_CONTAINER_BACKGROUND =
  Spanned.SPAN_EXCLUSIVE_EXCLUSIVE or Spanned.SPAN_PRIORITY
```

android/src/main/java/com/swmansion/enriched/markdown/spans/BlockquoteSpan.kt
```diff
@@ -125,10 +147,10 @@ class BlockquoteSpan(
     c: Canvas,
     top: Int,
     bottom: Int,
-    layout: Layout?,
+    right: Int,
   ) {
     val bgColor = blockquoteStyle.backgroundColor?.takeIf { it != Color.TRANSPARENT } ?: return
     val backgroundPaint = configureBackgroundPaint(bgColor)
-    c.drawRect(0f, top.toFloat(), layout?.width?.toFloat() ?: 0f, bottom.toFloat(), backgroundPaint)
+    c.drawRect(0f, top.toFloat(), right.toFloat(), bottom.toFloat(), backgroundPaint)
```
hryhoriiK97: LineBackgroundSpan.drawBackground provides a left: Int parameter that's the actual left edge of the line rect. Hard-coding 0f matches the previous behavior so there's no regression, but using left.toFloat() would be more correct in edge cases (RTL layouts, future padding changes).

android/src/main/java/com/swmansion/enriched/markdown/spans/BlockquoteSpan.kt
```diff
@@ -73,6 +73,28 @@ class BlockquoteSpan(
     }
   }
 
+  /**
+   * Drawn BEFORE glyphs and before other [LineBackgroundSpan]s attached to the
+   * same line, so inline backgrounds painted by code spans render on
+   * top of the blockquote fill instead of being covered by it.
+   */
```
hryhoriiK97: Please let's remove this KDoc 🙏

android/src/main/java/com/swmansion/enriched/markdown/utils/text/span/SpanFlags.kt
```diff
@@ -1,5 +1,17 @@
...
 const val SPAN_FLAGS_EXCLUSIVE_EXCLUSIVE = SpannableString.SPAN_EXCLUSIVE_EXCLUSIVE
+
+/**
+ * EXCLUSIVE_EXCLUSIVE flags with the maximum span priority bit set. Higher
+ * priority means the span is iterated FIRST during the text view's draw
+ * passes (e.g. `Layout.drawBackground`), which means it's painted FIRST — so
+ * lower-priority spans drawn afterwards end up on top visually. Use this for
+ * full-width container backgrounds (like blockquote) that must sit UNDER any
+ * inline chip / pill backgrounds on the same line.
+ */
```
hryhoriiK97, line 15: Let’s update this KDoc to this version:
```
/**
 * `SPAN_EXCLUSIVE_EXCLUSIVE` with the maximum span priority.
 *
 * Higher-priority spans are iterated — and therefore drawn — first, so any
 * lower-priority span on the same line ends up painted *on top* visually.
 * Use this for full-width container backgrounds (e.g. blockquote fill in
 * [BlockquoteSpan]) that must sit under inline pill/chip backgrounds.
 */
 ```

File: android/src/main/java/com/swmansion/enriched/markdown/renderer/BlockquoteRenderer.kt [hunk 5]
```diff
@@ -45,12 +46,16 @@ class BlockquoteRenderer(
         .map { builder.getSpanStart(it) to builder.getSpanEnd(it) }
         .sortedBy { it.first }
 
-    // The Accent Bar Span covers the full range for visual continuity
+    // The Accent Bar Span covers the full range for visual continuity.
+    // Use high-priority flags so BlockquoteSpan's LineBackgroundSpan pass
+    // runs FIRST on each line — the blockquote fill is painted first, then
+    // inline chip/pill backgrounds (mention pills etc.) draw on top of it
+    // instead of being covered by it.
```
hryhoriiK97:
```
// The accent bar span covers the full range for visual continuity.
// SPAN_FLAGS_CONTAINER_BACKGROUND keeps the blockquote fill under any
// inline chip/pill backgrounds on the same line.
```

hryhoriiK97 (changes requested):
@xindixu thank you for the PR! I've left a few comments 🙂

xindixu:
@hryhoriiK97 - Thank you for the review! I've addressed all your comments. Here's the updated screenshot after the fix. PTAL 😃 

Screenshot 2026-04-22 at 2 40 15 PM

hryhoriiK97 (approved):
LGTM 🚀 Thanks for the contribution, @xindixu! This will go out in tomorrow's nightly release.

PR #270, opened 2026-04-25
feat: add GFM streaming support for tables and block math

### LINKED ISSUE #241: Feature Request: Support Streaming Animation for GFM
Currently when using GFM, streaming animation is not supported. I see that there is a TODO comment in the code to implement already.

hryhoriiK97:
### What/Why?

Fixes: #241 

Adds iOS support for smoother `streamingAnimation` with GitHub-flavored markdown tables and block LaTeX.

Previously, GFM block views could be recreated repeatedly while markdown was streaming, causing visible jank and extra UI thread work. This PR splits parsed GFM content into native-renderable segments: text, table, and block math. Each segment gets a deterministic content signature, so iOS can reconcile new renders against existing native views instead of rebuilding completed blocks.

During reconciliation:

- unchanged segments reuse their existing native views,
- shifted unchanged segments can still be reused by signature,
- changed segments of the same kind update in place,
- new or different segment kinds create new views,
- unused old views are removed.

When `streamingAnimation` is active, incomplete trailing tables and block math are hidden until their markdown is complete. This avoids repeatedly rendering expensive partially parsed block views while text continues to stream normally.

The PR also adjusts iOS measurement/layout for streaming GFM so height updates track rendered content more reliably, including newly completed block segments, and keeps table interactions working when layout catches up.

Static GFM behavior remains unchanged.

#### Screenshots
https://github.com/user-attachments/assets/b5d369db-c7c1-4e24-b49a-2a74e2fd5db6

https://github.com/user-attachments/assets/a0bf7b90-eacc-4842-865b-a07ff5d59ecf

### PR Checklist

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -17,12 +17,15 @@ import com.swmansion.enriched.markdown.styles.StyleConfig
 import com.swmansion.enriched.markdown.utils.common.FeatureFlags
 import com.swmansion.enriched.markdown.utils.common.MarkdownSegmentRenderer
 import com.swmansion.enriched.markdown.utils.common.RenderedSegment
+import com.swmansion.enriched.markdown.utils.common.SegmentReconciler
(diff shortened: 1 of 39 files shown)


thanggggg1:
can i have example code for you video case

PR #275, opened 2026-04-28
feat: create test playground + introduce navigation

eszlamczyk:
### What/Why?
First phase of setting up Maestro E2E tests for
react-native-enriched-markdown. To make automated testing
feasible, a dedicated `PlaygroundScreen` was added to the
example app as a stable, testable surface. Additionaly introduced `react-navigation` to the app to make example app better to read.

##### React Navigation
To incorporate react navigation introduced *Home screen* that includes routing buttons for each available example screen (Playground, Text, Input, Stream). On top of that added simple header informing about current screen and giving ability to go back to home screen.

##### Playground
The playground includes:
  - Focus / Blur / Clear controls for the editor
  - Size toggle (base <-> max height) to test dynamic layout
  - Formatting toolbar with Bold, Italic, Underline,
  Strikethrough, Spoiler, and Link toggles
  - Link modal for inserting/editing links
  - "Get Raw Markdown" button that shows the current markdown
  via an alert
  - Live preview panel using `EnrichedMarkdownText` that
  reflects editor content (he made by typing using either features of `EnrichedMarkdownTextInput` or `*normal markdown syntax*`) in real time

### Testing
Run the example app on Android or iOS and navigate to the
Playground screen. Verify that:
  - The formatting toolbar correctly applies and removes
  styles
  - The link modal opens, accepts input, and inserts/removes
  links
  - The size toggle resizes the editor
  - "Get Raw Markdown" displays the current content
  - The preview panel updates live as you type

#### Screenshots

| | Android | IOS |
| -- | -- | -- | 
| Home screen | image | image |
| New Header <br> (example in text view) | image | image
| Playground | image | image|

##### Input element screenshot
image

### PR Checklist


apps/example/src/screens/HomeScreen.tsx
```diff
@@ -0,0 +1,106 @@
+import { View, Text, StyleSheet, TouchableOpacity } from 'react-native';
+import type { NavigationProp } from '@react-navigation/native';
+
+type Props = {
+  navigation: NavigationProp<any>;
```
hryhoriiK97: Thinking about improving our types instead of relying on NavigationProp<any> 🤔. Let's define a RootStackParamList and pass it to createNativeStackNavigator<RootStackParamList>(). This will ensure our route names remain type-safe as we add more examples.

You can find more info (or perhaps a better approach) here: https://reactnavigation.org/docs/typescript/

apps/example/src/screens/HomeScreen.tsx
```diff
@@ -0,0 +1,106 @@
...
+      </Text>
+
+      <TouchableOpacity
+        style={[styles.button, styles.playgroundButton]}
+        onPress={() => navigation.navigate('Playground')}
+        testID="home-block-playground"
+      >
+        <Text style={styles.buttonText}>Playground</Text>
+        <Text style={styles.buttonSubtext}>live editor with preview</Text>
+      </TouchableOpacity>
```
hryhoriiK97: This logic is quite repetitive. We could create a separate component for this TouchableOpacity (including the label) and define an array for the data, e.g., const screens = [{testID, title, description}, ...]. Then, we can simply iterate over the array to render them.

apps/example/src/screens/PlaygroundScreen.tsx
```diff
@@ -0,0 +1,366 @@
...
+                  : styles.input
+              }
+              markdownStyle={MARKDOWN_STYLE}
+              onChangeState={setState}
+              onChangeMarkdown={setMarkdown}
+              onChangeSelection={(sel) =>
+                setHasSelection(sel.start !== sel.end)
+              }
+            />
+            <View style={styles.toolbar} testID="formatting-toolbar">
```
hryhoriiK97: Let's move the formatting toolbar into its own component for PlaygroundScreen and InputScreen. The items and styles are almost the same, so one FormattingToolbar would make changes easier.

Also, can we make the toolbar look better? Maybe use some cool icons? It would help a lot when we make demos for our posts 🙂

hryhoriiK97 (changes requested):
Overall looks really nice - we are moving in the right direction. I've left a few comments, please take a look 🙏

hryhoriiK97 (approved):
LGTM 🚀

PR #280, opened 2026-04-29
fix(ios): spoiler not rendering after keystroke

eszlamczyk:
### What/Why?
While (check video section) working on playground I've found out that `||spoilers||` on IOS are not rendering properly when we try to render them after last `|` keystroke (so `||test|` -> `||test||`). This fixes the issue by setting `needsLayout` flag to trigger render of spoiler "frame" during re-render.  

### Testing
- open playground on IOS
- type `||test|` into input field -> add missing `|` to create spoiler 
    - expected: spoiler is being created
- turn on `||` (spoiler) mode 
    - expected: spoiler is being created in both input field and rendered field

##### Before

https://github.com/user-attachments/assets/78082dd7-6376-41b1-9703-ed8140101a2b

##### After

https://github.com/user-attachments/assets/63a1e15a-e470-45c6-b77f-da3807dd597c 

### PR Checklist

### DIFF
--- a/ios/views/EnrichedMarkdownInternalText.m
+++ b/ios/views/EnrichedMarkdownInternalText.m
@@ -78,6 +78,7 @@
   [_textView.layoutManager invalidateLayoutForCharacterRange:NSMakeRange(0, text.length) actualCharacterRange:NULL];
 
   [_spoilerManager setNeedsUpdate];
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (approved):
LGTM 🎖️

PR #299, opened 2026-05-06
test: maestro text block elements tests

eszlamczyk:
### What/Why?                                                            
                                                                     
Adds Maestro E2E tests covering all block-level markdown elements
rendered by EnrichedMarkdownText. The tests verify correct visual    
rendering by capturing screenshots and comparing against baselines on
both iOS and Android.                                               
                
Block elements covered:
- Blockquote                                   
- Code block (fenced)                                                
- Headers (H1–H6)    
- Block image & inline image                                         
- Math display `($$...$$)`   
- Nested list, ordered list, unordered list
- Task list                                                          
- Paragraph
- Table                                                              
- Thematic break

Also includes:
- Updates to playground (two buttons for creating inline and block text examples)
- Updates to helper scripts (to make maestro e2e tests more robust and less flaky

### Testing
Run all e2e tests
```
yarn test:e2e:mobile
```

Run specific flow
```
yarn test:e2e:mobile -- .maestro/enrichedMarkdownText/flows/[flow name].yaml
```

### PR Checklist

- [x] E2E tests are passing


.maestro/enrichedMarkdownText/flows/blockquote_test.yaml
```diff
@@ -0,0 +1,22 @@
...
+    file: '../../subflows/move_to_playground.yaml'
+
+- runFlow:
+    file: '../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        > this is a text inside a blockquote
+
+         >this is also a text inside a blockquote
+
```
hryhoriiK97, line 16: I think we should also add tests for the nested blockquotes 🤔 Let's create this as a separate test case.

.maestro/enrichedMarkdownText/flows/code_block_test.yaml
```diff
@@ -0,0 +1,25 @@
...
+- runFlow:
+    file: '../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        ```
+        sum = 0
+        for i in range(20):
+          print(i % 3)
+          sum += i % 3
+        print(sum)
```
hryhoriiK97, line 18: Just a consideration – what do you think about adding tests where code blocks are combined with other elements like paragraphs, images, or lists? It would be great to test that margins are preserved correctly between elements. I think this applies to all block elements.

.maestro/enrichedMarkdownText/flows/mixed_nested_list_test.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - Frontend
+          - React
+            - Hooks
+            - Components
+          - Vue
+        - Backend
+          - Node.js
+          - Python
+            1. Django
+            2. FastAPI
```
hryhoriiK97, line 22: Shouldn't we rename this to mixed_nested_list_text.yaml (maybe you will have better name xd)? 🤔 And create a new one, ordered_nested_list_test.yaml (the same for unordered)? What do you think?

.maestro/enrichedMarkdownText/flows/table_test.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+- runFlow:
+    file: '../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        | col1 | col2 | looooooooong col | a |
+        | - | -- | -------------------- | -------- |
+        | test | of | the | t |
+        | the | one | piece is | real
+        | third | Row | to test trailing | symbols | |asdf|||||blabla|
```
hryhoriiK97, line 17: Why do we have these extra pipes at the end? `| symbols | |asdf|||||blabla|`

.maestro/enrichedMarkdownText/screenshots/android/image_display.png
hryhoriiK97, line 1: Just curious why this logo is cut? 🤔

.maestro/enrichedMarkdownText/screenshots/android/nested_list_display.png
hryhoriiK97, line 1: Nice, looks like we found a bug on Android. Can you create a ticket in Linear, pls?

Image

.maestro/enrichedMarkdownText/screenshots/ios/code_block_display.png
hryhoriiK97, line 1: Standalone code block is cut off on iOS – looks like a bug too. We also can add this to our backlog.

Image

hryhoriiK97 (commented):
Looks really nice! I've left a few comments, please take a look at them 🙂

.maestro/enrichedMarkdownText/flows/mixed_nested_list_test.yaml
eszlamczyk: Well, I don't want to write too much tests, especialy on android the tests take (about) 9.5 minutes. This is mainly due to fact that writing text on android is much slower (about 2,3 characters per second and 2 seconds pause each step). Maybe we could leave this as a general test and note to do this when (or even if, im still not sure how to) I manage to speed the text up. 

What do you think about this?

.maestro/enrichedMarkdownText/flows/table_test.yaml
eszlamczyk: Well markdown (or at least all engines I found are suposed to not parse trailing symbols)
for example:
```
| col | col2 |
| -- | -- |
| row | row | | || test ||
```
In this example everything after second `row` should not be visible, this test is to check if for example we don't render `row | | || test |` in the cell , or we don't get additional column

.maestro/enrichedMarkdownText/flows/code_block_test.yaml
eszlamczyk: Alright, will be added in future pr

.maestro/enrichedMarkdownText/screenshots/android/image_display.png
eszlamczyk: Hmm looks like an android bug

eszlamczyk: Between android and IOS there is difference in image rounding, not sure if this is a "bug" or just a "feature" to be honest

Image

.maestro/enrichedMarkdownText/flows/mixed_nested_list_test.yaml
hryhoriiK97: @eszlamczyk yes, makes sense. Let's leave it as is for now.

.maestro/enrichedMarkdownText/screenshots/android/image_display.png
hryhoriiK97: @eszlamczyk I think we should align rounding between iOS and Android. Can you also create a ticket for this?

hryhoriiK97 (approved):

eszlamczyk:
Looks like there is another issue with Images

This is propably caused by a race condition in Image rendering (async job finishes before main loop when image is being provided very fast for example from assets/metro/localy) And it gets overriden by further code


https://github.com/user-attachments/assets/b695b977-54ce-4d2d-af0b-a55d8f340f33

PR #306, opened 2026-05-08
test: introduce better test structure and create inline elements tests

eszlamczyk:
### What/Why?
Sets up test structure better for integrating multiple test suites (for example: only block elements, only inline elements, only basic tests, only advanced tests)

Additionaly adds new "raw markdown button" to playground to better reflect real markdown not the Enriched Input -> enriched Markdown pipeline"

### Testing
`yarn test:e2e:mobile`

### PR Checklist

- [x] E2E tests are passing


.maestro/enrichedMarkdownText/flows/basic/inline_elements/bold_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+- launchApp
+
+- runFlow:
+    file: '../../../../subflows/move_to_playground.yaml'
+
+- runFlow:
+    file: '../../../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        **Bold text** and normal text and **more bold**.
```
hryhoriiK97: I think we should also add double underscores for bold: `__bold__`

.maestro/enrichedMarkdownText/flows/basic/inline_elements/italic_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+- launchApp
+
+- runFlow:
+    file: '../../../../subflows/move_to_playground.yaml'
+
+- runFlow:
+    file: '../../../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        *Italic text* and normal text and _also italic_.
```
hryhoriiK97, line 17: We should remember the md4cFlags prop — when we set underline to true, the underscore will work only for underline, and for italic we will only have `* *`.

I think we also should take this into account in tests.

.maestro/enrichedMarkdownText/flows/basic/inline_elements/link_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+- launchApp
+
+- runFlow:
+    file: '../../../../subflows/move_to_playground.yaml'
+
+- runFlow:
+    file: '../../../subflows/set_enriched_text_value.yaml'
+    env:
+      VALUE: |
+        Visit https://reactnative.dev for docs.
```
hryhoriiK97: In this case, `https://reactnative.dev` is an autolink. We can have tests for autolinks, but for CommonMark links, I think something like this would be better: `[react native](https://reactnative.dev)`.

hryhoriiK97 (commented):
Overall, LGTM! I've left a few comments.

Also, I think we should add nested combination tests, such as: `**bold _and italic_**`, code inside bold, and `[link with **bold**](url)` etc.

Additionally, I noticed that tests for superscripts and subscripts are missing, please add them to this PR.

eszlamczyk:
> Also, I think we should add nested combination tests, such as: **bold _and italic_**, code inside bold, and [link with **bold**](url) etc.
> 
> Additionally, I noticed that tests for superscripts and subscripts are missing, please add them to this PR.

I wanted to add this to the "advanced" tests, so we don't have to run theese every time

.maestro/enrichedMarkdownText/flows/basic/inline_elements/bold_test.yaml
eszlamczyk: Alright, I'll need to add new button to playground, but this shouldn't be an issue

.maestro/enrichedMarkdownText/flows/basic/inline_elements/italic_test.yaml
eszlamczyk: > Alright, I'll need to add new button to playground, but this shouldn't be an issue

.maestro/enrichedMarkdownText/flows/basic/inline_elements/link_test.yaml
eszlamczyk: This does not work, due to it being inputed via EnrichedMarkdownInput,

Although this might be a bug... (because it renders correctly on android)

hryhoriiK97: Okay, nice 😄  Can you create a ticket for this, please?

hryhoriiK97 (approved):

PR #315, opened 2026-05-11
test: introduce configs to e2e tests

eszlamczyk:
### What/Why?
Improves the E2E tests by introducing a tag/config system for filtering test runs

- Added smoke and advanced Maestro config files
- Introduced tags - `basic tests -> smoke`, `more robust "run sometimes" tests -> advanced', tags for specific type of markdown (`inline`, `block`)
- Extended run-tests.sh with --config, --include-tags, and --exclude-tags flag support.
- Added `--config` tests to run script in package.json
- Updated CONTRIBUTING.md to document the naming pattern to new script

### Testing
Run a specific config subset:          
```                   
yarn test:e2e:smoke:ios
yarn test:e2e:advanced:android
```
Or use `--include-tags` / `--exclude-tags` directly via the   
script for ad-hoc filtering.

### PR Checklist

- [x] Code compiles and runs on iOS
- [x] Code compiles and runs on Android
- [x] Updated documentation/README if applicable
- [x] Ran example app to verify changes
- [x] E2E tests are passing
- [x] Required E2E tests have been added (if applicable)

PR #316, opened 2026-05-11
test: introduce input elements test

eszlamczyk:
### What/Why?
Introduced E2E tests for `EnrichedMarkdownInput`. To make that possible added Ids to `LinkModal`

Every test should be correctly tagged according to conventions from #315 

This should cover every basic InputElementTest and the more obvious ones. Additional Tests to consider in future pull requests:
- Creating example text:
**bold** not bold
- Moving cursor to the **bold** part
- Checking if button is light up

### Testing
To run every test within this flow

```
yarn test:e2e:mobile ./.maestro/enrichedMarkdownInput/flows
```

### PR Checklist

- [x] E2E tests are passing


hryhoriiK97 (commented):
Overral LGTM! 

A few things which I think should be addressed:
1. Can we have better flow organization? Right now we only have inline styles for input, but we'll have block elements in the future. Let's organize into subfolder for inline_elements, similarly to how we have it for text.
2. All toggle-off tests are tagged as advanced — shouldn't we include at least one in smoke? 
3. The link_test only covers manual link insertion via the toolbar/modal. Should we also add a test for auto-link detection (typing a URL followed by a space and verifying it's detected/styled)?

eszlamczyk:
Good points! I'll add them

hryhoriiK97 (approved):
Overall LGTM!

A few considerations before merging:

- All other toggle-off tests (`italic_toggle_off`, `strikethrough_toggle_off`, `underline_toggle_off`, `spoiler_toggle_off`) are in advanced/ and tagged advanced. However, `bold_toggle_off_test.yaml` is in basic/ and tagged smoke. Maybe we should keep them all as basic with the smoke tag? 🤔 
- The existing `enrichedMarkdownText` tests have a comment like: `# checks the parsing of Markdown blockquotes (> text)` . I think we should decide: do we want to keep the comments or remove them all to be consistent?

hryhoriiK97 (approved):

PR #318, opened 2026-05-12
chore: bump react-native to 0.85.3 in example app

hryhoriiK97:
### What/Why?
Bump React Native from 0.84.1 to 0.85.3 in the example app. 

#### Screenshots

| iOS | Android |
| - | - |
| <video src="https://github.com/user-attachments/assets/034d79f9-989d-4362-8888-b238c7e3e658" width=300 /> | <video src="https://github.com/user-attachments/assets/79f72027-2e4d-476b-8f40-ad6a1b24c668" width=300 /> |

### PR Checklist

- [ ] E2E tests are passing


apps/example/Gemfile
```diff
@@ -14,3 +14,4 @@ gem 'bigdecimal'
 gem 'logger'
 gem 'benchmark'
 gem 'mutex_m'
+gem 'nkf'
```
eszlamczyk, line 17: question: why do we need new gem?

eszlamczyk (approved):
LGTM

(run E2Es if possible)

apps/example/Gemfile
hryhoriiK97: Required by upgrade-helper - https://react-native-community.github.io/upgrade-helper/?from=0.84.1&to=0.85.3#RnDiffApp-Gemfile

Ruby 3.4.0 removed several libraries from the standard library, and nkf is one of them.

Screenshot 2026-05-12 at 15 23 56

PR #324, opened 2026-05-15
fix(ios): reset state in prepareForRecycle on both host components

rhnmht30:
### What/Why?

When a chat / feed UI virtualises its rows (FlatList, FlashList, etc.), Fabric pools the host `RCTViewComponentView` instances and reuses them for new mounts. Both `EnrichedMarkdown` (GFM container) and `EnrichedMarkdownText` (commonmark) keep mutable state across that boundary that the new mount has no way to clear before its own props arrive:

- `EnrichedMarkdown` holds `_segmentViews`, `_segmentSignatures`, the cached / rendered markdown strings, the `ENRMTailFadeInAnimator` instances attached to each segment text view via `objc_setAssociatedObject`, the streaming flags and the dirty bitset.
- `EnrichedMarkdownText` holds `_fadeAnimator`, `_previousTextLength`, the cached / rendered markdown, the streaming flag, the accessibility cache and the text view contents.

Concrete consequences I hit while wiring this up against a chat list:

- The reconciler reuses subviews from the previous message because their signatures happen to match the new content, and the new content's leading bytes briefly show the prior message's typography until the new render commits.
- The tail-fade animator that was mid-animation when the host was pooled resumes against the new attributed string and fades a slice that has no semantic relationship to the appended tail.
- `_renderedMarkdown` from the previous mount short-circuits `hasRenderedMarkdown:` for the new content if the strings happen to match, skipping a render.

Override `prepareForRecycle` on both host components to drop subviews, cancel any per-text-view tail-fade animators, clear the cached / rendered markdown, reset the streaming flags and invalidate the accessibility cache before calling `super`.

Android and macOS are unaffected. Android does not opt views into Fabric's recycle pool (no `setupViewRecycling`) and already releases per-instance state via `onDropViewInstance` on the view managers. macOS uses the same `RCTViewComponentView` shape; the reset is platform-neutral and the helpers used (`ENRMSetAttributedText`, the `objc_setAssociatedObject` clear) are available there too.

### Testing

- Verified the host pod compiles cleanly on an iOS simulator build (Xcode 26.2 SDK, Debug).
- `yarn typecheck`, `yarn lint`, `yarn test` green.
- Did not exercise the example app's E2E suite locally for this PR — the change is reset-only and has no rendering surface; happy to run it on request if you'd prefer.

### PR Checklist

- [x] Code compiles and runs on Android (no change)
- [ ] E2E tests are passing

### DIFF
--- a/ios/EnrichedMarkdown.mm
+++ b/ios/EnrichedMarkdown.mm
@@ -768,6 +768,8 @@ static char kENRMSegmentFadeAnimatorKey;
   // segment subviews, signatures, cached markdown and any running tail-fade
   // animators from the previous mount would leak into the next mount and the
   // reconciler would reuse stale views against unrelated content.
+  [_renderCoordinator invalidate];
+
   for (RCTUIView *segment in _segmentViews) {
     if ([segment isKindOfClass:[EnrichedMarkdownInternalText class]]) {
       EnrichedMarkdownInternalText *textSegment = (EnrichedMarkdownInternalText *)segment;
--- a/ios/EnrichedMarkdown.mm
+++ b/ios/EnrichedMarkdown.mm
@@ -762,6 +762,34 @@ static char kENRMSegmentFadeAnimatorKey;
   }
 }
 
+- (void)prepareForRecycle
+{
+  // Fabric pools `RCTViewComponentView` instances. Without resetting here, the
+  // segment subviews, signatures, cached markdown and any running tail-fade
+  // animators from the previous mount would leak into the next mount and the
+  // reconciler would reuse stale views against unrelated content.
+  for (RCTUIView *segment in _segmentViews) {
+    if ([segment isKindOfClass:[EnrichedMarkdownInternalText class]]) {
+      ENRMTailFadeInAnimator *animator =
+          objc_getAssociatedObject(((EnrichedMarkdownInternalText *)segment).textView, &kENRMSegmentFadeAnimatorKey);
+      [animator cancel];
+      objc_setAssociatedObject(((EnrichedMarkdownInternalText *)segment).textView, &kENRMSegmentFadeAnimatorKey, nil,
+                               OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+    }
+    [segment removeFromSuperview];
+  }
+  [_segmentViews removeAllObjects];
+  [_segmentSignatures removeAllObjects];
+
+  _cachedMarkdown = nil;
+  _renderedMarkdown = nil;
+  _streamingAnimation = NO;
+  _tableStreamingMode = ENRMTableStreamingModeHidden;
+  _dirtyFlags = ENRMDirtyNone;
+
+  [super prepareForRecycle];
+}
+
 Class<RCTComponentViewProtocol> EnrichedMarkdownCls(void)
 {
   return EnrichedMarkdown.class;
--- a/ios/EnrichedMarkdown.mm
+++ b/ios/EnrichedMarkdown.mm
@@ -770,10 +770,10 @@ static char kENRMSegmentFadeAnimatorKey;
   // reconciler would reuse stale views against unrelated content.
   for (RCTUIView *segment in _segmentViews) {
     if ([segment isKindOfClass:[EnrichedMarkdownInternalText class]]) {
-      ENRMTailFadeInAnimator *animator =
-          objc_getAssociatedObject(((EnrichedMarkdownInternalText *)segment).textView, &kENRMSegmentFadeAnimatorKey);
+      EnrichedMarkdownInternalText *textSegment = (EnrichedMarkdownInternalText *)segment;
+      ENRMTailFadeInAnimator *animator = objc_getAssociatedObject(textSegment.textView, &kENRMSegmentFadeAnimatorKey);
       [animator cancel];
-      objc_setAssociatedObject(((EnrichedMarkdownInternalText *)segment).textView, &kENRMSegmentFadeAnimatorKey, nil,
+      objc_setAssociatedObject(textSegment.textView, &kENRMSegmentFadeAnimatorKey, nil,
                                OBJC_ASSOCIATION_RETAIN_NONATOMIC);
     }
     [segment removeFromSuperview];
--- a/ios/EnrichedMarkdown.mm
+++ b/ios/EnrichedMarkdown.mm
@@ -764,10 +764,6 @@ static char kENRMSegmentFadeAnimatorKey;
 
 - (void)prepareForRecycle
 {
-  // Fabric pools `RCTViewComponentView` instances. Without resetting here, the
-  // segment subviews, signatures, cached markdown and any running tail-fade
-  // animators from the previous mount would leak into the next mount and the
-  // reconciler would reuse stale views against unrelated content.
   [_renderCoordinator invalidate];
 
   for (RCTUIView *segment in _segmentViews) {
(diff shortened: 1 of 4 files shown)


rhnmht30:
@hryhoriiK97 ready when you have time. Two checklist items unchecked:

- Example app: nothing in apps/example exercises a virtualised list, so the recycle path isn't reachable from the example without adding a new screen. Happy to add a small RecycleStress story if you want one bundled in.
- E2E: the assertion is "no stale subviews after recycle" which is hard to baseline with screenshot diffs. Can add a Maestro flow that mounts two long messages alternately if you prefer.        

Manual repro on a real consumer: render EnrichedMarkdown rows inside a FlatList with removeClippedSubviews, scroll a recycled row back into view while another row is mid-stream, and the pooled host briefly shows the prior message's segments before the new render lands. Same shape on EnrichedMarkdownText.                                                                  

Open to splitting the GFM and commonmark resets into separate commits if you'd rather review them independently.

ios/EnrichedMarkdown.mm
```diff
@@ -762,6 +762,34 @@ - (void)didMoveToWindow
   }
 }
 
+- (void)prepareForRecycle
```
hryhoriiK97, line 765: What happens if _renderCoordinator already dispatched a render to its background queue when this fires? The completion block would apply stale content to a recycled view.

Can we add _renderCoordinator.blockAsyncRender = YES; at the top here? And ideally verify the coordinator's renderId check discards in-flight completions. It's related to both EnrichedMarkdown and EnrichedMarkdownText

ios/EnrichedMarkdownText.mm
```diff
@@ -522,6 +522,30 @@ - (void)didMoveToWindow
   }
 }
 
+- (void)prepareForRecycle
```
hryhoriiK97, line 525: _spoilerManager still holds text ranges from the old content after we clear the text view. Should we nil it out here or or add a `[_spoilerManager invalidate]` / reset call.? It gets recreated in the props path anyway.

ios/EnrichedMarkdown.mm
```diff
@@ -762,6 +762,34 @@ - (void)didMoveToWindow
...
+{
+  // Fabric pools `RCTViewComponentView` instances. Without resetting here, the
+  // segment subviews, signatures, cached markdown and any running tail-fade
+  // animators from the previous mount would leak into the next mount and the
+  // reconciler would reuse stale views against unrelated content.
+  for (RCTUIView *segment in _segmentViews) {
+    if ([segment isKindOfClass:[EnrichedMarkdownInternalText class]]) {
+      ENRMTailFadeInAnimator *animator =
+          objc_getAssociatedObject(((EnrichedMarkdownInternalText *)segment).textView, &kENRMSegmentFadeAnimatorKey);
```
hryhoriiK97: The cast is repeated twice — a local would clean it up:

`EnrichedMarkdownInternalText *textSegment = (EnrichedMarkdownInternalText *)segment;`

hryhoriiK97 (changes requested):
Thank you for creating this PR! I've left a few comments 🙂 

Would be great to add a FlatList screen to the example app in a follow-up PR — something with ~50 markdown rows and removeClippedSubviews so we can actually exercise the recycle path manually. Would also help catch future regressions.

ios/EnrichedMarkdown.mm
rhnmht30: Done in 86f00b3, pulled the cast into a local `textSegment` inside the `isKindOfClass:` branch.

ios/EnrichedMarkdown.mm
rhnmht30: Looked into this. `blockAsyncRender = YES` on its own does not actually discard the in-flight render, it only gates future `scheduleRender:` calls. The completion block for a render that was already dispatched still lands on main with `renderId == _currentRenderId` (nothing in `prepareForRecycle` bumps the counter), so the stale apply runs against the recycled view.

A related concern: `blockAsyncRender` is only set to `YES` today inside `renderMarkdownSynchronously`, and is never reset to `NO`. Setting it in `prepareForRecycle` would leave the recycled instance with async renders permanently blocked on its next mount.

Proposing instead to add a small helper on the coordinator that bumps the counter:

```objc
// ENRMAsyncRenderCoordinator.h
- (void)invalidate;

// ENRMAsyncRenderCoordinator.m
- (void)invalidate { ++_currentRenderId; }
```

Then call `[_renderCoordinator invalidate]` at the top of both `prepareForRecycle` paths. The existing `renderId == _currentRenderId` check at the apply site fires correctly and drops the stale completion, and the next mount goes through `scheduleRender:` normally with a fresh ID.

Happy to use a different name if `invalidate` reads wrong, e.g. `discardPendingRenders`. Let me know which you prefer and I will push the change.

ios/EnrichedMarkdownText.mm
rhnmht30: Worth flagging that `_spoilerManager` is not actually re-created via the props path on recycle: the creation in `updateProps` is gated on `if (_config == nil)`, and we do not reset `_config` in `prepareForRecycle`. So nil-ing `_spoilerManager` alone would leave the recycled instance without a spoiler manager on its next mount.

`ENRMSpoilerOverlayManager` does not have an `invalidate` method today, but it does expose `removeAllOverlays`, which removes the overlay subviews and clears the internal `_overlaysByKey` dict. The manager itself stays attached to the (reused) text view, and the next render's `setNeedsUpdate` / `updateIfNeeded` rebuilds overlays against the new content.

Proposing to call `[_spoilerManager removeAllOverlays]` in `prepareForRecycle` (before clearing the text view's attributed string). If you would rather have the manager fully recreated, I can also reset `_config = nil` here so the existing guard in `updateProps` reallocates it. Let me know which direction you prefer.

ios/EnrichedMarkdown.mm
hryhoriiK97: Good catch on the blockAsyncRender issue. The invalidate approach looks correct to me 👍

ios/EnrichedMarkdownText.mm
hryhoriiK97: Makes sense, removeAllOverlays is enough. Let's go with that 👍

ios/EnrichedMarkdown.mm
rhnmht30: Done in 7a83784. Added `-[ENRMAsyncRenderCoordinator invalidate]` and call it at the top of both `prepareForRecycle` paths.

ios/EnrichedMarkdownText.mm
rhnmht30: Done in 98fcc89. Calling `[_spoilerManager removeAllOverlays]` in `prepareForRecycle` before clearing the text view.

ios/EnrichedMarkdown.mm
```diff
@@ -762,6 +762,36 @@ - (void)didMoveToWindow
   }
 }
 
+- (void)prepareForRecycle
+{
+  // Fabric pools `RCTViewComponentView` instances. Without resetting here, the
+  // segment subviews, signatures, cached markdown and any running tail-fade
+  // animators from the previous mount would leak into the next mount and the
+  // reconciler would reuse stale views against unrelated content.
```
hryhoriiK97: Let's clean up this comment, please - same goes for EnrichedMarkdownText

hryhoriiK97 (changes requested):

ios/EnrichedMarkdown.mm
rhnmht30: Done in f4f5c50, dropped the explanatory comment from both `prepareForRecycle` methods.

hryhoriiK97 (approved):

PR #325, opened 2026-05-15
fix: skip streaming animations when Reduce Motion is enabled

rhnmht30:
### What/Why?

The streaming-animation paths in the GFM container run regardless of the system-wide accessibility preference for reduced motion:

- `ENRMTailFadeInAnimator.animateFrom:to:` always snapshots colors, sets the new tail to alpha 0 and runs a `CADisplayLink` fade up to alpha 1.
- `EnrichedMarkdown.animateBlockViewIfNeeded:` always sets `view.alpha = 0` and runs a 200ms opacity animation on new table / math segments.
- `EnrichedMarkdown.updateTableView:withSegment:` always calls `animateNewRowsFromPreviousCount:` on growing tables.
- The Android equivalents in `TailFadeInAnimator.animate` and `EnrichedMarkdown.animateBlockViewFadeIn` / `animateNewRows` have the same shape.

Users who have enabled \"Reduce Motion\" on iOS or \"Remove animations\" on Android explicitly opt out of non-essential animations. The current implementation produces visible fades for them, which is both an accessibility regression (against `UIAccessibilityIsReduceMotionEnabled` / `AccessibilityManager.isAnimatorEnabled`) and a perceptual mismatch with how UIKit and AppKit's own widgets behave under the same preference.

Gate the three iOS sites and three Android sites on the platform flag. When it's set, the new content is shown at its final opacity with no transition.

To keep call sites short, a small helper is added on each platform:

- iOS: `ENRMShouldReduceMotion()` in `ENRMUIKit.h`, which folds `UIAccessibilityIsReduceMotionEnabled` and the macOS `NSWorkspace.accessibilityDisplayShouldReduceMotion` equivalent behind one call.
- Android: `isReducedMotionEnabled(context)` in `utils/common/MotionPreferences.kt`, using `AccessibilityManager.isAnimatorEnabled` on API 26+ and falling back to allowing animations on older API levels.

The flag is read on each animation entry rather than cached on the host. macOS observers (`NSWorkspace.accessibilityDisplayOptionsDidChangeNotification`) could update a cached value reactively, but the current call sites only fire on render boundaries, so the polled read is cheap and avoids extra notification plumbing.

### Testing

- Verified the iOS host pod compiles cleanly on an iOS simulator build (Xcode 26.2 SDK, Debug) with the helper and gates applied.
- `yarn typecheck`, `yarn lint`, `yarn test` green.
- Manual: with Reduce Motion enabled in the iOS Simulator's Accessibility settings, no fades are triggered when streaming markdown that contains a paragraph, a table that grows row-by-row, and a block math segment; with the setting off, animations behave as before.

### PR Checklist

- [x] Code compiles and runs on Android (no JVM-level changes besides the new helper and three callers)
- [ ] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/utils/text/TailFadeInAnimator.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/utils/text/TailFadeInAnimator.kt
@@ -7,6 +7,7 @@ import android.text.Spannable
 import android.view.animation.LinearInterpolator
 import android.widget.TextView
 import com.swmansion.enriched.markdown.spans.FadeInSpan
+import com.swmansion.enriched.markdown.utils.common.isReducedMotionEnabled
 import java.lang.ref.WeakReference
 
 class TailFadeInAnimator(
@@ -24,6 +25,9 @@ class TailFadeInAnimator(
 
     val textView = viewRef.get() ?: return
     val spannable = textView.text as? Spannable ?: return
+
+    // Honor the system accessibility preference and leave the new tail visible.
+    if (isReducedMotionEnabled(textView.context)) return
 
     val fadeSpan = FadeInSpan().apply { alpha = 0f }
     spannable.setSpan(fadeSpan, tailStart, tailEnd, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
--- a/android/src/main/java/com/swmansion/enriched/markdown/utils/text/TailFadeInAnimator.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/utils/text/TailFadeInAnimator.kt
@@ -26,7 +26,6 @@ class TailFadeInAnimator(
     val textView = viewRef.get() ?: return
     val spannable = textView.text as? Spannable ?: return
 
-    // Honor the system accessibility preference and leave the new tail visible.
     if (isReducedMotionEnabled(textView.context)) return
 
     val fadeSpan = FadeInSpan().apply { alpha = 0f }
--- a/ios/utils/ENRMTailFadeInAnimator.m
+++ b/ios/utils/ENRMTailFadeInAnimator.m
@@ -49,8 +49,6 @@ typedef struct {
     return;
 
 #if !TARGET_OS_OSX
-  // Honor the system-wide accessibility preference: when Reduce Motion is on
-  // we leave the new tail at full alpha instead of running the fade.
   if (UIAccessibilityIsReduceMotionEnabled()) {
     return;
   }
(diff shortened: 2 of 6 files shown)


rhnmht30:
@hryhoriiK97 ready when you have time. A couple of notes:
- macOS: the helper uses NSWorkspace.accessibilityDisplayShouldReduceMotion (10.12+) for symmetry with the existing macOS branches in ENRMUIKit.h. Gated on `@available` so older targets degrade to "no reduction". 
- E2E unchecked: gating animations on a system flag isn't easy to assert through screenshot diffs. Manual repro is Settings > Accessibility > Motion > Reduce Motion (iOS) or Settings > Accessibility > Remove animations (Android), then stream a message with prose, a growing table, and a block math segment. Without the gate the cells and the new tail fade in; with it they appear at final alpha.

Helper is in ENRMUIKit.h and utils/common/MotionPreferences.kt so it can be reused if other future code paths need the same check.

android/src/main/java/com/swmansion/enriched/markdown/utils/text/TailFadeInAnimator.kt
```diff
@@ -25,6 +26,9 @@ class TailFadeInAnimator(
     val textView = viewRef.get() ?: return
     val spannable = textView.text as? Spannable ?: return
 
+    // Honor the system accessibility preference and leave the new tail visible.
```
hryhoriiK97: I think we can remove this comment, the logic is clean and self-explanatory.

ios/utils/ENRMTailFadeInAnimator.m
```diff
@@ -48,6 +48,14 @@ - (void)animateFrom:(NSUInteger)tailStart to:(NSUInteger)tailEnd
   if (!storage || tailEnd <= tailStart || tailEnd > storage.length)
     return;
 
+#if !TARGET_OS_OSX
+  // Honor the system-wide accessibility preference: when Reduce Motion is on
+  // we leave the new tail at full alpha instead of running the fade.
```
hryhoriiK97: The same for this comment - we can remove it

ios/utils/ENRMUIKit.h
```diff
@@ -162,6 +162,22 @@ static inline void ENRMSetContentSize(ENRMPlatformTextView *textView, CGSize siz
...
+static inline BOOL ENRMShouldReduceMotion(void)
+{
+#if !TARGET_OS_OSX
+  return UIAccessibilityIsReduceMotionEnabled();
+#else
+  if (@available(macOS 10.12, *)) {
+    return NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion;
+  }
+  return NO;
+#endif
```
hryhoriiK97, line 178: when macOS animation is eventually implemented here, it'll need its own NSWorkspace.sharedWorkspace.accessibilityDisplayShouldReduceMotion check. Maybe worth a small TODO note next to the existing one.

hryhoriiK97 (changes requested):
Thank you for the fix! I've left a few small comments 🙂

hryhoriiK97 (approved):
LGTM! Thank you for the contribution 🙌 This fix will be available in tomorrow's nightly release 🙂

PR #326, opened 2026-05-15
Remove maximum line height setting in ParagraphStyle

Gregoirevda:
### What/Why?

respects the user line height, but avoids cutting content and making each line dynamic.

### Testing

`\\displaystyle\\frac{4/2}`

### DIFF
--- a/ios/utils/ParagraphStyleUtils.m
+++ b/ios/utils/ParagraphStyleUtils.m
@@ -118,7 +118,6 @@ void applyLineHeight(NSMutableAttributedString *output, NSRange range, CGFloat l
   NSMutableParagraphStyle *style = getOrCreateParagraphStyle(output, range.location);
 
   style.minimumLineHeight = lineHeight;
-  style.maximumLineHeight = lineHeight;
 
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (commented):
@Gregoirevda thanks for flagging this!

However, we intentionally set both `minimumLineHeight` and `maximumLineHeight` to match how React Native's `<Text>` handles `lineHeight` on iOS (see [RCTAttributedTextUtils.mm](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/react/renderer/textlayoutmanager/platform/ios/react/renderer/textlayoutmanager/RCTAttributedTextUtils.mm#L231-L232)). Removing `maximumLineHeight` globally would break baseline offset calculations and cause inconsistent line spacing for all text, not just math.

Could you open an issue describing the problem instead? We'd like to solve this in a more targeted way without affecting normal text layout.

Gregoirevda:
@hryhoriiK97 thank you for your reply. I created an issue https://github.com/software-mansion-labs/react-native-enriched-markdown/issues/337

Issue #328, opened 2026-05-16
Web parser silently drops content / crashes on inputs around ~64 KB


bobbyg603:
### Before submitting a new issue

- [x] I tested using the latest version of the library (`0.5.0`).
- [x] N/A — this bug is in the web (WASM) parser, exercised without React Native.
- [x] I checked for possible duplicate issues.

### Bug summary

`EnrichedMarkdownText` on web feeds its input to the bundled `md4c` WASM module (`lib/module/web/wasm/md4c.js`). For inputs around or above ~64 KB the module fails in two ways depending on size:

1. **Silent truncation** (~65,000 – ~69,000 bytes): the WASM call returns `{"type":"Document"}` with no `children`, i.e. the parsed AST is empty. No exception is thrown — the React component sees a "successful" parse and renders nothing, with no signal to the consumer that data was dropped.
2. **`RuntimeError: memory access out of bounds`** (≳70,000 bytes): the WASM module aborts. This surfaces in `EnrichedMarkdownText` as `[EnrichedMarkdownText] Parse failed: RuntimeError: memory access out of bounds` and the component falls back to its parse-error state.

The threshold is content-independent — plain ASCII (`'a'.repeat(N)`) reproduces both modes — so this isn't a markdown-feature edge case.

Originally hit when rendering a ~108 KB user-uploaded markdown glossary; the component rendered nothing and only this console message appeared:

```
[EnrichedMarkdownText] Parse failed: RuntimeError: memory access out of bounds
    at wasm://wasm/0006298a:wasm-function[6]:0x4ff
    at ccall (.../node_modules/react-native-enriched-markdown/lib/module/web/wasm/md4c.js)
    at parseMarkdown (.../node_modules/react-native-enriched-markdown/lib/module/web/parseMarkdown.js)
```

### Library version

0.5.0

### Environment info

```
react-native-enriched-markdown: 0.5.0
Reproducible in pure Node (24.14) using the package's bundled WASM; the same
code path runs in the browser via react-native-web. The repro does not require
React Native because the bug is in the WASM module itself, not the React
wrapper.
```

### Steps to reproduce

Standalone repro repository: **https://github.com/bobbyg603/md4c-wasm-64kb-repro**

```sh
git clone https://github.com/bobbyg603/md4c-wasm-64kb-repro
cd md4c-wasm-64kb-repro
npm install
node repro.cjs
```

[...]

PR #329, opened 2026-05-16
fix(web): raise WASM stack size + enable memory growth

bobbyg603:
## Why

The current `cpp/wasm/build.sh` uses Emscripten defaults, giving the WASM module a **64 KB stack** with no memory growth. md4c parses its input on stack-allocated structures, so any markdown input around or above ~64 KB triggers one of two failures on `EnrichedMarkdownText` for web:

- **~65–69 KB**: the WASM call returns `{"type":"Document"}` with no children — the parse silently drops the entire input. No exception is thrown, so the component renders nothing without any signal to the consumer that data was lost.
- **70 KB+**: `RuntimeError: memory access out of bounds`.

Both modes are content-independent (plain `'a'.repeat(N)` reproduces them, see the linked repo). Native iOS / Android paths aren't affected because they don't go through Emscripten.

Tracking issue with full investigation and repro: #328
Standalone reproducer: https://github.com/bobbyg603/md4c-wasm-64kb-repro

## What

Add four flags to the link step of `cpp/wasm/build.sh`:

```
-s STACK_SIZE=8MB
-s INITIAL_MEMORY=16MB
-s MAXIMUM_MEMORY=512MB
-s ALLOW_MEMORY_GROWTH=1
```

These values are conservative — they leave plenty of headroom for realistic documents while keeping `INITIAL_MEMORY` modest (memory can grow up to the maximum on demand). Happy to tune the specific numbers in review.

The regenerated `src/web/wasm/md4c.js` is included in the commit.

## After

Same reproducer that fails on `0.5.0` now returns a populated `Document` at every size we tested:

```
  64,000 bytes -> JSON  64095 bytes, Document.children=1
  65,000 bytes -> JSON  65095 bytes, Document.children=1
  65,260 bytes -> JSON  65355 bytes, Document.children=1
  70,000 bytes -> JSON  70095 bytes, Document.children=1
 128,000 bytes -> JSON 128095 bytes, Document.children=1
```

A real-world 108 KB markdown glossary that previously rendered as a blank component now parses in ~11 ms and renders correctly.

## Test plan

- [ ] CI WASM build (`yarn build:wasm`) succeeds against the Emscripten image.
- [ ] Existing web example continues to render normally for small inputs.
- [ ] An input >64 KB now renders content instead of an empty fallback (or `RuntimeError` for >70 KB).


bobbyg603:
Full disclosure, this was driven primarily by Claude Code with Opus 4.7 xhigh. However, I reviewed everything and to the best of my knowledge it seems accurate and correct. The initial problem occurred when I tried to load a 108 KB markdown file preview. This resulted in a blank preview and the following stack trace:

```
[EnrichedMarkdownText] Parse failed: RuntimeError: memory access out of bounds
    at wasm://wasm/0006298a:wasm-function[6]:0x4ff
    at ccall (eval at <anonymous> (http://localhost:8081/frontend/node_modules/expo-router/entry.bundle?platform=web&dev=true&hot=false&lazy=true&transform.routerRoot=app:84964:16), <anonymous>:371:19)
    at eval (eval at <anonymous> (http://localhost:8081/frontend/node_modules/expo-router/entry.bundle?platform=web&dev=true&hot=false&lazy=true&transform.routerRoot=app:84964:16), <anonymous>:385:29)
    at parseMarkdown (http://localhost:8081/frontend/node_modules/expo-router/entry.bundle?platform=web&dev=true&hot=false&lazy=true&transform.routerRoot=app:323106:31)
    at async Promise.all (index 0)

Call Stack
  at addLog (node_modules/.bun/@expo+log-box@55.0.11+644225bbac4d1ea0/node_modules/@expo/log-box/src/Data/LogBoxData.tsx:205:30)
  at consoleErrorMiddleware (node_modules/.bun/@expo+log-box@55.0.11+644225bbac4d1ea0/node_modules/@expo/log-box/src/LogBox.ts:103:14)
  at console.error (node_modules/.bun/@expo+log-box@55.0.11+644225bbac4d1ea0/node_modules/@expo/log-box/src/LogBox.ts:33:25)
  at Promise.all.then._catch$argument_0 (node_modules/.bun/react-native-enriched-markdown@0.5.0+3799d304aee0924e/node_modules/react-native-enriched-markdown/lib/module/web/EnrichedMarkdownText.js:49:19)
```

I checked the [reproduction example](https://github.com/bobbyg603/md4c-wasm-64kb-repro) and it passed my sniff test. I also [published the proposed fix to npmjs](https://www.npmjs.com/package/@bobbyg603/react-native-enriched-markdown) and the initial results look promising. I'm currently able to load the file and don't see any noticeable lag or hitches, but I'll admit I haven't done much testing up near the max memory limit.

hryhoriiK97 (approved):
LGTM! @bobbyg603 thank you for the fix! It will be available in tomorrow's nightly release.

PR #333, opened 2026-05-18
fix(android): use value animators are animators enabled for reduced motion

eszlamczyk:
### What/Why?
`MotionPreferences.kt` referenced `AccessibilityManager.isAnimatorEnabled` which does not exist in the Android SDK, causing a compilation failure on Android.
                                                                             
Replaced with ValueAnimator.areAnimatorsEnabled() (API 26+), the correct   
platform API for detecting whether system animations are enabled. As a
bonus, this also covers Battery Saver mode, which disables animations      
system-wide and wasn't previously accounted for.

### Testing
Due to some other bug, android does not display animation character by character, it does work (and uses the same code) for table row animation.

In developer mode:
Set Animator duration scale in Developer Options to:                       
- 1x → table fade-in animations play during streaming
- Animation off (0x) → table fade-in animations are skipped, content       
appears instantly

##### Animations on 
https://github.com/user-attachments/assets/f28153dd-abf7-4ecc-afe4-11ca88517499

##### Animations off 

https://github.com/user-attachments/assets/939efb84-0c00-4bd0-85ee-7abcbec91361

### PR Checklist

- [x] ~Code compiles and runs on iOS~
- [x] ~E2E tests are passing~
- [x] ~Required E2E tests have been added (if applicable)~

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/utils/common/MotionPreferences.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/utils/common/MotionPreferences.kt
@@ -1,21 +1,18 @@
 package com.swmansion.enriched.markdown.utils.common
 
+import android.animation.ValueAnimator
 import android.content.Context
 import android.os.Build
-import android.view.accessibility.AccessibilityManager
 
 /**
  * Returns true when the user has asked the system to minimise motion.
  *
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (approved):
LGTM! Thanks for the fix 🙌

PR #336, opened 2026-05-18
fix(ci): include library native sources in turbo cache inputs

hryhoriiK97:
### What/Why?
Turbo was serving stale cache hits for build:android and build:ios because inputs only tracked files relative to apps/example, missing the library's android/, ios/, cpp/, and src/ directories at the workspace root. Use `$TURBO_ROOT$` to reference library sources and update the GitHub Actions cache key to hash native source files.

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):
LGTM

PR #355, opened 2026-05-26
feat(android): replace math engine with RaTeX

eszlamczyk:
### What/Why?
Introduces new math engine - `RaTeX` - replacing `Android math`.

AndroidMath was unmaintained, required JitPack, and shipped `libc++_shared.so` which caused duplicate native library conflicts in instrumentation builds (Detox). It also used reflection-heavy metric extraction for baseline alignment of inline math.

##### What changed:
- build.gradle - swaps `com.github.gregcockroft:AndroidMath` for `io.github.erweixin:ratex-android:0.1.9`, removes the `JitPack` repository, and removes the packaging { pickFirsts } workarounds that existed solely for AndroidMath
- `consumer-rules.pro` - updates ProGuard keep rules from `com.agog.mathdisplay.**` / `com.pvporbit.freetype.**` to `io.ratex.**`
- MathInlineSpan - replaces `MTMathView` + reflection-based ascent/descent extraction with `RaTeXEngine.parseBlocking` + `RaTeXRenderer` drawn to a Bitmap; removes the reflection companion entirely
- MathMeasureHelper - drops the `CountDownLatch/main-thread` dispatch (no longer needed since RaTeX is allegedly thread-safe); renamed from `measureOnMainThread` to `measure`
- MathContainerView - replaces `MTMathView` with a lightweight inline View that renders synchronously via `RaTeXEngine.parseBlocking`

### Testing

Start example app and try different markdown combinations via set markdown button (in playground - can be done one by one or via a combination):
```
$$E = mc^2$$

$$\psi(x, t) = A e^{i(kx - \omega t)}$$

$$\begin{aligned}
\nabla \cdot \mathbf{E} &= \frac{\rho}{\varepsilon_0} \\
\nabla \cdot \mathbf{B} &= 0 \\
\nabla \times \mathbf{E} &= -\frac{\partial \mathbf{B}}{\partial t} \\
\nabla \times \mathbf{B} &= \mu_0\mathbf{J} + \mu_0\varepsilon_0\frac{\partial \mathbf{E}}{\partial t}
\end{aligned}$$

$$\mathbf{R}_z(\theta) = \begin{bmatrix} 
\cos\theta & -\sin\theta & 0 \\ 
\sin\theta & \cos\theta & 0 \\ 
0 & 0 & 1 
\end{bmatrix}$$

$$f'(x) = \lim_{h \to 0} \frac{f(x+h) - f(x)}{h}$$

$$\int_{-\infty}^{\infty} e^{-x^2} \, dx = \sqrt{\pi}$$

$$f(n) = \begin{cases} 
\frac{n}{2} & \text{if } n \text{ is even} \\ 
3n + 1 & \text{if } n \text{ is odd} 
\end{cases}$$

$$aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa$$

$a^2$

Random text with random $inline$ and $$a^2$$ 
```
Additionaly change `apps/example/src/screens/playground/PlaygroundScreeen.tsx`'s `MARKDOWN_STYLE` fields (line 50) to include `math` and `inlineMath` stylings, for example:
```ts
math: {
  color: '#1e40aa',
  backgroundColor: '#eff6ff',
  padding: 12,
  marginTop: 8,
  marginBottom: 8,
  textAlign: 'center' as const,
},
inlineMath: {
  color: '#7c3aed',
},
```

### PR Checklist

- [x] E2E tests are passing


android/src/math/java/com/swmansion/enriched/markdown/views/MathContainerView.kt
```diff
@@ -80,7 +99,15 @@ class MathContainerView(
 
   fun applyLatex(latex: String) {
     cachedLatex = latex
-    mathView.latex = latex
+    try {
+      val dl = RaTeXEngine.parseBlocking(latex, displayMode = true, color = mathStyle.color)
+      mathView.renderer = RaTeXRenderer(dl, mathStyle.fontSize) { RaTeXFontLoader.getTypeface(it) }
+    } catch (e: Exception) {
+      Log.e("MathContainerView", "Failed to render LaTeX", e)
```
hryhoriiK97: Let's use TAG companion constant:

```
companion object {
  private const val TAG = "MathContainerView"
  // ...
}
```

android/src/math/java/com/swmansion/enriched/markdown/views/MathContainerView.kt
```diff
@@ -80,7 +99,15 @@ class MathContainerView(
 
   fun applyLatex(latex: String) {
     cachedLatex = latex
-    mathView.latex = latex
+    try {
+      val dl = RaTeXEngine.parseBlocking(latex, displayMode = true, color = mathStyle.color)
```
hryhoriiK97: nit: I know there can be a lot of shorthand naming in the codebase (and this is a good example) - we definitely want to eliminate this. If you see something like this, feel free to suggest a clearer name.

dl => displayList

android/src/math/java/com/swmansion/enriched/markdown/views/MathContainerView.kt
```diff
@@ -3,54 +3,73 @@ package com.swmansion.enriched.markdown.views
...
     when (mathStyle.textAlign) {
-      "left" -> MTMathView.MTTextAlignment.KMTTextAlignmentLeft to Gravity.START
-      "right" -> MTMathView.MTTextAlignment.KMTTextAlignmentRight to Gravity.END
-      else -> MTMathView.MTTextAlignment.KMTTextAlignmentCenter to Gravity.CENTER_HORIZONTAL
+      "left" -> Gravity.START
+      "right" -> Gravity.END
+      else -> Gravity.CENTER_HORIZONTAL
+    }
+
+  private val mathView =
```
hryhoriiK97: I think we can extract this anonymous View(context) object into a small, named class. Something like this:

```
private class RaTeXCanvasView(context: Context) : View(context) {
  var renderer: RaTeXRenderer? = null

  override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
    val currentRenderer = renderer
      ?: return setMeasuredDimension(0, 0)
    setMeasuredDimension(
      currentRenderer.widthPx.toInt().coerceAtLeast(1),
      currentRenderer.totalHeightPx.toInt().coerceAtLeast(1),
    )
  }

  override fun onDraw(canvas: Canvas) {
    renderer?.draw(canvas)
  }
}
```

android/src/math/java/com/swmansion/enriched/markdown/spans/MathMeasureHelper.kt
```diff
@@ -1,84 +1,37 @@
...
-        labelMode =
-          when (request.mode) {
-            MathRenderMode.Display -> MTMathView.MTMathViewMode.KMTMathViewModeDisplay
-            else -> MTMathView.MTMathViewMode.KMTMathViewModeText
-          }
-        textAlignment = MTMathView.MTTextAlignment.KMTTextAlignmentLeft
-        fontSize = request.fontSize
-        latex = request.latex
-      }
+    RaTeXFontLoader.ensureLoaded(context)
```
hryhoriiK97: `ensureLoaded` is called inside `measureSingle`, which is invoked once per request inside the .map. For a message with 10 inline math spans, this results in 10 redundant calls.

We can move it to the top of measure.

Doing this also removes the need to pass context to `measureSingle`, since it's currently only required for ensureLoaded.

.maestro/enrichedMarkdownText/screenshots/android/paragraph_math_combo_display.png
hryhoriiK97, line 1: The 2a denominator in the quadratic formula screenshot is clipped at the bottom. This is caused by .toInt() truncating fractional pixel dimensions (e.g. 87.6 becomes 87px). Replace with ceil(...).toInt() (from kotlin.math.ceil) in `MathInlineSpan.kt`, `MathContainerView.kt`, and `MathMeasureHelper.kt` wherever widthPx / totalHeightPx are converted to int - this ensures sub-pixel content isn't cut off.

Image

hryhoriiK97 (commented):
LGTM! This is a great change 🙌 I've left a few comments - please take a look at them 🙏

.maestro/enrichedMarkdownText/screenshots/android/paragraph_math_combo_display.png
eszlamczyk: Great find, thanks!

hryhoriiK97 (approved):

eszlamczyk:
Note, bumping `RaTeX` in this pr as new version has important [fix](https://github.com/erweixin/RaTeX/issues/97)

eszlamczyk:
| Configuration | Size (MB) | Size (KB) |
| :--- | ---: | ---: |
| ratex + math=true | 34MB | 35,326.06 KB |
| ratex + math=false | 31MB | 31,807.24 KB |
| android math + math=true | 33MB | 34,014.68 KB |
| android math + math=false | 31MB | 31,807.24 KB |

PR #356, opened 2026-05-26
fix(plugin): use named imports from @expo/config-plugins so the plugin loads under tsx/cjs

### LINKED ISSUE #192: Error when using Expo config plugin
**Describe the bug**
When using the expo config plugin like this: 
```txt
[
  "react-native-enriched-markdown",
   {
      enableMath: false,
   },
],
```

The following error occurs:
```txt
PluginError: Unable to resolve a valid config plugin for react-native-enriched-markdown.
• main export of react-native-enriched-markdown does not appear to be a config plugin: the following error was thrown when importing /xxx/node_modules/react-native-enriched-markdown/app.plugin.js: Cannot destructure property 'withDangerousMod' of 'import_config_plugins.default' as it is undefined. 
Verify that react-native-enriched-markdown includes a config plugin. If it does not, then remove the entry from plugins in your app config file. Learn more: https://docs.expo.dev/guides/config-plugins/
```

**To Reproduce**
1. Add the expo config plugin like above to the app config
2. Run prebuild

**Expected behavior**
No error should occur.

hknakn:
## Summary

The Expo config plugin files use a default-import + destructure pattern against `@expo/config-plugins`:

```ts
import configPlugins, { type ConfigPlugin } from '@expo/config-plugins';
const { withDangerousMod } = configPlugins;
```

`@expo/config-plugins` is a pure CJS module with named exports and **no default export**. The published `lib/module/plugin/*.js` keeps these as raw ESM `import` statements, so when a consumer's `app.config.ts` uses `tsx/cjs` (a very common Expo setup), the default import resolves to `undefined` and the top-level destructure throws at module load — before the plugin's own `enableMath: false` check can run.

The trace from #192:

```
PluginError: Unable to resolve a valid config plugin for react-native-enriched-markdown.
TypeError: Cannot destructure property 'withDangerousMod' of 'import_config_plugins.default' as it is undefined.
  at fs (.../lib/module/plugin/withIosMath.js:7:3)
```

Switching to named imports compiles to direct property access — no `__importDefault` interop helper needed, no `default` lookup at runtime.

## How to reproduce

1. Expo SDK 55 project with an `app.config.ts` whose first line is `import 'tsx/cjs';`.
2. Add the plugin to `plugins`: `['react-native-enriched-markdown', { enableMath: false }]`.
3. `npx expo prebuild` → error above.

After this patch, prebuild runs through, `enableMath: false` is honoured, and the generated `ios/Podfile` and `android/gradle.properties` contain the expected `ENRICHED_MARKDOWN_ENABLE_MATH=0` / `enrichedMarkdown.enableMath=false` entries.

## Risks

None functional — the change is purely import shape. The compiled output drops a level of indirection but produces the same `withDangerousMod`/`withGradleProperties` references at call time. `yarn typecheck`, `yarn lint`, and `yarn prepare` all pass locally.

Closes #192.

### DIFF
--- a/src/plugin/withAndroidMath.ts
+++ b/src/plugin/withAndroidMath.ts
@@ -1,6 +1,4 @@
-import configPlugins, { type ConfigPlugin } from '@expo/config-plugins';
-
-const { withGradleProperties } = configPlugins;
+import { withGradleProperties, type ConfigPlugin } from '@expo/config-plugins';
 
 export const withAndroidMath: ConfigPlugin<{ enableMath?: boolean }> = (
   config,
--- a/src/plugin/withIosMath.ts
+++ b/src/plugin/withIosMath.ts
@@ -1,8 +1,6 @@
-import configPlugins, { type ConfigPlugin } from '@expo/config-plugins';
+import { withDangerousMod, type ConfigPlugin } from '@expo/config-plugins';
 import fs from 'fs';
 import path from 'path';
-
-const { withDangerousMod } = configPlugins;
 
 const IOS_MATH_OPTION = "ENV['ENRICHED_MARKDOWN_ENABLE_MATH'] = '0'";
 


hknakn:
Hi @hryhoriiK97 — this resolves the report from #192 (cc @JustJoostNL).

The minimal reproducer and fix are in the PR description: the plugin's `import configPlugins from '@expo/config-plugins'; const { withDangerousMod } = configPlugins;` shape resolves to `undefined` under a CJS loader (e.g. `tsx/cjs`, which Expo's `app.config.ts` users typically wire in), so it throws at module load before the `enableMath` check runs. Switching to named imports fixes it with no behaviour change.

I verified `yarn typecheck`, `yarn lint`, and `yarn prepare` locally, and ran `npx expo prebuild` against our app with the patched build — prebuild now completes and `enableMath: false` is honoured (`ENV['ENRICHED_MARKDOWN_ENABLE_MATH'] = '0'` lands in the generated `ios/Podfile`).

Note: I didn't exercise the example app or the Maestro E2E suite — the change is build-time plugin module-load only, covered by `yarn prepare` plus the real-world prebuild repro. Happy to add a small Jest smoke test for the plugin if you'd like one (there are no plugin tests today). Thanks for looking!

eszlamczyk (approved):
Thanks! Verified the compiled output and confirmed no default export - the fix looks great! Should be merged next week

hknakn:
@eszlamczyk  thanks for merging! can we release a new version?

eszlamczyk:
@hknakn we aren't planning any releases for now, but you can use the _nightly release_, the fix should be already released [there](https://www.npmjs.com/package/react-native-enriched-markdown/v/0.7.0-nightly-20260601-94731448f)

PR #357, opened 2026-05-27
feat(ios): integrate RaTeX for LaTeX math rendering, replacing iosMath

hryhoriiK97:
### What/Why?

Replace **iosMath** with **RaTeX** for iOS LaTeX math rendering, aligning iOS with the Android migration (PR #355).

**iosMath** only supports a limited subset of LaTeX and is no longer actively maintained. **RaTeX** is a modern, Rust-based, KaTeX-compatible engine that renders math natively via CoreGraphics — no WebView, no JavaScript. This gives iOS feature parity with web (KaTeX) and Android (RaTeX).

### Changes

**New file:**
- `ios/math/ENRMRaTeXBridge.swift` — Swift-to-ObjC bridge exposing `ENRMRaTeXBridge.parse()` and `ENRMRaTeXRenderResult` (width, ascent, descent, draw) to existing Objective-C code.

**Podspec (`ReactNativeEnrichedMarkdown.podspec`):**
- Removed `iosMath` dependency.
- Added conditional `spm_dependency` for RaTeX v0.1.9+ (requires `cocoapods-spm` plugin).
- Swift source files conditionally included only when math is enabled.
- C/C++ headers marked as private to prevent module map conflicts with `use_frameworks!`.

**Migrated ObjC files:**
- `ENRMMathContainerView.m` — Replaced `MTMathUILabel` with a custom `ENRMRaTeXCanvasView` that draws via `ENRMRaTeXRenderResult`. Added alignment support (center/left/right) and correct overflow scrolling.
- `ENRMMathInlineAttachment.m` — Replaced `MTMathUILabel` with `ENRMRaTeXBridge.parse()`. Uses `ceil()` for dimension rounding to prevent sub-pixel clipping.
- `ENRMMathInlineAttachment+macOS.m` — Same migration for macOS. Added CTM flip since RaTeX expects top-left origin (UIKit-style) but `NSImage.lockFocus` uses bottom-left origin.
- `ENRMMathInlineAttachmentShared.h` — Replaced `MTMathListDisplay` ivar with `ENRMRaTeXRenderResult`.
- `ENRMFeatureFlags.h` — Auto-detection now checks for RaTeX headers instead of iosMath.

**Example app & docs:**
- `Podfile` — Sets `USE_FRAMEWORKS=dynamic` when math is enabled (required for SPM interop).
- `LATEX_MATH.md` — Updated references from iosMath to RaTeX, documented `use_frameworks!` requirement.

### Consumer impact

[...]

PR #361, opened 2026-05-31
latex: upgrade iosMath 2.2.0

Gregoirevda:
### What/Why?
Upgrade iosMath which has recently fixed many math rendering issues 
https://github.com/kostub/iosMath/releases

The recent package is published on SPM only, switching to SPM works but requires adding the header search paths (iosMath resolves 3 sibling directories). 

Note: 
The flag set [here](https://github.com/software-mansion-labs/react-native-enriched-markdown/blob/af2d2f863fde6b7d6c21064433fb3274ae2c7496/ios/utils/ENRMFeatureFlags.h#L4) will only rely on the flag set in cocoapods spec, the check on `<iosMath.h` won't work

### DIFF
--- a/ReactNativeEnrichedMarkdown.podspec
+++ b/ReactNativeEnrichedMarkdown.podspec
@@ -22,11 +22,15 @@ Pod::Spec.new do |s|
   preprocessor_defs = '$(inherited) MD4C_USE_UTF8=1'
   if enable_math
     preprocessor_defs += ' ENRICHED_MARKDOWN_MATH=1'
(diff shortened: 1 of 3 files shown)


eszlamczyk:
Hi @Gregoirevda thanks for your contribution, we appreciate that!

Unfortunately we are migrating of `iosMath` (and `androidMath`) for [RaTeX](https://github.com/erweixin/RaTeX). This change should be merged this week in #355 and #357, so this change will not be merged

Gregoirevda:
Thanks for the reply, closing the PR

PR #363, opened 2026-05-31
fix(ci): fix iOS build by removing stale codegen reset

hryhoriiK97:
### What/Why?

The iOS CI build was failing with:
```
[!] No podspec found for `ReactAppDependencyProvider` in `build/generated/ios/ReactAppDependencyProvider`
```

The cocoapods cache could restore `Pods/` without the corresponding `build/generated/ios/` artifacts, causing `pod install` to be skipped while codegen files were missing. When `react-native build-ios` then ran its own `pod install`, it failed because `ReactAppDependencyProvider` didn't exist.

### Changes

- Reset `build/`, `Pods/`, and `Podfile.lock` together
- Always run `pod install` when turbo cache misses — deleting `Podfile.lock` ensures the cocoapods cache primary key never matches, so `pod install` always regenerates app codegen artifacts (`ReactAppDependencyProvider`)
- Keep cocoapods cache restore via `restore-keys` prefix match as a warm start for faster installs
- Keep `yarn prepare` (needed for library codegen — `EventEmitters.h`, `Props.h`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
-      - name: Build library and generate codegen
+      - name: Build library
         run: yarn prepare
 
       - name: Set up Xcode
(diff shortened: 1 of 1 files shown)


eszlamczyk (approved):
Looks correct but I believe this sollution was already present and has issues with hitting the cache (but lets keep it for now)

PR #370, opened 2026-06-02
fix(android): task list press not working

eszlamczyk:
### What/Why?
Fix task list checkbox toggle not re-rendering in flavor="github" on Android.

In the _github_ flavor (`EnrichedMarkdown`), `setMarkdownContent` only sets a `renderPending` flag - the actual render is deferred until `commitProps()` is called, which only happens during a React prop update cycle (`onAfterUpdateTransaction`). When a checkbox was tapped, the callback called `setMarkdownContent(updatedMarkdown)` but `commitProps()` was never called, so the render never fired.

Additionally, event callbacks (link press, link long press, task list item press) were incorrectly set inside the `@ReactProp setMarkdown` setter in both `EnrichedMarkdownManager` and `EnrichedMarkdownTextManager`, causing them to be re-registered on every React prop update. Moved all callbacks to `createViewInstance` where one-time view setup belongs.

### Testing
Open playground
Create any task list for example

```
- [ ] test
- [x] test2
```

### Android

https://github.com/user-attachments/assets/fdf2f134-e478-4aa6-af60-a09082d920eb

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
+++ b/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownManager.kt
@@ -1,7 +1,6 @@
 package com.swmansion.enriched.markdown
 
 import android.content.Context
-import android.util.Log
 import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.bridge.ReadableMap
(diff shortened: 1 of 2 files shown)


hryhoriiK97 (approved):
LGTM!

PR #374, opened 2026-06-03
feat: add line break and text break props

### LINKED ISSUE #366: Bug(android): Long URLs cause incorrect EnrichedMarkdownText height
## Bug description ##
A long URL in Markdown text that renders over more than one line causes an `EnrichedMarkdownText` element to set its height incorrectly, causing unnecessary scrolling. This appears to be an Android-specific problem, as it renders with the correct height on iOS.

## To reproduce ##
Render an `EnrichedMarkdownText` element on a React Native page with a URL that wraps more than one line. The exact reproduction will depend on font choice, screen size, etc., but this is a minimal reproduction in my context (screenshots below).

```
<Text>Some text before rendered Markdown</Text>
<EnrichedMarkdownText
  markdown={'Find related online resources here: https://www.example.com/a-long-bit-of-text-that-keeps-going'}
/>
<Text>Some text after rendered Markdown</Text>
```

## Expected behaviour ##
The `EnrichedMarkdownText` element should render with the correct calculated height, so it does not cut off text or cause a scroll.

### Correct behaviour (iOS) ###
Correctly rendered URL on iOS

### Incorrect behaviour (Android) ###
Incorrectly rendered URL on Android, showing URL string is cut off<br/>
Note that text is cut off, due to rendering with only two lines of height, rather than three.
<br/>
Incorrectly rendered URL on Android, showing scrolling behaviour<br/>
The text can be dragged up to scroll to show all three lines.

## Devices affected ##
Occurs on all Android devices tested; details listed below for information.
 - Android Emulator, API 36.0
 - Google Pixel 10 Pro XL, Android v16
 - Motorola Moto G71 5G, Android v11
 - Google Pixel 8, Android v14
 - Samsung Galaxy Tab S10 Plus, Android v15

## Additional context ##
This behaviour only seems to occur when there is some text immediately before the long URL, and there is enough space between that text and the line end that some of the URL _could_ render there (although it does not: it always starts on a new line on Android).

eszlamczyk:
### What/Why?
Adds `textBreakStrategy` (Android) and `lineBreakStrategyIOS` (iOS) props to both `EnrichedMarkdownText`, mirroring the same props available on React Native's core `Text` component.

On Android, the break strategy must match between the measurement pass (`StaticLayout.Builder`) and the render pass (`TextView.breakStrategy`) - a mismatch causes the measured line count to differ from the rendered one, which manifests as incorrect view height and clipped or scrolled content, most visibly with long links that wrap across lines. Previously the strategy was hardcoded to `simple` via a top-level function; this PR promotes it to a singleton (`BreakStrategyUtils`) that both paths read from, so changing the prop keeps them in sync.

On iOS, the `lineBreakStrategy` is applied to the `NSParagraphStyle` used when building attributed strings, allowing consumers to opt into `standard`, `hangul-word`, or `push-out`
breaking on iOS 14+.

*Note: the defaults were kept in a way not to have any breaking changes*

### Testing
Go to playground in example app

paste (via set markdown option) string containing long text/links for example:
```
Check out this link: [A very long link label that should wrap onto the next line in a narrow container (https://example.com)

And a bare URL: https://this-is-a-very-long-url-that-will-force-line-wrapping.example.com/some/deep/path
```

To make this more visible adjust max width of the container (to for example 220)

#### Alternative Test
Create any view that has:
- some `<Text>`
- Markdown Text block with text before and link that is longer than one line
- some `<Text>`

<details>
<summary> Example </summary>

Replace `App.tsx` in example app with

```ts
import { Text, View } from 'react-native';
import { EnrichedMarkdownText } from 'react-native-enriched-markdown';

export default function App() {
  return (
    <View style={{ padding: 16, marginTop: 50 }}>
      <Text>Some text before rendered Markdown</Text>
      <EnrichedMarkdownText
        markdown={
          'Find related: https://www.example.com/a-long-bit-of-text-that-keeps-going'
        }
      />
      <Text>Some text after rendered Markdown</Text>
    </View>
  );
}
```
</details>

##### Android

| `simple` | `highQuality` (default) | `balanced` |
| -- | -- | -- |
| image | image | image |

##### iOS

| `none` (default) | `standard` | `hangul-word` | `push-out` |
| -- | -- | -- | -- |
| image | image | image |  image |

#### Alternative test

*Note: Before photo is partialy scrolled to display the issue* 

| Before | After |
| - | - |
| image | image |

### PR Checklist

- [x] E2E tests are passing

closes #366


hryhoriiK97 (commented):
Looks like this approach fixes the current problem, but changing from HIGH_QUALITY to SIMPLE could be a breaking change for existing library consumers. I'm thinking about introducing a lineBreakStrategy prop — the same as RN Text has: https://reactnative.dev/docs/text#textbreakstrategy-android

android/src/main/java/com/swmansion/enriched/markdown/utils/common/BreakStrategyUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+ * is enough for both paths.
+ *
+ * Note: call sites in StaticLayout.Builder suppress "WrongConstant" lint. This is
+ * intentional - Layout.BREAK_STRATEGY_SIMPLE and LineBreaker.BREAK_STRATEGY_SIMPLE
+ * are the same integer (0), but the @IntDef annotation on StaticLayout.Builder
+ * .setBreakStrategy() was changed from Layout.* to LineBreaker.* in API 29.
+ * The suppression is safe; the Layout.* constants share the same integer values.
+ */
+object BreakStrategyUtils {
+  private var strategy: String = "simple"
```
hryhoriiK97: The default/fallback here is still `"simple"`, but managers, codegen, and the JS wrapper all default to `"highQuality"`.  Let's use 'highQuality' as default.

android/src/main/java/com/swmansion/enriched/markdown/utils/common/BreakStrategyUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+
+  fun setStrategy(newStrategy: String?) {
+    strategy = newStrategy ?: "simple"
+  }
```
hryhoriiK97: `BreakStrategyUtils` is a process-wide singleton, but `textBreakStrategy` is a per-view prop - with multiple markdown views on screen, the last setter wins for all measurement/render paths.

RN `Text` doesn't do this: each shadow node stores its own strategy (`mTextBreakStrategy` / `ParagraphAttributes.textBreakStrategy`) and passes the same value to both `StaticLayout.Builder` (measure) and `ReactTextView.setBreakStrategy()` (render).

Could we follow the existing `MeasurementStore.fontScalingSettings` pattern instead - store per `viewId`, resolve during measurement, set on each view's `TextView`?

ios/utils/ParagraphStyleUtils.m
```diff
@@ -3,6 +3,20 @@
 
 NSAttributedString *kNewlineAttributedString;
 static NSParagraphStyle *kBlockSpacerTemplate;
+static NSLineBreakStrategy gLineBreakStrategy = NSLineBreakStrategyNone;
+
+void ENRMSetLineBreakStrategy(NSString *strategy)
+{
```
hryhoriiK97, line 8: This file already has statics like `kBlockSpacerTemplate`, but those are immutable shared templates (copied per use, same for all views).

`gLineBreakStrategy` is different — it's mutable state for a per-view prop. If two markdown views use different `lineBreakStrategyIOS` values, the last setter wins in `getOrCreateParagraphStyle`.

Font scaling is already stored per view here; RN `Text` does the same for `lineBreakStrategy`. Could we store this on the view/renderer and pass it through instead of a global?

src/types/MarkdownTextProps.ts
```diff
@@ -213,4 +213,26 @@ export interface EnrichedMarkdownTextProps extends Omit<ViewProps, 'style'> {
...
+  /**
+   * Sets the line break strategy on iOS (iOS 14+).
+   * - `'none'` (default): no additional line break strategy.
+   * - `'standard'`: standard line breaking rules.
+   * - `'hangul-word'`: Korean word-boundary breaking.
+   * - `'push-out'`: pushes text out to avoid orphaned words.
+   * @default 'none'
+   * @platform ios
+   */
+  lineBreakStrategyIOS?: 'none' | 'standard' | 'hangul-word' | 'push-out';
```
hryhoriiK97, line 237: Could we update API_RENCE doc with these props? 🙏

hryhoriiK97 (changes requested):

ios/utils/ParagraphStyleUtils.m
```diff
@@ -3,6 +3,20 @@
...
+  if ([strategy isEqualToString:@"standard"]) {
```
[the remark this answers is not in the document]
eszlamczyk: duplicate

android/src/main/java/com/swmansion/enriched/markdown/utils/common/BreakStrategyUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Note: call sites in StaticLayout.Builder suppress "WrongConstant" lint. This is
+ * intentional - Layout.BREAK_STRATEGY_SIMPLE and LineBreaker.BREAK_STRATEGY_SIMPLE
+ * are the same integer (0), but the @IntDef annotation on StaticLayout.Builder
+ * .setBreakStrategy() was changed from Layout.* to LineBreaker.* in API 29.
+ * The suppression is safe; the Layout.* constants share the same integer values.
+ */
+object BreakStrategyUtils {
+  private var strategy: String = "simple"
+
```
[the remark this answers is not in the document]
eszlamczyk: duplicate

ios/utils/ParagraphStyleUtils.m
```diff
@@ -183,6 +195,27 @@ void applyTextAlignment(NSMutableAttributedString *output, NSRange range, NSText
   [output addAttribute:NSParagraphStyleAttributeName value:style range:range];
 }
 
+void ENRMApplyLineBreakStrategyToParagraphStyles(NSMutableAttributedString *output,
+                                                 NSLineBreakStrategy lineBreakStrategy)
+{
+  if (output.length == 0) {
+    return;
+  }
+  if (@available(iOS 14.0, *)) {
```
hryhoriiK97: @available(iOS 14.0, *) is redundant given RN’s min iOS 15.1.

hryhoriiK97 (approved):
LGTM 🚀 I've left one additional comment

PR #376, opened 2026-06-03
feat: forward nativeID and testID to the native input

PierreCapo:
### What/Why?

`EnrichedMarkdownTextInput`'s wrapper forwards only an explicit allow-list of props to `EnrichedMarkdownTextInputNativeComponent` and drops everything else, so `nativeID` and `testID` never reach the underlying native view — even though `NativeProps extends ViewProps` already supports them.

This forwards both props through.

Concrete motivation for `nativeID`: `react-native-keyboard-controller`'s `KeyboardGestureArea` / `KeyboardChatScrollView` link a composer input via a `textInputNativeID`. With a chat input built on `EnrichedMarkdownTextInput`, that linkage is impossible today because the `nativeID` is dropped before it reaches the native view. `testID` is forwarded alongside it as the other standard identifier prop, useful for E2E/unit targeting.

### Testing

- Set `nativeID` / `testID` on `<EnrichedMarkdownTextInput>` — both now reach `EnrichedMarkdownTextInputNativeComponent`, whose `NativeProps` already extends `ViewProps` (so no native changes are needed).
- JS-only change. I wasn't able to run the example app / E2E suite locally, so I'd appreciate CI confirming lint/typecheck. Happy to adjust naming, placement, or add a test if you'd prefer.

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/docs/API_REFERENCE.md
+++ b/docs/API_REFERENCE.md
@@ -442,6 +442,22 @@ Style for the input view. Accepts `ViewStyle` and `TextStyle` properties (e.g., 
 | ----------------------- | ------------- | -------- |
 | `ViewStyle \| TextStyle` | -             | Both     |
 
+### `nativeID`
+
+Identifier forwarded to the underlying native view. Useful for integrations that need to reference the input natively, e.g. `react-native-keyboard-controller`'s `textInputNativeID`.
+
+| Type     | Default Value | Platform |
+| -------- | ------------- | -------- |
+| `string` | -             | Both     |
+
+### `testID`
+
+Identifier forwarded to the underlying native view for use in end-to-end and unit tests.
+
+| Type     | Default Value | Platform |
+| -------- | ------------- | -------- |
+| `string` | -             | Both     |
+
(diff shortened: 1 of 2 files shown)


hryhoriiK97 (changes requested):
Hi @PierreCapo, thank you for the PR! Instead of adding these props explicitly, I think it's better to extend EnrichedMarkdownTextInputProps from ViewProps — something like this:

```
export interface EnrichedMarkdownTextInputProps
  extends Omit<ViewProps, 'style'>
```

and then pass the rest props to the component. Also, no need to include these props in the API_REFERENCE — they are built into the React Native View component.

PierreCapo:
Hello @hryhoriiK97 thanks for your reply. I think it's indeed better, I wasn't sure if you had somehow a reason to not spread all props to the component.

hryhoriiK97 (approved):
@PierreCapo thank you for the contribution! This change will be included in tomorrow's nightly release 🙂

PR #404, opened 2026-06-17
chore: move cpp folder into packages/core

graszka22:
### What/Why?

Moves packages/react-native-enriched-markdown/cpp into packages/core. packages/core will be the source of truth for cpp core and used by other packages. 

There is a hack in CmakeLists.txt and ReactNativeEnrichedMarkdown.podspec‎:
- in dev we use files from package/core, on ios the package/core is a pod (no need to publish it anywhere) and on android we use a relative path
- in release we use files copied from package/core to packages/react-native-enriched-markdown/cpp. The prepack script copies them automatically before packing the npm package. 

I don't like it but I see no other easy way, because cocoapods doesn't follow symlinks and doesn't allow sources from outside packages/react-native-enriched-markdown. And cmake has issues with following symlinks too. I think the only other solution would be publishing the core separately but we'd rather avoid this? 

After react-native-enriched-markdown will depend on the native libraries the release flow will change anyway and probably will be nicer. 

### Testing
- Built the example apps. 
- Created a tarball and added a dependency on the tarball in the example app. The tarball contained cpp code. Built the app and it worked.

### PR Checklist

- [x] E2E tests are passing


packages/core/EnrichedMarkdownCore.podspec
```diff
@@ -0,0 +1,24 @@
...
+Pod::Spec.new do |s|
+  s.name         = "EnrichedMarkdownCore"
+  s.version      = package["version"]
+  s.summary      = package["description"]
+  s.homepage     = "https://github.com/software-mansion/react-native-enriched-markdown"
+  s.license      = { :type => "MIT" }
+  s.authors      = "Software Mansion"
+  s.source       = { :git => "https://github.com/software-mansion/react-native-enriched-markdown.git" }
+
+  s.platforms    = { :ios => "15.1", :osx => "14.0" }
```
hryhoriiK97: This hardcodes the iOS minimum to 15.1, while ReactNativeEnrichedMarkdown.podspec uses the dynamic min_ios_version_supported from React Native. If the RN minimum changes, these will drift apart and could cause CocoaPods resolution issues. Let's use the same helper here

packages/react-native-enriched-markdown/ReactNativeEnrichedMarkdown.podspec
```diff
@@ -13,13 +16,20 @@ Pod::Spec.new do |s|
...
-  s.private_header_files = "ios/**/*.h", "cpp/**/*.{h,hpp}"
+  s.private_header_files = "ios/**/*.h"
+  s.source_files = "ios/**/*.{h,m,mm,cpp,swift}"
+
+  unless monorepo
+    s.private_header_files = "ios/**/*.h", "cpp/**/*.{h,hpp}"
+    s.source_files = "ios/**/*.{h,m,mm,cpp,swift}", "cpp/md4c/*.{c,h}", "cpp/parser/*.{hpp,cpp}"
+  else
+    s.dependency "EnrichedMarkdownCore"
+  end
```
hryhoriiK97, line 26: The first two assignments to s.private_header_files and s.source_files are immediately overridden in the unless branch - they only matter for the else (monorepo) path. This reads a bit confusingly. Would be clearer as a straight if/else. What do you think about this?

```
if monorepo
  s.private_header_files = "ios/**/*.h"
  s.source_files = "ios/**/*.{h,m,mm,cpp,swift}"
  s.dependency "EnrichedMarkdownCore"
else
  s.private_header_files = "ios/**/*.h", "cpp/**/*.{h,hpp}"
  s.source_files = "ios/**/*.{h,m,mm,cpp,swift}", "cpp/md4c/*.{c,h}", "cpp/parser/*.{hpp,cpp}"
end
```

packages/react-native-enriched-markdown/package.json
```diff
@@ -27,7 +27,7 @@
     "app.plugin.js"
   ],
   "scripts": {
-    "build:wasm": "bash cpp/wasm/build.sh",
+    "build:wasm": "bash ../../packages/core/cpp/wasm/build.sh",
```
hryhoriiK97: Nit: the path goes up to the repo root and back down. Since we're already in packages/react-native-enriched-markdown, this is equivalent and shorter:

```
"build:wasm": "bash ../core/cpp/wasm/build.sh",
```

hryhoriiK97 (approved):
LGTM! I've tested this locally - I generated a `.tgz` file and installed it in a newly created RN project - everything works as before. Also, I've left a few comments — please take a look at them 🙏

hryhoriiK97 (approved):

Issue #407, opened 2026-06-18
Allow hiding the built-in "Format" and "Copy as Markdown" options in the text input's selection menu


sgrvl:
**Is your feature request related to a problem? Please describe.**
`EnrichedMarkdownTextInput`'s text-selection menu always shows a fixed **Format** submenu and a **Copy as Markdown** action, with no way to hide any of them. We only use a subset of these styles and don't need to let users **Copy as Markdown**.

Localization makes it worse: we ship in many languages, but these items are English-only with no way to override the labels. Hiding the ones we don't use helps on both fronts, and for the ones we keep, being able to supply our own localized labels would let them match the rest of our UI.

The read-only components (`EnrichedMarkdownText` / `EnrichedMarkdown`) already got a `selectionMenuConfig` prop to hide built-in selection-menu actions (#272 → #285), but it was never wired to the editable input.

**Describe the solution you'd like**
Add a `selectionMenuConfig` prop to `EnrichedMarkdownTextInput`, mirroring the one already on the read-only components (#285). It should be default-on and backward compatible: omitting it (or any field) reproduces today's exact menu.

At minimum, a coarse on/off for the two built-ins, hiding the whole **Format** submenu and/or the **Copy as Markdown** action:

```tsx
<EnrichedMarkdownTextInput
  selectionMenuConfig={{ format: false, copyAsMarkdown: false }}
/>
```

Two refinements would be valuable on top of that:

- **Per-item format control**: rather than all-or-nothing on the Format submenu, let us choose which individual items appear, so we can drop just Spoiler while keeping Bold / Italic / Underline / Strikethrough / Link. Widening `format` from a plain boolean to `boolean | Partial<Record<FormatStyle, boolean>>` covers both cases, with `format: false` still meaning "hide everything".

- **Custom labels**: let us supply our own string for each built-in item, so the menu can be localized to match the rest of our app instead of showing the English labels only.

hryhoriiK97:
Hi @sgrvl, I've created PRs (#408 #409) with selectionMenuConfig and formatMenuConfig. They will be merged tomorrow. 

Regarding labels - this will be handled in a separate PR. We should think about how we want to solve localization.

hryhoriiK97:
@sgrvl both PRs have been merged and will be available in tomorrow's nightly release. I don't close this issue as we'll return to the localizations/labels next week.

hryhoriiK97:
[...]

PR #415, opened 2026-06-22
feat: localize selection menu labels via selectionMenuLabels prop

### LINKED ISSUE #198: Context menu items are not localized
## Description

The custom context menu items ("Copy", "Copy as Markdown", "Copy Image URL") are hardcoded in English in the native code. On devices with a non-English locale, the system menu items (Look Up, Translate, Search Web, etc.) are properly localized by the OS, but the library's custom items remain in English — creating an inconsistent experience.

## Expected behavior

Custom menu items should respect the device locale, or at minimum provide a way to pass localized strings from the JS side.

## Current behavior

On a German device for example:
- System items: "Nachschlagen", "Übersetzen", "Im Web suchen" ✅
- Library items: "Copy", "Copy as Markdown" ❌

## Affected files

- iOS: `ios/utils/EditMenuUtils.m` — uses `@"Copy"` instead of `NSLocalizedString`
- Android: `android/.../SelectionActionMode.kt` — uses hardcoded `"Copy"` instead of `R.string`
- Same pattern in `TableContainerView` and `MathContainerView` on both platforms

## Suggestion

Either:
1. Use `NSLocalizedString` / Android string resources with bundled translations
2. Expose a prop to pass localized labels from the JS side

marcellov7:
## Summary

Closes #198.
This addresses #407.

The built-in copy actions (**Copy**, **Copy as Markdown**, **Copy Image URL**) are hardcoded in English in the native code. On non-English apps they stay in English while the OS-provided menu items (Look Up, Translate…) are localized, creating an inconsistent menu — exactly the problem reported in #198.

This adds a **`selectionMenuLabels`** prop on `EnrichedMarkdownText` so the labels can be overridden from JS (option 2 in the issue), typically wired to the app's i18n:

```tsx
<EnrichedMarkdownText
  markdown={content}
  selectionMenuLabels={{
    copy: t('copy'),
    copyAsMarkdown: t('copyAsMarkdown'),
    copyImageUrl: t('copyImageUrl'),
    copyImageUrls: t('copyImageUrls'), // "{count}" is replaced by the image count
  }}
/>
```

Any label left `undefined` keeps its English default, so the change is fully backward compatible.

## How it works

The labels are folded into the existing `selectionMenuConfig` codegen struct (normalized JS-side, where an empty string means "use the native default"), so no new native prop plumbing is introduced.

- **TS** — new `SelectionMenuLabels` type + prop; normalization in `native/EnrichedMarkdownText.tsx`; struct fields added to both text/github codegen specs.
- **Android** — labels read in `MarkdownViewManagerUtils`, applied in `SelectionActionMode` and threaded to the table and math copy menus.
- **iOS / macOS** — labels carried on `ENRMSelectionMenuConfig` (strong owners on the view) and applied in `EditMenuUtils`(+macOS) and the `TableContainerView` / `ENRMMathContainerView` menus.

## Scope

Covers `EnrichedMarkdownText` (commonmark + github flavor), including the table and math block copy menus listed in #198. `EnrichedMarkdownTextInput` keeps only its visibility config for now (the Format submenu labels could be a follow-up).

## Notes for reviewers

- Docs updated: `docs/COPY_OPTIONS.md` (new "Localizing Menu Labels" section) and `docs/API_REFERENCE.md`.
- `yarn typecheck` and `eslint` pass locally. I couldn't run the native (Xcode/Gradle) builds or clang-format/ktlint in my environment, so please double-check the native formatting matches the hooks — happy to adjust.
- No public API removed; existing behavior is unchanged when the prop is omitted.


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -477,6 +479,10 @@ class EnrichedMarkdown
           resolvedClass
             .getConstructor(Context::class.java, StyleConfig::class.java)
             .newInstance(context, style) as View
+        resolvedClass.getMethod("setCopyLabel", String::class.java)
+          .invoke(view, selectionMenuConfig.copyLabel)
+        resolvedClass.getMethod("setCopyAsMarkdownLabel", String::class.java)
+          .invoke(view, selectionMenuConfig.copyAsMarkdownLabel)
```
eszlamczyk, line 512: Looks like this code is missing a assign in the *prop change* section. We need to update everything in case of re-update

For example if someone updates a prop (customer changes language on page without a remount), they would get wrong language
Look at `EnrichedMarkdown.kt 229-235` for an example

packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
eszlamczyk, line 1: Looks like this has the same issues as the kotlin. The props are not being walked down to children on updateProps:

follow the existing iOS prop-update idiom - see `pushWritingDirectionToTableSegments` at line 386 for the model. Add an equivalent pushSelectionMenuLabelsToSegments covering tables and math, call it from updateProps: after the label ivars are reassigned.

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
```diff
@@ -7,6 +7,11 @@
 
 #if TARGET_OS_OSX
 
+static NSString *resolveMenuLabel(NSString *_Nullable label, NSString *fallback)
+{
+  return label.length > 0 ? label : fallback;
+}
+
```
eszlamczyk: Nitpick: On macOS, the table and math context menus fall back to `NSLocalizedString(@"Copy", nil)` / `NSLocalizedString(@"Copy as Markdown", nil)` (see `ENRMMathContainerView.m:143-145` and the `TableContainerView.m` macOS branch). This helper returns the literal English string instead, so an app shipping its own Localizable.strings gets two different fallback behaviors between the main edit menu and the table/math menus. Suggest wrapping the fallback in `NSLocalizedString` here so all three macOS menus behave consistently

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
```diff
@@ -39,13 +45,21 @@
...
-                          : [NSString stringWithFormat:@"Copy %lu Image URLs", (unsigned long)imageURLs.count];
+    NSString *title;
+    if (imageURLs.count == 1) {
+      title = resolveMenuLabel(selectionMenuConfig.copyImageUrlLabel, @"Copy Image URL");
+    } else if (selectionMenuConfig.copyImageUrlsLabel.length > 0) {
+      NSString *countString = [@(imageURLs.count) stringValue];
+      title = [selectionMenuConfig.copyImageUrlsLabel stringByReplacingOccurrencesOfString:@"{count}"
+                                                                                 withString:countString];
+    } else {
+      title = [NSString stringWithFormat:@"Copy %lu Image URLs", (unsigned long)imageURLs.count];
```
eszlamczyk: Same issue as the comment above in this file

packages/react-native-enriched-markdown/src/native/EnrichedMarkdownText.tsx
```diff
@@ -146,8 +149,13 @@ export const EnrichedMarkdownText = ({
     () => ({
       copyAsMarkdown: selectionMenuConfig?.copyAsMarkdown ?? true,
       copyImageUrl: selectionMenuConfig?.copyImageUrl ?? true,
+      // Empty string is the sentinel for "use the native English default".
+      copyLabel: selectionMenuLabels?.copy ?? '',
+      copyAsMarkdownLabel: selectionMenuLabels?.copyAsMarkdown ?? '',
+      copyImageUrlLabel: selectionMenuLabels?.copyImageUrl ?? '',
+      copyImageUrlsLabel: selectionMenuLabels?.copyImageUrls ?? '',
```
eszlamczyk: ~Very Nitpicky nit: I can see a world where a libray consumer using i18next with returnEmptyString: true could accidentally pass '' thinking they're suppressing the menu item, and silently get the English default instead. Instead I'd use `null` as a dedicated marker (or very clearly document it in JSDoc)~

Edit: Now we think the best idea would be to do `label ?? DEFAULT_VALUE` for each label, and resolve this at the level of `typescript`. This way all the native code could drop those *if empty* clauses

packages/react-native-enriched-markdown/ios/views/TableContainerView.h
```diff
@@ -30,6 +30,10 @@ typedef void (^TableLinkPressBlock)(NSString *url);
 @property (nonatomic, assign) ENRMWritingDirectionMode writingDirectionMode;
 @property (nonatomic, assign) NSWritingDirection resolvedLayoutDirection;
 
+// Localized labels for the copy menu. Empty/nil means "use the English default".
```
eszlamczyk: This comment is unnecessary:
1. It restates the property name
2. The second half is correct, but its duplicated in 4 places already.

I'd drop this (and other like that) comment and just keep it in `docs/` and JSDoc on the JS types

packages/react-native-enriched-markdown/src/EnrichedMarkdownNativeComponent.ts
```diff
@@ -230,6 +230,13 @@ export interface ContextMenuItemConfig {
 export interface SelectionMenuConfig {
   copyAsMarkdown: boolean;
   copyImageUrl: boolean;
+  // Localizable labels for the built-in selection menu actions. An empty
+  // string means "use the built-in English default". `copyImageUrlsLabel`
+  // is a template where the `{count}` token is replaced by the image count.
```
eszlamczyk: Drop this comment as well (its already restated in other places)

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextNativeComponent.ts
```diff
@@ -230,6 +230,13 @@ export interface ContextMenuItemConfig {
 export interface SelectionMenuConfig {
   copyAsMarkdown: boolean;
   copyImageUrl: boolean;
+  // Localizable labels for the built-in selection menu actions. An empty
+  // string means "use the built-in English default". `copyImageUrlsLabel`
+  // is a template where the `{count}` token is replaced by the image count.
```
eszlamczyk: Same treatment for this comment

eszlamczyk (changes requested):
Hi @marcellov7 Thanks for the pull request! The cross-platform coverage is mostly solid, although I have one structural request, one open question about plurals, and a handful of smaller things to address before we merge.

Main issue is:
Consolidate `selectionMenuLabels` into `selectionMenuConfig`

for example:
```ts
export interface SelectionMenuConfig {
  /** System Copy item can't be hidden — only its label is configurable. */
  copy?: { label?: string };
  copyAsMarkdown?: { enabled?: boolean; label?: string };
  copyImageUrl?: {
    enabled?: boolean;
    label?: string;
    pluralLabels?: TBD;
  };
}
```

This is a breaking change for selectionMenuConfig ({ copyAsMarkdown: false } → { copyAsMarkdown: { enabled: false } }), but the 0.6 → 0.7 bump pays for it. Drop the standalone selectionMenuLabels prop entirely - it hasn't been released yet.

Although we want to make sure to mark this change as deprecated in non-ts apps, my proposition is to soften the break with a one-release runtime fallback. TS consumers will get a compile error on upgrade and migrate cleanly, but JS-only consumers and anyone whose config flows through any would silently keep using the old shape with no signal anything changed. Accept the boolean form at runtime for 0.7 with a one-time warnOnce (but not `__DEV__`-gated — deprecation warnings need to surface in staging/TestFlight/CI prod builds too, otherwise developers miss them until end-users hit them), and mark it as deprecated for removal in 0.8.

the code might look something like:

```ts
const warned = new Set<string>();
const warnOnce = (key: string, msg: string) => {
  if (warned.has(key)) return;
  warned.add(key);
  console.warn(msg);
};
const normalizeItem = (raw: unknown, field: string, defaultEnabled: boolean) => {
  if (raw === undefined) return { enabled: defaultEnabled, label: '' };
  if (typeof raw === 'boolean') {
    warnOnce(
      `selectionMenuConfig.${field}`,
      `[react-native-enriched-markdown] selectionMenuConfig.${field} as boolean is deprecated; ` +
      `use { enabled: ${raw} }. Boolean form will be removed in 0.8.`
    );
    return { enabled: raw, label: '' };
  }
  const obj = raw as { enabled?: boolean; label?: string };
  return { enabled: obj.enabled ?? defaultEnabled, label: obj.label ?? '' };
};
```

Keep the public type pure (only the new shape) and confine the boolean acceptance behind a single as unknown cast at the wrapper boundary - the type stays clean while the runtime catches stragglers.

--- 

About the plural labels:
In some non-germanic languages there are multiple plurar form, so using "just a string" for plural form is probably incorrect. I see two viable options:

1. Create a callback function `f(count) => string` that  user can implement the value itself. This would be the best but I have concerns about performance of this option (there might be some race conditions that we render the incorrect label for the value). If you go this route, the race window is the gap between the native selection event and the JS-resolved label arriving back via a prop commit. The case that matters is fast long-press: user makes a selection that includes images and immediately invokes the menu - if the menu opens before the JS round-trip completes, native will fall back to the English template. Please verify the fallback is acceptable on both Android (ActionMode) and iOS (UIEditMenuInteraction), and that no platform shows the menu with an empty title in the worst case.

2. Use [pluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules). This for me looks like the most practical solution for this pr as it is not so complicated and shouldn't need to do a lot of changes. (Its good enough).

---

As a Nit, I'd like also this pull request to include a storybook entrance for the prop. Mirror the shape of the existing selectionMenuConfig storybook entry (or whichever is closest).

packages/react-native-enriched-markdown/src/native/EnrichedMarkdownText.tsx
marcellov7: Done in 66bfca4. Defaults are now resolved JS-side in `normalizeItem` (`label ?? DEFAULT_VALUE`), so the native code always receives a concrete string and there's no more empty-string sentinel. `selectionMenuLabels` was dropped and folded into `selectionMenuConfig`.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
marcellov7: Done in 5992cdd. `setSelectionMenuConfig` now re-pushes the copy labels to the cached table and math segment views (new `pushCopyLabelsToBlockSegments`), so a language change without a remount updates them too — mirroring the existing text-segment update at lines 229-235.

packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
marcellov7: Done in 3f1d70f. Added `pushSelectionMenuLabelsToSegments` (modelled on `pushWritingDirectionToTableSegments`) covering both `TableContainerView` and `ENRMMathContainerView`, called from `updateProps:` right after the label ivars are reassigned.

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
marcellov7: Done in 008f566. `resolveMenuLabel`'s fallback now uses `NSLocalizedString(fallback, nil)`, matching the table/math macOS menus. (The fallback is defensive now that labels are resolved JS-side.)

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
marcellov7: Done in 17c21fb. The `Copy %lu Image URLs` literal is gone: the image-URL title is now built by the shared `ENRMResolveImageURLsTitle` helper, which falls back to the (JS-resolved) `copyImageUrlsLabel` template — no English literal.

packages/react-native-enriched-markdown/ios/views/TableContainerView.h
marcellov7: Done in b6527da. Dropped the comment (and the equivalent ones on ENRMMathContainerView.h and the two Kotlin views); the behaviour lives in the JS types and docs now.

packages/react-native-enriched-markdown/src/EnrichedMarkdownNativeComponent.ts
marcellov7: Done in 17c21fb. Dropped this comment from the codegen spec.

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextNativeComponent.ts
marcellov7: Done in 17c21fb. Dropped this comment from the codegen spec.

marcellov7:
Thanks for the thorough review @eszlamczyk! All points addressed — pushed as separate commits on top:

**Structural**
- **Consolidated** `selectionMenuLabels` into `selectionMenuConfig` with the nested shape you proposed (`copy { label }`, `copyAsMarkdown { enabled, label }`, `copyImageUrl { enabled, label, pluralLabels }`); dropped the standalone prop. The legacy boolean form is accepted at runtime behind a single `as unknown` cast with a one-time `warnOnce` (not `__DEV__`-gated), deprecated for removal in 0.8. Public type stays object-only.
- **Defaults resolved JS-side** (`label ?? DEFAULT`), so native always receives a concrete string — no empty-string sentinel.
- **Plurals via `Intl.PluralRules`**: `copyImageUrl.pluralLabels` takes CLDR categories. JS precomputes a per-count template table (0..100) and packs it into one codegen string; native indexes it (wrapping counts > 100 with period 100, since the CLDR integer rules only depend on n, n%10, n%100) and substitutes `{count}`. Falls back to the singular/`{count}` templates when no plural labels are set or `Intl` is unavailable.
- **Storybook**: expanded the Selection Menu story with controls for the labels and the plural template.

**Native prop-update**
- Android: `setSelectionMenuConfig` now re-pushes labels to the table/math views (`pushCopyLabelsToBlockSegments`).
- iOS: added `pushSelectionMenuLabelsToSegments` (mirrors `pushWritingDirectionToTableSegments`), called from `updateProps:`.

**Nits**
- macOS `resolveMenuLabel` fallback now uses `NSLocalizedString`; the `Copy %lu Image URLs` literal is gone (shared `ENRMResolveImageURLsTitle` helper used by iOS + macOS).
- Dropped the redundant copy-label comments.

All review threads resolved. As before, I couldn't run the native (Xcode/Gradle) builds or clang-format/ktlint locally, so please double-check the native formatting.

apps/react-native-example/.rnstorybook/stories/components/EnrichedMarkdownText/props/SelectionMenu.stories.tsx
```diff
@@ -30,16 +53,40 @@ export default storyMeta('Props', 'Selection Menu');
...
     markdown: MARKDOWN,
-    copyAsMarkdown: true,
-    copyImageUrl: true,
+    copyAsMarkdownEnabled: true,
+    copyImageUrlEnabled: true,
+    // Defaults shown in Italian to demonstrate localization.
+    copyLabel: 'Copia',
+    copyAsMarkdownLabel: 'Copia come Markdown',
+    copyImageUrlLabel: 'Copia URL immagine',
+    copyImageUrlsLabel: 'Copia {count} URL immagini',
```
eszlamczyk: Praise: good idea to default it to other language!

docs/API_REFERENCE.md
```diff
@@ -350,16 +350,38 @@ Controls built-in actions added to the native text selection menu. Custom app-pr
 
 | Type                 | Default Value                                  | Platform |
 | -------------------- | ---------------------------------------------- | -------- |
-| `SelectionMenuConfig` | `{ copyAsMarkdown: true, copyImageUrl: true }` | iOS, Android, macOS |
+| `SelectionMenuConfig` | `{ copyAsMarkdown: { enabled: true }, copyImageUrl: { enabled: true } }` | iOS, Android, macOS |
```
eszlamczyk: Nit: looks like this default value does not match actual default value, lets replace it with 
```suggestion
| `SelectionMenuConfig` | `{} (see shape below for per-field defaults)` | iOS, Android, macOS |
```

docs/API_REFERENCE.md
```diff
@@ -350,16 +350,38 @@ Controls built-in actions added to the native text selection menu. Custom app-pr
...
+}
+
+interface SelectionMenuPluralLabels {
+  /** CLDR plural categories. `{count}` is replaced by the image count. Only `other` is required. */
+  zero?: string;
+  one?: string;
+  two?: string;
+  few?: string;
+  many?: string;
+  other?: string;
```
eszlamczyk: Acording to what docs state, what was agreed upon, I believe other should be the only one required
```suggestion
  other: string;
```

so we don't fall back to english

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/SelectionActionMode.kt
```diff
@@ -22,11 +22,44 @@ private const val MENU_ITEM_COPY_IMAGE_URL = 1001
...
+        templates.getOrNull(index)
+      }
+      ?: if (count == 1) {
+        copyImageUrlLabel.ifEmpty { "Copy Image URL" }
+      } else {
+        copyImageUrlsLabel.ifEmpty { "Copy {count} Image URLs" }
+      }
+  return template.replace("{count}", count.toString())
+}
+
```
eszlamczyk, line 51: This is a big blocker: 

Assume we have 101 images (a bit irrational but in case) then the index would calculate to (101 - 1 ) % 100 + 1 = 1

and we would get Copy Image Url (even though we have 101 images).

I think the safest and _good enough_ solution would be to resolve for "other" in case of count > 100

If you'd like to pursuit the "other if greater than 100" solution, please document it

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/SelectionActionMode.kt
```diff
@@ -61,7 +94,12 @@ fun createSelectionActionModeCallback(
         textView.selectionStart >= 0 &&
         textView.selectionEnd > textView.selectionStart
       ) {
-        menu.add(Menu.NONE, MENU_ITEM_COPY_MARKDOWN, Menu.NONE, "Copy as Markdown")
+        menu.add(
+          Menu.NONE,
+          MENU_ITEM_COPY_MARKDOWN,
+          Menu.NONE,
+          selectionMenuConfig.copyAsMarkdownLabel.ifEmpty { "Copy as Markdown" },
```
eszlamczyk: In the previous iteration I asked for these fallbacks to be removed. Keeping the ifEmpty blocks gives us two sources of truth: whatever JS resolves the defaults to, and whatever's written here. If someone updates the wording on the JS side, the default values drift - and a reader of the native code still thinks the label is "Copy as Markdown" when it's actually something else.

Let's strip the sentinels from all the native render paths. The only `?: ""` worth keeping is in parseSelectionMenuConfig (MarkdownViewManagerUtils.kt), which exists purely because ReadableMap.getString() returns String?. Everything else should trust the JS-resolved value.

Sites to clean up: SelectionActionMode.kt:56,58,101, TableContainerView.kt:…, MathContainerView.kt:…, plus iOS/macOS equivalents.

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
eszlamczyk: The same argument as in kotlin side, we don't really want the defensive programming here, due to differing sources of truth

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils.h
```diff
@@ -9,8 +9,36 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// Uses the precomputed plural templates (count 0..100, wrapping with period 100
+/// for larger counts) when present, otherwise the singular/`{count}` templates.
+/// Labels are resolved JS-side, so they are always concrete strings.
+static inline NSString *ENRMResolveImageURLsTitle(ENRMSelectionMenuConfig config, NSUInteger count)
+{
+  NSString *titleTemplate = nil;
+  NSString *packed = config.copyImageUrlPluralTemplates;
+  if (packed.length > 0) {
+    NSArray<NSString *> *templates = [packed componentsSeparatedByString:@"\x1f"];
+    NSUInteger index = count == 0 ? 0 : (count <= 100 ? count : ((count - 1) % 100) + 1);
```
eszlamczyk: Same as for android

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextNativeComponent.ts
```diff
@@ -230,6 +230,13 @@ export interface ContextMenuItemConfig {
 export interface SelectionMenuConfig {
   copyAsMarkdown: boolean;
   copyImageUrl: boolean;
+  copyLabel: string;
+  copyAsMarkdownLabel: string;
+  copyImageUrlLabel: string;
+  copyImageUrlsLabel: string;
+  // Precomputed plural templates (count 0..100) joined by U+001F. Empty when no
+  // pluralLabels are set; native then uses copyImageUrlLabel/copyImageUrlsLabel.
+  copyImageUrlPluralTemplates: string;
```
eszlamczyk: Nit: this could just be `ReadonlyArray<string>` (same change at `EnrichedMarkdownNativeComponent.ts:239`). Codegen already produces `std::vector<std::string>` for that - see `mentionIndicators` in `EnrichedMarkdownTextInputNativeComponent.ts:236 → Props.h:2798`. Drops the `\u001F` separator constant, the `templates.join(...)` in buildPluralTemplates, and the componentsSeparatedByString: / split(PLURAL_SEPARATOR) in native consumers (`SelectionActionMode.kt:50`, `EditMenuUtils.h:30`). Same wire-format cost, strictly more honest types

packages/react-native-enriched-markdown/src/types/MarkdownTextProps.ts
```diff
@@ -20,17 +20,67 @@ export interface ContextMenuItem {
...
 }
 
+/**
+ * Plural forms for the "Copy Image URL(s)" action, keyed by CLDR plural
+ * category. The right form is chosen at runtime from the number of selected
+ * images using `Intl.PluralRules` (resolved against the app's default locale).
+ *
+ * Each value is a template where the `{count}` token is replaced with the
+ * number of selected images. Only `other` is required by CLDR; any category
+ * left `undefined` falls back to `other`.
```
eszlamczyk, line 30: While you're fixing other stuff its worth tagging that this is technically speaking incorrect as value for 'one'  fallbacks to the default label (non-plular). It would be lovely if you could fix that along the way

eszlamczyk (changes requested):
Hi @marcellov7! Most of the fixes look clean, unfortunately I do have to block this pr second time, due to mostly small and quick-to-fix issues. You can tag me when you're done and Im gonna review and verify everything then (and hopefully we'll be able to merge it tomorrow so it can be added to the day after tomorrow's nightly release)

apps/react-native-example/.rnstorybook/stories/components/EnrichedMarkdownText/props/SelectionMenu.stories.tsx
marcellov7: Thanks! 🙂

docs/API_REFERENCE.md
marcellov7: Done in ad32877. Default shown as `{}` with per-field defaults in the shape below.

docs/API_REFERENCE.md
marcellov7: Done in ad32877. `other` is now required; the other categories fall back to it, so the menu never reaches the English default once pluralLabels is set.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/SelectionActionMode.kt
marcellov7: Good catch — fixed in ad32877. Dropped the period-100 wrap entirely: the table is indexed by exact count for 0..100, and counts > 100 now resolve to the `other` form (copyImageUrlsLabel). Documented in buildPluralTemplates and the native helpers.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/SelectionActionMode.kt
marcellov7: Done in ad32877. Stripped the ifEmpty sentinels from all native render paths (SelectionActionMode, table & math views, iOS/macOS edit menus). Only `?: ""` in parseSelectionMenuConfig remains, since ReadableMap.getString() is nullable.

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils.h
marcellov7: Done in ad32877. The helper now takes the templates array directly (no separator/split) and uses the JS-resolved labels with no English literal.

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextNativeComponent.ts
marcellov7: Done in ad32877. Switched both specs to `ReadonlyArray<string>`; dropped the U+001F separator, the join in buildPluralTemplates, and the split in SelectionActionMode.kt / EditMenuUtils.h. Native now reads std::vector / ReadableArray.

packages/react-native-enriched-markdown/src/types/MarkdownTextProps.ts
marcellov7: Done in ad32877. The 'one' category no longer falls back to the non-plural label — it falls back to `other`. The singular `label` is used only for the literal single-image case (index 1).

packages/react-native-enriched-markdown/ios/utils/EditMenuUtils+macOS.m
marcellov7: Agreed — removed the defensive programming entirely in ad32877 (reverted the earlier NSLocalizedString change). `resolveMenuLabel` is gone; the macOS edit menu and the table/math menus now use the JS-resolved labels directly, so there's a single source of truth.

marcellov7:
Thanks again @eszlamczyk — addressed the follow-up round in `ad32877`. All threads resolved.

- **`pluralLabels.other` is now required**; the other CLDR categories fall back to `other`, so the menu never reaches the English default once `pluralLabels` is set. The `one` form no longer borrows the non-plural single-image `label` (the singular `label` is used only for the literal 1-image case).
- **Wire format → `ReadonlyArray<string>`** (codegen `std::vector<std::string>`): dropped the `U+001F` separator, the `join` in `buildPluralTemplates`, and the `split` in the native consumers.
- **count > 100 fix (blocker)**: removed the period-100 wrap; counts > 100 now resolve to the `other` form (`copyImageUrlsLabel`) instead of wrapping back to `one`. Documented in `buildPluralTemplates` and the native helpers.
- **Removed the defensive English-literal fallbacks from all native render paths** (Android `ifEmpty`, iOS `resolveLabel`, macOS `resolveMenuLabel`, table/math menus). JS is the single source of truth; only `parseSelectionMenuConfig` keeps `?: ""` because `ReadableMap.getString()` is nullable.
- **Docs**: default shown as `{}` (per-field defaults in the shape), `other` marked required.

One thing I couldn't verify locally: passing `ReadonlyArray<string>` inside the `selectionMenuConfig` struct on Android (read via `getArray`) — followed the `mentionIndicators` precedent you pointed to. As before, I couldn't run the native builds / clang-format / ktlint, so please double-check the native formatting.

eszlamczyk (approved):
Everything looks great to me! Everything should be in next _nightly_ release

PR #422, opened 2026-06-23
fix(android): measure rendered plain text in InputMeasurementStore.initialMeasure, not raw markdown

### LINKED ISSUE #421: Android: mention/link placeholders inflate EnrichedMarkdownTextInput auto-grow height on first render (initialMeasure measures raw markdown)
## Bug description

On Android, an auto-grow `EnrichedMarkdownTextInput` (`scrollEnabled={false}`) reports a height that is too tall on first render, leaving empty space below the content. The height self-corrects on the first text change (typing one character). iOS is unaffected.

This is the same family as #281 and #366, but on the **input** component (`InputMeasurementStore`), which those fixes didn't cover.

## Root cause

`InputMeasurementStore.initialMeasure` measures the **raw markdown source** instead of the rendered plain text:

```kotlin
// android/.../input/layout/InputMeasurementStore.kt
val text = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
return measure(width, text, paint)
```

For a mention/link, the source is e.g. `[Label](placeholder://a.very.long.token.address.here)`. The URL is invisible when rendered but is counted by the first measure, so the text is over-estimated into extra lines. Yoga caches that first measure until an input change forces a re-measure (which goes through `store()` using the already-parsed `view.text`), which is why typing fixes it.

## Reproduction

```tsx
<ScrollView>
  <EnrichedMarkdownTextInput
    scrollEnabled={false}
    style={{ flexGrow: 1 }}
    defaultValue={"Hi [Label](placeholder://a.very.long.token.address.here)!"}
  />
</ScrollView>
```

Open the screen: the input is taller than its single line of rendered text (empty space below). Type one character: it snaps to the correct height.

Observed: ~**652dp on open → ~554dp after one keystroke**, same content.

## Expected

The first measure should match the rendered text height (iOS already does this).

## Environment

- react-native-enriched-markdown 0.6.0
- React Native 0.85.x (New Architecture)
- Android only (iOS correct)

## Suggested fix

Parse the markdown to plain text before measuring in `initialMeasure`, using the existing `InputParser.parseToPlainTextAndRanges(...)` (already used by `setValueFromJS`). PR follows.

andreavrr:
## Summary

On Android, an auto-grow `EnrichedMarkdownTextInput` (`scrollEnabled={false}`) reports a height that is too tall on first render, leaving empty space below the content. It self-corrects on the first keystroke. iOS is unaffected.

The first measure counts the hidden URLs of mention/link placeholders, which aren't rendered. This parses the markdown into its rendered plain text **before** measuring, using the parser the library already uses elsewhere.

Fixes #421

## Problem

With a `defaultValue` that contains a mention/link:

```tsx
<ScrollView>
  <EnrichedMarkdownTextInput
    scrollEnabled={false}
    style={{ flexGrow: 1 }}
    defaultValue={"Hi [Label](placeholder://a.very.long.token.address.here)!"}
  />
</ScrollView>
```

- Before: the input is taller than its rendered text on open (empty space below); it snaps to the correct height on the first keystroke.
- Observed: ~652dp on open → ~554dp after one keystroke, same content.

## Root cause

`InputMeasurementStore.initialMeasure` measures the raw markdown string instead of the rendered plain text:

```kotlin
val text = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
return measure(width, text, paint)
```

For a mention the source is `[Label](placeholder://...)`: the URL is invisible on screen but counted by the measure, over-estimating the height into extra lines. Yoga caches this first measure until a text change forces a re-measure (which goes through `store()` with the already-parsed `view.text`), which is why typing corrects it.

## Fix

Parse the markdown into rendered plain text before measuring, using the existing `InputParser.parseToPlainTextAndRanges(...)` (the same one used by `setValueFromJS`):

```kotlin
import com.swmansion.enriched.markdown.input.formatting.InputParser
// ...
val markdown = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
val text = InputParser.parseToPlainTextAndRanges(markdown).plainText
```

## Why this approach

- Reuses the library's existing parser, so the measure uses the same representation as the actual render (no duplicated parsing logic).
- Touches only the `initialMeasure` path (the first measure); subsequent measures already go through `store()` with the parsed text and are unchanged.
- Same family as #281 / #366 (incorrect Android measurement, iOS correct), but on the input component those fixes didn't cover.

## Test plan

- Auto-grow input (`scrollEnabled={false}`) with a `defaultValue` containing a mention/link with a long URL, inside a `ScrollView`.
- Before: extra height on open, snaps shorter on the first keystroke.
- After: correct height on open; no change on keystroke.
- iOS: no regression (the path is Android-only).

## Notes

- Android-only; iOS measures via TextKit on the already-rendered attributed string, so it doesn't have the issue.
- No public API change; internal to measurement.

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
@@ -9,6 +9,7 @@ import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.uimanager.PixelUtil
 import com.facebook.yoga.YogaMeasureMode
 import com.facebook.yoga.YogaMeasureOutput
+import com.swmansion.enriched.markdown.input.formatting.InputParser
 import java.util.concurrent.ConcurrentHashMap
 
 object InputMeasurementStore {
@@ -93,7 +94,12 @@ object InputMeasurementStore {
     width: Float,
     props: ReadableMap?,
   ): Long {
-    val text = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
+    // Measure the rendered plain text, not the raw markdown. A mention link such as
+    // [Label](placeholder://x) hides its URL in the source, so measuring the markdown counts those
+    // invisible characters and over-estimates the height into extra lines, until the next text
+    // change forces a re-measure. Parsing first matches what the editor actually renders.
+    val markdown = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
+    val text = InputParser.parseToPlainTextAndRanges(markdown).plainText
     val fontSize = props?.getDouble("fontSize")?.toFloat() ?: 16f
     val spSize = kotlin.math.ceil(PixelUtil.toPixelFromSP(fontSize).toDouble()).toFloat()
 


eszlamczyk:
Hi @andreavrr thanks for the issue! Unfortunately everything I try I cannot reproduce the issue on the main branch, could you either provide a full reproduction code you tested the issue with (via either a minimal repo you can link or via code snippet or full Entrypoint file). When I'm gonna be able to verify the fix we'll probably merge it (code looks decent)

andreavrr:
@eszlamczyk thanks for taking a look! Here's a minimal repo:

https://github.com/andreavrr/enriched-markdown-android-measure-mre

Expo SDK 56, RN 0.85.3, New Architecture, `react-native-enriched-markdown@0.6.0` (unpatched). Run with `npx expo run:android` (dev build, Expo Go won't work).

A few conditions all have to hold at once, which is probably why it didn't reproduce for you:

- **Android only** (iOS measures via TextKit on the already-rendered text, so it's correct).
- The input must be **auto-grow** (`scrollEnabled={false}`) **and inside a `ScrollView`**. Outside a `ScrollView` the `AT_MOST` clamp hides the over-estimate; inside, the child is measured with `UNDEFINED`, so the raw (too tall) height is what you get.
- `defaultValue` must contain a **mention/link with a long URL**, e.g. `[Label](placeholder://a.very.long.token.address)`. The over-estimate scales with the hidden URL length, so a short URL barely shows.

Steps (in the repo):

1. `npx expo run:android`, tap **Open editor**.
2. Scroll to the bottom of the message — there's extra empty space below the last line.
3. Type one character — it snaps to the correct height and the empty space disappears.

The repo runs the released `0.6.0` (no patch), so it shows the bug; applying this PR fixes it (the first measure then matches the rendered plain text). Happy to adjust the repro if you need anything else.

eszlamczyk:
## Before

https://github.com/user-attachments/assets/61c34663-85df-4e47-951a-c44e4f9b0ee6

## After
https://github.com/user-attachments/assets/45f90e5f-783d-49aa-a1a4-1b2c98605605

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
```diff
@@ -93,7 +94,12 @@ object InputMeasurementStore {
     width: Float,
     props: ReadableMap?,
   ): Long {
-    val text = props?.getString("defaultValue") ?: props?.getString("placeholder") ?: "I"
+    // Measure the rendered plain text, not the raw markdown. A mention link such as
+    // [Label](placeholder://x) hides its URL in the source, so measuring the markdown counts those
+    // invisible characters and over-estimates the height into extra lines, until the next text
+    // change forces a re-measure. Parsing first matches what the editor actually renders.
```
eszlamczyk, line 100: Small nit for the future: We prefer a bit smaller comments so maybe something like 

```
 // Parse first: raw markdown counts hidden link URLs (e.g. [Label](url)) and over-measures.
```

would be better

eszlamczyk (approved):
Code looks clean, and the fix works! Thanks for your help

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
andreavrr: Thanks a lot for merging and for the review! Noted on the shorter comments, I'll keep them more concise going forward. 🙏

PR #427, opened 2026-06-24
feat(android): add parser AST JNI bridge

graszka22:
### What/Why?
Just copied jni stuff from react-native-enriched-markdown. 

### Testing
App example and e2e tests later. 

### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (commented):
Let's consider using fbjni here instead of raw JNI. The current `createJavaNode` has a few problems - leaked local refs (`nodeClass`, `mapClass`, `listClass` are never cleaned up), repeated `FindClass`/`GetMethodID` on every node, and since it's recursive, we'll hit the JNI local ref table limit on larger documents.

fbjni handles all of this automatically (RAII refs, cached lookups, exception safety). We planned to migrate the jni-adapter in `react-native-enriched-markdown` to fbjni but didn't have time to pick it up - this is a good opportunity to do it right from the start.

hryhoriiK97 (approved):
LGTM!

PR #429, opened 2026-06-24
feat(android): add UI foundation text utils

graszka22:
### What/Why?
Stuff related to text and fonts. Partially copied from react-native-enriched-markdown.

### Testing
App example and e2e tests.

### PR Checklist

- [ ] E2E tests are passing


packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/TypefaceUtils.kt
```diff
@@ -0,0 +1,120 @@
...
+
+  fun parseFontWeight(fontWeight: String?): Int =
+    when (val normalized = fontWeight?.lowercase()) {
+      null, "", "normal", "regular", "400" -> Typeface.NORMAL
+      "bold", "700", "800", "900" -> Typeface.BOLD
+      "600", "semibold" -> Typeface.BOLD
+      "500", "medium" -> 500
+      "300", "light" -> 300
+      else -> normalized.toIntOrNull()?.coerceIn(1, 1000) ?: Typeface.NORMAL
+    }
```
hryhoriiK97, line 33: `parseFontWeight` returns `Typeface.BOLD` (1) for "bold"/"700"/"800"/"900", but `createTypefaceWithWeight` treats anything below 100 as the legacy path where `1 < 600` maps to `NORMAL`. So bold text ends up rendering as regular on API 28+.

Since `createTypefaceWithWeight` already handles the pre-28 fallback (`weight >= 600 → BOLD`), we can just return real weight values here:

```
  fun parseFontWeight(fontWeight: String?): Int =
    when (val normalized = fontWeight?.lowercase()) {
      null, "", "normal", "regular", "400" -> 400
      "300", "light" -> 300
      "500", "medium" -> 500
      "600", "semibold" -> 600
      "bold", "700" -> 700
      "800" -> 800
      "900" -> 900
      else -> normalized.toIntOrNull()?.coerceIn(1, 1000) ?: 400
    }
```

packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/TypefaceUtils.kt
```diff
@@ -0,0 +1,120 @@
+package com.swmansion.enriched.markdown.utils.text
+
+import android.content.Context
+import android.graphics.Typeface
+import android.os.Build
+
+object TypefaceUtils {
+  private val typefaceCache = mutableMapOf<String, Typeface>()
```
hryhoriiK97: `typefaceCache` is a plain `mutableMapOf()` (LinkedHashMap) which isn't thread-safe. If typeface loading happens off the main thread, concurrent access could corrupt the map. Should we use `ConcurrentHashMap` here? 🤔

PR #433, opened 2026-06-24
feat(android): add EnrichedMarkdownText entry point

graszka22:
### What/Why?
Mostly copied from react-native-enriched-markdown. The main EnrichedMarkdownText view. This will not be used by end users - it's internal for the library and will be wrapped by compose wrapper and maybe used in the future by react-native-enriched-markdown.
Also the parsing, rendering etc. is run on a separate thread.

### Testing
Example app and e2e tests.

### PR Checklist

- [ ] E2E tests are passing


packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
```diff
@@ -0,0 +1,230 @@
...
+    var md4cFlags: Md4cFlags = Md4cFlags.DEFAULT
+      private set
+
+    private var pendingStyledText: CharSequence? = null
+    private var selectionColor: Int? = null
+    private var selectionHandleColor: Int? = null
+    private var isSelectable = true
+
+    init {
+      markdownStyle = StyleConfig.default(context)
```
hryhoriiK97: `markdownStyle` is already initialized to `StyleConfig.default(context)` in the property declaration. This second call in `init` creates a duplicate `StyleConfig` that gets thrown away. Should we drop the one in `init`?

packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdownText.kt
```diff
@@ -0,0 +1,230 @@
...
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0,
+  ) : AppCompatTextView(context, attrs, defStyleAttr) {
+    private val parser = Parser.shared
+    private val renderer = Renderer()
+    private var onLinkPressCallback: ((String) -> Unit)? = null
+    private var onLinkLongPressCallback: ((String) -> Unit)? = null
+
+    private val mainHandler = Handler(Looper.getMainLooper())
+    private var currentRenderId = 0L
```
hryhoriiK97: This is read from background threads (in `isCancelled` and inside the task) but written on the main thread. Should we add `@Volatile` here so background threads see the latest value and skip cancelled work promptly?

PR #436, opened 2026-06-24
fix(android): preserve blank lines between paragraphs in input parser (parity with iOS)

andreavrr:
## Summary

On **Android**, an `EnrichedMarkdownTextInput` loaded with a value that contains blank lines between
paragraphs drops them: every run of blank lines collapses and the paragraphs render run together.
**iOS keeps every blank line.**

The Android input parser rebuilds the plain text from the md4c AST and emits no separator between
paragraph nodes. This re-reads the original blank-line runs from the source and replays them at each
paragraph break, so the rendered text matches the source the way it already does on iOS.

## Problem

With a `defaultValue` that contains blank lines (plain text, no markdown features):

```tsx
<EnrichedMarkdownTextInput
  scrollEnabled={false}
  defaultValue={"First paragraph.\n\n\n\nSecond paragraph.\n\nThird paragraph."}
/>
```

- **Before (Android):** the paragraphs render run together (`First paragraph.Second paragraph...`,
  `...paragraph.Third paragraph...`), all blank lines dropped.
- **iOS:** a large gap (three blank lines), then a small gap (one blank line), preserved.

Minimal reproducible example (Expo SDK 56, RN 0.85.3, New Architecture, `react-native-enriched-markdown@0.6.0`
unpatched): https://github.com/andreavrr/enriched-markdown-android-blank-lines-mre

## Root cause

`InputParser.walkNode` builds the plain text by appending `Text` and `LineBreak` nodes but emits
nothing between `Paragraph` nodes:

```kotlin
for (child in node.children) {
  walkNode(child, plainText, ranges, activeStyles)
}
```

A blank line is a paragraph break in CommonMark, so md4c splits the text into separate `Paragraph`
nodes (and collapses any run of blank lines into a single break). With no separator emitted, the
paragraphs concatenate. iOS doesn't hit this because its input parser derives the plain text from the
original source buffer (removing only the syntax markers), so blank lines, which are content rather
than syntax, survive.

## Fix

Re-emit the source blank lines between paragraphs during the walk. md4c collapses the runs in the AST,
but the source string is still available in `parseToPlainTextAndRanges`, so read the real run lengths
from it and replay them at each paragraph break:

```kotlin
private val PARAGRAPH_BREAK_RE = Regex("\\n[ \\t]*(?:\\n[ \\t]*)+")
// ...
val blankRuns =
  ArrayDeque(
    PARAGRAPH_BREAK_RE
      .findAll(completed.trim())
      .map { match -> match.value.count { it == '\n' } }
      .toList(),
  )
// ...
for ((index, child) in node.children.withIndex()) {
  if (index > 0 && child.type == NodeType.Paragraph && plainText.isNotEmpty()) {
    plainText.append("\n".repeat(blankRuns.removeFirstOrNull() ?: 2))
  }
  walkNode(child, plainText, ranges, activeStyles, blankRuns)
}
```

## Why this approach

- Mirrors what iOS already produces (blank lines preserved) without reimplementing the iOS
  buffer-subtraction strategy on Android.
- Mention/style offsets stay consistent: they are computed live from `plainText.length` during the
  same walk, so emitting more newlines never desyncs them.
- Graceful fallback: a paragraph break with no matching source run (exotic input) emits a single
  blank line (the prior behaviour), never a crash.
- Touches only the parse path; serialization (`getMarkdown`) already preserves blank lines verbatim,
  so the round-trip stays consistent.

## Test plan

- Load a `defaultValue` with single and multiple blank lines between paragraphs (see the MRE above).
- **Before:** the paragraphs run together on Android.
- **After:** blank lines are preserved (single and multiple runs), matching iOS; a blank line deletes
  with one backspace.
- **iOS:** no regression (the path is Android-only).

## Notes

- Android-only; iOS derives the plain text from the source buffer, so it already preserves blank lines.
- md4c collapses multiple consecutive blank lines in the AST; the original run length is recovered
  from the source string, so 2+ blank lines round-trip as typed.
- No public API change; internal to the input parser.

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
@@ -13,6 +13,9 @@ data class ParseResult(
 )
 
 object InputParser {
+  // A source paragraph break: a newline followed by one or more blank lines.
+  private val PARAGRAPH_BREAK_RE = Regex("\\n[ \\t]*(?:\\n[ \\t]*)+")
+
   fun parseToPlainTextAndRanges(markdown: String): ParseResult {
     if (markdown.isEmpty()) {
       return ParseResult("", emptyList())
@@ -25,7 +28,17 @@ object InputParser {
     val plainText = StringBuilder()
     val ranges = mutableListOf<FormattingRange>()
 
-    walkNode(ast, plainText, ranges, ArrayDeque())
+    // md4c collapses any blank-line run into a single break, unlike iOS which keeps them.
+    // Re-read the real runs so each break replays its original number of newlines.
+    val blankRuns =
+      ArrayDeque(
+        PARAGRAPH_BREAK_RE
+          .findAll(completed.trim())
+          .map { match -> match.value.count { it == '\n' } }
+          .toList(),
+      )
+
+    walkNode(ast, plainText, ranges, ArrayDeque(), blankRuns)
 
     return ParseResult(plainText.toString(), ranges)
   }
@@ -35,6 +48,7 @@ object InputParser {
     plainText: StringBuilder,
     ranges: MutableList<FormattingRange>,
     activeStyles: ArrayDeque<ActiveStyle>,
+    blankRuns: ArrayDeque<Int>,
   ) {
     val styleType = nodeTypeToStyleType(node.type)
 
@@ -49,8 +63,12 @@ object InputParser {
       plainText.append("\n")
     }
 
-    for (child in node.children) {
-      walkNode(child, plainText, ranges, activeStyles)
+    for ((index, child) in node.children.withIndex()) {
+      // Keep the source's blank lines between paragraphs (md4c drops them, iOS keeps them).
+      if (index > 0 && child.type == NodeType.Paragraph && plainText.isNotEmpty()) {
+        plainText.append("\n".repeat(blankRuns.removeFirstOrNull() ?: 2))
+      }
+      walkNode(child, plainText, ranges, activeStyles, blankRuns)
     }
 
     if (styleType != null) {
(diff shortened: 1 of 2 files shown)


eszlamczyk (changes requested):
Hi @andreavrr thanks for the pr!

Found pretty big issue though: 
The interactive surface (StyleType.kt, toolbar, toggleInlineStyle) only exposes the six inline styles: BOLD/ITALIC/UNDERLINE/STRIKETHROUGH/LINK/SPOILER. No headings, no block-quotes, no lists, no code blocks. Soft enter just appends \n (LineBreak), it doesn't create a new block.

But any markdown string can still reach InputParser.parseToPlainTextAndRanges via defaultValue (EnrichedMarkdownTextInputManager.kt:108) or programmatic setValueFromJS (EnrichedMarkdownTextInputView.kt:564) - both pass the raw JS string straight to md4c with no pre-filter. So block elements are reachable; they just won't be rendered with their block styling, they'll fall out as plain text from the walk.

For example:
```
"P1.\n\n# H\n\nP2."
```
(left is android, incorrect; right is IOS, correct)
Image

I'd solution like this:
Instead of checking if the node is `paragraph`, check against a helper function
```kt 
  private fun MarkdownASTNode.NodeType.isTopLevelBlock(): Boolean =
    when (this) {
      NodeType.Paragraph,
      NodeType.Heading,
      NodeType.Blockquote,
      NodeType.UnorderedList,
      NodeType.OrderedList,
      NodeType.CodeBlock,
      NodeType.ThematicBreak,
      NodeType.Table,
      NodeType.LatexMathDisplay -> true
      else -> false
    }
```

this can be done as either:
1. As an extension on NodeType in `MarkdownASTNode.kt` - reusable across the codebase, future-proof if other walkers need it.
2. As a private fun inside InputParser - narrower scope, no API surface added, matches the existing nodeTypeToStyleType helper right below in the same file.

I'd lean into `1.` but Im gonna leave the choice for you

andreavrr:
Thanks for the review @eszlamczyk ! 
Went with option 1, which is also my preference: I kept it centralized and reusable in `MarkdownASTNode.kt`, and the walk now checks `child.type.isTopLevelBlock()`.

eszlamczyk (approved):
Looks great!

PR #442, opened 2026-06-25
feat(input): cross-platform contentInset prop (iOS textContainerInset + Android scroll container)

### LINKED ISSUE #426: Feature request: `contentInset` prop for `EnrichedMarkdownTextInput` (non-clipping internal inset)
### Problem

There's no way to give `EnrichedMarkdownTextInput` an internal, non-clipping vertical inset. There is no `contentInset` prop, and `style` padding is not a substitute:

- **Android:** it maps to `TextView` padding, which clips the text at the padding band while the content scrolls.
- **iOS:** the input doesn't apply it as a text inset at all (no `padding` prop on the input; `textContainerInset` is left at the default and isn't exposed).

So a top/bottom cushion that scrolls together with the content (e.g. under a translucent header, or above a keyboard-sticky toolbar) can't be achieved today.

### Proposal

Add a `contentInset` prop (`{ top, right, bottom, left }`) to `EnrichedMarkdownTextInput`:

- **iOS:** map it to `UITextView.contentInset`. Since `UITextView` is a `UIScrollView`, this is non-clipping by nature.
- **Android:** apply an equivalent non-clipping inset at the scroll/layout level (not as `TextView` padding, which clips mid-scroll).

### Prior art

The closed PR #208 added `contentInset` to the read-only display component (`EnrichedMarkdown` / `EnrichedMarkdownText`), **not** the input. This request is specifically for the editable `EnrichedMarkdownTextInput`.

### Use case

Rich-text editor with a translucent navigation header and a keyboard-sticky formatting toolbar: it needs a top/bottom cushion that doesn't clip the text at a fixed band.

andreavrr:
Closes #426

## Summary

Adds a cross-platform `contentInset` prop to `EnrichedMarkdownTextInput`: a content cushion that is part of the scrollable content and scrolls away with the text, instead of clipping the way style padding does today. `top` pushes the first line down (e.g. below a translucent header); `bottom` keeps room past the last line. It exposes the iOS `UITextView.textContainerInset` behaviour and brings the same to Android.

## Problem

There is no prop for a content cushion that scrolls away with the text. The only workaround is style padding, and on **both iOS and Android** `paddingVertical` on the editor is clipped while the field scrolls (the cushion eats the text), and a long autofocused value does not reveal the end. iOS has `UITextView.textContainerInset` for exactly this, but it is not exposed; Android has no equivalent.

```tsx
<EnrichedMarkdownTextInput
  autoFocus
  defaultValue={LONG_TEXT}
  scrollEnabled
  style={{ flex: 1, fontSize: 18, paddingVertical: 48 }}
/>
```

Minimal reproducible example (Expo SDK 56, RN 0.85.3, New Architecture):
https://github.com/andreavrr/enriched-markdown-content-inset-mre

## API

```ts
contentInset?: { top: number; bottom: number; left: number; right: number };
```

## Fix

- **iOS:** map `contentInset` to `UITextView.textContainerInset`, and grow the caret-scroll target rect by the bottom inset so the caret stays revealed while typing.
- **Android:** the ViewManager returns a `NestedScrollView` container whose child is the editor (`wrap_content`, non-scrolling). The inset is the editor's own padding — part of the scrolled content — so it scrolls away without clipping (Android `TextView` couples its draw clip to the padding while the field scrolls internally). The container scrolls and reveals the caret. Border/background/padding go on the container (the visible frame). A manual measure/layout pass and `clipToOutline` cover what RN does not here (it swallows the non-Fabric child's `requestLayout`, and RN parents default to `overflow: visible`). The shadow node stays an auto-grow measuring leaf, so consumers without an explicit height still grow to content.

## Test plan

- iOS and Android, `contentInset` with a long autofocused value: the cushion scrolls away (no clip) and `autoFocus` reveals the end.
- Auto-grow (no explicit height) and `maxHeight` consumers still size correctly.
- Existing props unaffected (border/background, `scrollEnabled`, selection menus, etc.).

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
@@ -38,36 +38,37 @@ import com.swmansion.enriched.markdown.utils.input.MarkdownStyleParser
 
 @ReactModule(name = EnrichedMarkdownTextInputManager.NAME)
 class EnrichedMarkdownTextInputManager :
-  SimpleViewManager<EnrichedMarkdownTextInputView>(),
-  EnrichedMarkdownTextInputManagerInterface<EnrichedMarkdownTextInputView> {
-  private val delegate: ViewManagerDelegate<EnrichedMarkdownTextInputView> =
+  SimpleViewManager<EnrichedMarkdownTextInputScrollView>(),
+  EnrichedMarkdownTextInputManagerInterface<EnrichedMarkdownTextInputScrollView> {
+  private val delegate: ViewManagerDelegate<EnrichedMarkdownTextInputScrollView> =
     EnrichedMarkdownTextInputManagerDelegate(this)
 
-  override fun getDelegate(): ViewManagerDelegate<EnrichedMarkdownTextInputView> = delegate
+  override fun getDelegate(): ViewManagerDelegate<EnrichedMarkdownTextInputScrollView> = delegate
 
   override fun getName(): String = NAME
 
-  override fun createViewInstance(reactContext: ThemedReactContext): EnrichedMarkdownTextInputView =
-    EnrichedMarkdownTextInputView(reactContext)
+  override fun createViewInstance(reactContext: ThemedReactContext): EnrichedMarkdownTextInputScrollView =
+    EnrichedMarkdownTextInputScrollView(reactContext)
 
   override fun updateState(
-    view: EnrichedMarkdownTextInputView,
+    view: EnrichedMarkdownTextInputScrollView,
     props: ReactStylesDiffMap?,
     stateWrapper: StateWrapper?,
   ): Any? {
-    view.stateWrapper = stateWrapper
+    // Route Fabric state to the editor so its auto-grow loop still drives the re-measure.
+    view.input.stateWrapper = stateWrapper
     return super.updateState(view, props, stateWrapper)
   }
 
-  override fun onAfterUpdateTransaction(view: EnrichedMarkdownTextInputView) {
+  override fun onAfterUpdateTransaction(view: EnrichedMarkdownTextInputScrollView) {
     super.onAfterUpdateTransaction(view)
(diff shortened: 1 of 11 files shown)


andreavrr:
## Before / after

|  | Before (`paddingVertical` clips) | After (`contentInset`) |
| :--: | :--: | :--: |
| **iOS** | <video src="https://github.com/user-attachments/assets/7292ddf9-dc67-48c0-a033-14475efcc844"></video> | <video src="https://github.com/user-attachments/assets/d71c2a8c-19ae-40b6-8354-280d5e17dc19"></video> |
| **Android** | <video src="https://github.com/user-attachments/assets/0c78f254-a1b6-4d2c-a34a-84911bd1a667" ></video> | <video src="https://github.com/user-attachments/assets/4feecbbb-fbb0-4967-bc3f-630f06038a61" ></video> |

hryhoriiK97 (commented):
@andreavrr thank you for this PR! It will require more consideration. I will look into this next week and let you know how we want to handle it.

andreavrr:
Thanks a lot @hryhoriiK97! I figured, there's probably still quite a bit to sort out. Either way it feels like an important feature (well, more of a fix) on the UI side. Thanks especially for giving me a sense of the timeline, happy to make any changes whenever you get to it.

andreavrr:
Hi @hryhoriiK97, any updates on this?

hryhoriiK97:
> Hi @hryhoriiK97, any updates on this?

@andreavrr sorry for the late reply!

After looking into this, we have a few concerns. The contentInset name is an iOS-only ScrollView concept - Android has no equivalent. And on iOS, RN's TextInput already maps padding style to textContainerInset under the hood, so this essentially duplicates existing behavior under a new name.

The actual issue here is that Android's EditText clips padding during scroll (facebook/react-native#27623). This is a known platform limitation - even an attempt to bring contentInset to RN's Android ScrollView (facebook/react-native#49145) was closed due to bugs. Wrapping the EditText in a NestedScrollView works, but it changes the view hierarchy in ways that can cause edge cases with caret reveal, selection handles, Fabric layout, etc.

We won't be merging this as-is, but we'll keep the issue open. If we tackle it later, we'd rather fix padding to just work on Android than add a new prop. That said, if you have any other ideas or a different approach in mind, we're definitely open to hearing them!

PR #447, opened 2026-06-26
fix(android): InputRemend mishandles combined bold+italic *** delimiter

multiplechuks:
### What/Why?
InputRemend.complete() appended phantom closing asterisks when bold+italic (***) markdown was set directly as a raw value on Android.

The root cause: DELIMITER_PAIRS listed ** before * but had no entry for ***. When parsing ***bold italic***, the scanner consumed ** as one delimiter and * as another (pushing two separate stack entries). On the closing side, ** was matched first again — leaving the stack non-empty at EOF and causing extra closers to be appended (e.g. ***bold italic*** → ***bold italic*** + 6 asterisks).

Fix: Add DelimiterPair("***", "***", true) before ** in DELIMITER_PAIRS so triple-asterisk is matched atomically as a single unit.

### Testing
- Android unit tests added in InputRemendTest.kt covering balanced and unclosed *** cases
- Maestro E2E test added (bold_italic_raw_markdown_test.yaml) — sets ***Bold italic*** via the raw markdown input and asserts the screenshot renders correctly without phantom asterisks

| Before | After |
| - | - |
|  |  |

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputRemend.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputRemend.kt
@@ -9,6 +9,7 @@ object InputRemend {
 
   private val DELIMITER_PAIRS =
     arrayOf(
+      DelimiterPair("***", "***", true),
       DelimiterPair("**", "**", true),
       DelimiterPair("*", "*", true),
       DelimiterPair("_", "_", true),


hryhoriiK97 (changes requested):
Hey @multiplechuks, thanks for this! The fix itself is good - let's just strip the tests from this PR.

We already have `bold_italic_test.yaml` covering bold+italic rendering, so the new Maestro flow is redundant. And for the unit test - we'd rather set up `InputRemend` test coverage properly as its own thing rather than adding cases one-off per bugfix.

So just keep the one-liner in `InputRemend.kt` and we're good to merge.

hryhoriiK97 (approved):

PR #448, opened 2026-06-26
feat(android): add Compose font registry bridge

graszka22:
### What/Why?
In compose we have a new way of specifying fonts family: https://developer.android.com/develop/ui/compose/text/fonts
And we need a way to "translate" compose fonts into fonts passable to normal TextView 

### Testing
example app

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/android-enriched-markdown/compose/src/main/java/com/swmansion/enriched/markdown/compose/style/ComposeFontRegistry.kt
+++ b/packages/android-enriched-markdown/compose/src/main/java/com/swmansion/enriched/markdown/compose/style/ComposeFontRegistry.kt
@@ -1,0 +1,21 @@
+package com.swmansion.enriched.markdown.compose.style
+
+import android.graphics.Typeface
+import com.swmansion.enriched.markdown.utils.text.TypefaceUtils
+import java.util.WeakHashMap
+import java.util.concurrent.atomic.AtomicInteger
+
+internal object ComposeFontRegistry {
+  private val idGenerator = AtomicInteger()
+  private val typefaceToKey = WeakHashMap<Typeface, String>()
+  private val lock = Any()
+
+  fun register(typeface: Typeface): String =
+    synchronized(lock) {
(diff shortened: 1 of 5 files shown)


packages/android-enriched-markdown/compose/src/main/java/com/swmansion/enriched/markdown/compose/style/ComposeFontRegistry.kt
```diff
@@ -0,0 +1,20 @@
...
+import java.util.WeakHashMap
+import java.util.concurrent.atomic.AtomicInteger
+
+internal object ComposeFontRegistry {
+  private val idGenerator = AtomicInteger()
+  private val typefaceToKey =
+    Collections.synchronizedMap(WeakHashMap<Typeface, String>())
+
+  fun register(typeface: Typeface): String =
+    typefaceToKey.getOrPut(typeface) {
```
hryhoriiK97: Minor: getOrPut on a synchronizedMap isn't atomic, so in theory two threads could register the same typeface under different keys. Probably fine since this runs during composition (main thread), but a synchronized block would make it bulletproof if you want.

packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/TypefaceUtils.kt
```diff
@@ -65,6 +65,14 @@ object TypefaceUtils {
     }
   }
 
+  fun registerComposeFont(
+    key: String,
+    typeface: Typeface,
+  ): String {
+    typefaceCache["family:$key"] = typeface
+    return key
+  }
```
hryhoriiK97, line 73: nit: this returns key but the caller never uses the return value - could just be Unit.

```
fun registerComposeFont(key: String, typeface: Typeface) {
  typefaceCache["family:$key"] = typeface
}
```

hryhoriiK97 (approved):

PR #452, opened 2026-06-26
feat(android): expand native example app with compose demo

graszka22:
### What/Why?
Just an example app for android based on the rn example app.

### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (approved):
LGTM! One thing — assets/fonts/ is unused here since the styles use Font(R.font.xxx) (goes through res/font/). The asset path only gets hit from the View-based/RN bridge API. Safe to remove those duplicates.

PR #453, opened 2026-06-27
fix(ios): stop bold/italic handlers from overwriting link variant colors

hryhoriiK97:
### What/Why?

On iOS, bold/italic handlers unconditionally set `NSForegroundColorAttributeName` with `boldColor ?: baseTextColor`. When no explicit `boldColor`/`italicColor` is configured, this redundantly writes `baseTextColor` over any link variant color — so toggling bold on a mention makes it turn black until re-parse.

This removes the root cause: bold/italic now only write foreground color when explicitly configured. The reset phase already applies `baseTextColor` everywhere, so the fallback was always a no-op that happened to be destructive on link overlaps. Also removes the two-pass workaround from #446 since the conflict can no longer occur.

Aligns iOS with Android, which already skips the color write when `boldColor`/`italicColor` is unset.
### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/input/ENRMInputFormatter.mm
+++ b/packages/react-native-enriched-markdown/ios/input/ENRMInputFormatter.mm
@@ -136,8 +136,6 @@
     return;
   }
 
-  NSMutableArray<ENRMFormattingRange *> *deferredLinkRanges = [NSMutableArray array];
-
   for (ENRMFormattingRange *formattingRange in ranges) {
     if (formattingRange.range.length == 0 || NSMaxRange(formattingRange.range) > textLength) {
       continue;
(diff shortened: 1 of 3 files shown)


hryhoriiK97 (commented):
@andreavrr, thanks again for the fix in #446!

I found a simpler approach that eliminates the issue at its source: instead of deferring links to a second pass, we make bold/italic only set foreground color when `boldColor`/`italicColor` is explicitly configured. Since the reset phase already applies `baseTextColor` to the full range, the `?: baseTextColor` fallback was always redundant — and it's what caused the overwrite.

This brings iOS in line with Android (which already skips the color write when unset), removes the two-pass workaround, and makes ordering irrelevant.

andreavrr:
Thanks a lot for the review and the quick merge, and for the follow-up in #453! Really appreciate it 🙏

PR #454, opened 2026-06-27
feat(input): treat links/mentions as atomic for selection and caret

andreavrr:
### What/Why?

On `EnrichedMarkdownTextInput`, links (e.g. mentions) render as a single styled chip, but under the hood they're just a link range over normal editable text. So today the selection can land *inside* a link: you can select only part of a mention and toggle a style on those few characters, or place the caret in the middle of it. Styling part of a link splits the chip and produces inconsistent / serialization-breaking output, and a caret sitting inside a mention is confusing for an element that's meant to be atomic.

This makes links behave as a single atomic unit for selection on both iOS and Android, in the selection-change handler (`textViewDidChangeSelection` / `onSelectionChanged`):

- **Partial selection → whole link.** If a selection's start or end falls inside a link, it's expanded to cover the entire link, so a subsequent Bold/Italic/… always applies to the complete chip.
- **Caret inside a link → its end.** Tapping inside a mention moves the caret to the end of the link (e.g. `John| Doe` → `John Doe|`) instead of leaving it in the middle. Tapping right before/after the link is unchanged, so you can still type around it.

The guards are strict (`start < pos < end`), so the snapped selection/caret lands on a link boundary and doesn't re-trigger, and selecting exactly the link or placing the caret at its edges is a no-op.

### Testing

Manually on iOS and Android:
- Drag-select part of a mention → selection expands to the whole link.
- Tap inside a mention → caret jumps to its end.
- Select from plain text into a mention → the mention is fully included.
- Tap just before/after a mention → caret stays (you can type around it).

## Before / after

|  | Before | After |
| :--: | :--: | :--: |
| **First fix** | <video src="https://github.com/user-attachments/assets/60c8d707-7f17-4729-b3bc-2b05c7ba8f4d"></video> | <video src="https://github.com/user-attachments/assets/f2cf6fd1-e542-42b1-ba2d-607af716c7d3"></video> |
| **Second fix** | <video src="https://github.com/user-attachments/assets/87d17e3d-51fe-4245-9887-2961ecf6f4a7"></video> | <video src="https://github.com/user-attachments/assets/87ad3cba-3eed-460e-a903-64d83f8abf88"></video> |

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -276,26 +276,13 @@ class EnrichedMarkdownTextInputView(
     super.onSelectionChanged(selStart, selEnd)
     if (!isComponentReady || isDuringTransaction) return
 
-    // Links (e.g. mentions) are atomic: a selection can't cover only part of a link, and the caret
-    // can't sit inside one. Snap a partial selection to the whole link, and a caret inside it to its end.
-    if (selStart != selEnd) {
-      var ns = selStart
-      var ne = selEnd
-      formattingStore.rangeOfType(StyleType.LINK, ns)?.let { ns = minOf(ns, it.start) }
-      if (ne > 0) formattingStore.rangeOfType(StyleType.LINK, ne - 1)?.let { ne = maxOf(ne, it.end) }
-      if (ns != selStart || ne != selEnd) {
-        setSelection(ns, ne)
-        return
-      }
-    } else {
-      val link = formattingStore.rangeOfType(StyleType.LINK, selStart)
-      if (link != null && selStart > link.start && selStart < link.end) {
-        setSelection(link.end)
-        return
-      }
-    }
-
     if (!isTextChanging) {
+      // Links (e.g. mentions) are atomic: snap a partial selection to the whole link, and a caret
+      // inside a link to its end. Returning lets the recursive onSelectionChanged emit for the result.
+      formattingStore.selectionAdjustedForAtomicLinks(selStart, selEnd)?.let { (newStart, newEnd) ->
+        setSelection(newStart, newEnd)
+        return
+      }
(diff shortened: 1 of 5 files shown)


hryhoriiK97 (changes requested):
Nice one - the feature itself totally makes sense, mentions should definitely behave as atomic units.

A few things I noticed:

**Consider moving the snapping logic into `FormattingStore`** — it's really just a query on formatting ranges, and the store already owns `rangeOfType` and all the range data. Adding something like `adjustSelectionForAtomicLinks` there would keep it testable on its own and avoid growing the root view further (it's already pretty large on both platforms). The selection handler would just become a one-liner call + act on the result.

**There's a bug on iOS with double event emission.** On Android you `return` after `setSelection()`, so the recursive `onSelectionChanged` handles everything — that's correct. But on iOS there's no `return` after setting `_textView.selectedRange`, and since that triggers `textViewDidChangeSelection:` synchronously, the full flow runs in the recursive call (emit selection, state, mention, caret rect) and then the original call continues into the same emitters again. So things like `onChangeSelection` would fire twice per snap.

**Android is missing the `isTextChanging` guard.** On iOS you check `!_isTextChanging` before the snap logic, but on Android it runs unconditionally. If `onSelectionChanged` fires during a text change, the snap would call `setSelection()` and return early — skipping the event emission and `didTextChangeRecently` reset that normally happens below.

andreavrr:
Thanks for the review @hryhoriiK97, really helpful! Just pushed a commit addressing all three:

1. Moved the snapping into `FormattingStore.selectionAdjustedForAtomicLinks` (a pure range query), so it's testable on its own and the selection handler is now a one-liner on both platforms.
2. **iOS double emit:** added the `return` after setting the selection, so the synchronous recursive `textViewDidChangeSelection:` emits once.
3. **Android guard:** the snap now runs inside `!isTextChanging`, matching iOS.

hryhoriiK97 (approved):

hryhoriiK97 (commented):
@andreavrr FYI I pushed a commit with macOS support - same atomic link snapping pattern in `textInputDidChangeSelection`, with all the same guards.

andreavrr:
@hryhoriiK97 Ok great!🚀

Issue #359, opened 2026-05-28
Feature Request: Support for line-height in editor. Support for H1-H3, ordered & unordered list formatting


maxwelljmckee:
**Line-Height in Editor**
I'm using a minimal implementation that swaps the readonly markdown viewer with the markdown editor in-place without navigation to a different page or UX. I have a comfortable line-height applied to the readonly view, but when I focus the text to edit, the line-height collapses and text is compressed. It would be great if edit-mode just showed a cursor and keyboard, without other observable changes to the UX.

**Full-featured MD editing**
It seems like this library is new and you're working hard on it. We appreciate you!! Just putting in a word for more rich MD features like header sizes 1-3, bullets, numbered lists, checkbox lists, links, etc.

| Readonly Mode | Edit Mode |
| - | - |
| <img width="300" height="650" alt="Image" src="https://github.com/user-attachments/assets/c6e6bab0-7b32-4bce-82e0-7a43c2843581" /> | <img width="300" height="650" alt="Image" src="https://github.com/user-attachments/assets/774cdcaf-b409-4921-8f75-2b43c005b3b2" /> |

eszlamczyk:
Hi @maxwelljmckee thanks for the feature request! For the Headers in input this was already requested in #339. For the line-height this seems like a great addition we will bring it up with the team. As for `bullets, numbered lists, checkbox lists etc` this might be a bigger milestone that is mentioned in [our roadmap](https://github.com/software-mansion-labs/react-native-enriched-markdown#future-plans)

maxwelljmckee:
Hi Ernest, thanks for the reply! I'll keep an eye out for updates. Thanks
for all the work you do!

Cheers,
Max

hryhoriiK97:
Hey @maxwelljmckee, both points are addressed in `react-native-enriched-markdown@nightly`:

1. **Line-height** - the editor respects `lineHeight` from the `style` prop.
2. **Rich MD features** - headings (H1–H6), ordered lists, unordered lists, and links are all supported in the input now.

Checkbox lists aren't available yet - feel free to create a separate issue for that!

PR #455, opened 2026-06-27
feat(input): add H1-H3 heading block support

wildseansy:
### What/Why?

Adds **H1–H3 heading** support to the editable `EnrichedMarkdownTextInput`, addressing the "expanded markdown support" half of #359. The readonly `EnrichedMarkdownText` already renders headings, but the input only supported inline marks (bold/italic/underline/strikethrough/spoiler/link) — there was no way to author block elements.

This is the first of a small series (headings → bullet list → numbered list); it deliberately scopes to headings so each block type lands as a reviewable unit.

**How it works**

- Heading kind is stored as a per-paragraph attribute (iOS `NSAttributedString` key / Android span) that the text system migrates across edits, so it survives typing/deletion/paste without manual range bookkeeping — mirroring how inline marks behave.
- The markers (`# `, `## `, `### `) exist **only in the serialized Markdown**, never in the editor text: they're stripped on load and re-added on save. Heading vs. `#hashtag`-mention can't collide because ATX headings require a space after the `#`, and the marker never appears in the editable text the mention detector scans.
- New imperative commands `toggleH1/H2/H3`; active level for the cursor's line is reported via `onChangeState` as `h1`/`h2`/`h3` (same `{ isActive }` shape as inline styles).
- A heading is single-line: pressing Enter ends it and starts a normal paragraph.

New public API: `toggleH1()`, `toggleH2()`, `toggleH3()`, and `h1`/`h2`/`h3` on `StyleState`. Documented in `docs/INPUT.md`.

### Testing

Built and ran the example app on **iOS** and **Android**:

- Toggle H1/H2/H3 from the toolbar, type — every typed character stays in the heading (not just the first).
- Headings render at their size in-editor; `Get Raw Markdown` round-trips to `# `/`## `/`### `.
- `Set Raw Markdown` with `# / ## / ###` content parses back to rendered headings with markers stripped.
- Enter ends a heading; the next line is a normal paragraph.
- Toggling the active level off restores a paragraph; switching levels replaces.

Added a basic Maestro flow: `enrichedMarkdownInput/flows/basic/block_elements/heading_test.yaml`. (Screenshot baselines are generated by the E2E job / `--update-screenshots`.)

### PR Checklist

- [ ] E2E tests are passing 

### DIFF
--- a/.maestro/enrichedMarkdownInput/flows/basic/block_elements/heading_test.yaml
+++ b/.maestro/enrichedMarkdownInput/flows/basic/block_elements/heading_test.yaml
@@ -1,0 +1,24 @@
+appId: swmansion.enriched.markdown.example
+tags:
+  - input
+  - smoke
+  - heading
+  - block
+---
+- launchApp
+
+- runFlow:
+    file: '../../../../subflows/move_to_playground.yaml'
+
+- tapOn:
+    id: toolbar-h1
+
+- runFlow:
+    file: '../../../subflows/set_enriched_input_value.yaml'
+    env:
+      VALUE: "Heading one"
+
+- runFlow:
+    file: '../../../subflows/capture_or_assert_screenshot.yaml'
+    env:
+      SCREENSHOT_NAME: 'heading_input'
--- a/apps/react-native-example/src/components/FormattingToolbar.tsx
+++ b/apps/react-native-example/src/components/FormattingToolbar.tsx
@@ -1,7 +1,7 @@
 import { useCallback, useRef, useState } from 'react';
 import {
   Text,
-  View,
+  ScrollView,
   TouchableOpacity,
   StyleSheet,
   type StyleProp,
@@ -30,6 +30,12 @@ interface FormattingToolbarProps {
 
 const ICON_COLOR = '#001A72';
 const ICON_SIZE = 18;
+
+const HEADING_TEXT_STYLE = {
+  color: ICON_COLOR,
+  fontWeight: '700' as const,
+  fontSize: 13,
+};
 
 const ITEMS = [
   {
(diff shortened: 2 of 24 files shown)


hryhoriiK97:
Hey @wildseansy , thanks for working on this — block editing is definitely needed.

However, this can't land in its current form. The main issue is that it bypasses the pipeline architecture we have for inline styles entirely. Our inline pipeline has clear layers — `StyleType` → `FormattingStore` → `StyleHandler` → `InputFormatter` → serializer/parser — so adding a new style means plugging into existing abstractions without touching the orchestrator. This PR needs to follow the same approach for blocks.

Right now, heading logic is wired directly into `EnrichedMarkdownTextInput.mm` as one monolithic method mixing store mutation, rendering, pending state, and events. There's no `BlockHandler` protocol, no generic store, no pluggable serializer mapping. Since this is "first of a series (headings → lists)" — and #456 for lists is already up — we *will* end up refactoring all of this the moment the next block type lands. We need to build the pipeline first and have headings be the first thing plugging into it, same way bold was for inline styles.

Also — we want to support all heading levels (H1–H6), not just H1–H3. The readonly renderer already handles all six (`h1`–`h6` in `MarkdownStyleInternal`), and the input should match.

Other things that need fixing:

- **Android perf** — `normalizeHeadingSpans()` re-adds every heading span in the entire document on every keystroke. This needs to be scoped to the edited line or use inclusive span flags.
- **iOS parser** — md4c already has `enter_block`/`leave_block` callbacks (we have TODOs there). The post-pass approach (`extractHeadingBlocks` + index remap) works around the parser instead of using it. Use the existing extension point.
- **Silent fallback** — if the "line counts align" invariant breaks, the serializer silently drops headings. That's data loss. Needs an assertion at minimum.
- **Hardcoded scales** — 1.6/1.4/1.2 duplicated on both platforms with no consumer-facing props. The readonly renderer already passes explicit `fontSize` per heading level. The input must follow the same pattern so consumers can configure heading sizes consistently across read and edit views.

Please split this into: (1) block pipeline foundation with the abstractions, (2) heading as the first implementation plugging into it, (3) toolbar/example/docs. I'd recommend creating a base branch (e.g. `feat/block-editing`) and stacking separate branches off it that add things incrementally — easier to review, easier to test in isolation.

wildseansy:
Thanks for the thorough review — agreed on all points. Closing this to re-implement properly following your guidance:

1. Build the block pipeline foundation first (BlockHandler protocol, generic block store, pluggable serializer/parser mapping) mirroring the inline-style layers, so blocks plug in without touching the orchestrator.
2. Headings as the first implementation on that pipeline — H1–H6 (matching the readonly renderer), configurable per-level fontSize props, md4c enter_block/leave_block on iOS, line-scoped span normalization on Android, and an assertion instead of the silent serializer fallback.
3. Toolbar/example/docs.

I'll stack these incrementally off a feat/block-editing base branch as you suggested. Keeping this PR as a reference for the heading feature set.

PR #456, opened 2026-06-27
feat(input): add nested unordered list support

wildseansy:
### What/Why?

Adds **nested unordered (bullet) lists** to `EnrichedMarkdownTextInput` on iOS and Android, rendered inline in the editable input and serialized to Markdown (`- `, indented two spaces per depth) — closing part of #359 (block styles in the input).

Capabilities:
- `toggleUnorderedList()`, `indentList()` / `outdentList()` (and **Tab** / **Shift+Tab** on hardware keyboards) for nesting.
- **Return** continues the list — a new bullet appears immediately on the next line (on *any* line, not just the last); Return on an empty item exits the list.
- **Backspace** at the start of an item outdents, then removes the marker at depth 0 (including the very first line of the document).
- The bullet shows the instant a list is toggled/continued — before the first keystroke — correctly aligned with the caret, with no caret or layout shift when typing begins.
- Survives iOS autocorrect / paste replacing the line's text (the line's block attribute is healed).
- The placeholder is hidden while a bullet is drawn on an empty editor so it doesn't overlap the marker.
- `unorderedList` is reported via `onChangeState` (same shape as inline styles).
- Markers live only in the serialized Markdown, never in the editor text, so they don't collide with `#`/`-`-prefixed mention text.

Also adds an opt-in **`listItemSpacing`** prop (points, default `0`) that adds vertical spacing above each item so bullets read as separate rows — iOS via `paragraphSpacingBefore`, Android via a `LineHeightSpan`.

### Testing

- Example app **Playground**: toggle/indent/outdent, Return-to-continue and exit-on-empty, Backspace-to-unlist (incl. first line), nesting depth markers, autocorrect on a bullet line, `listItemSpacing` spacing, loading existing Markdown then editing mid-list.
- Maestro E2E: `enrichedMarkdownInput/flows/basic/block_elements/unordered_list_test.yaml` with an iOS baseline committed.
- Builds and runs on iOS and Android.

| iOS empty line (immediate marker) | iOS continued list |
| - | - |
|  |  |

### PR Checklist

- [ ] E2E tests are passing — iOS baseline added; Android baseline to be regenerated on CI


wildseansy:
Closing in favor of a re-implementation that follows the architecture you laid out in #455. Rather than wiring list logic into the orchestrator, lists will plug into the new block pipeline (BlockHandler + generic store + pluggable serializer/parser mapping) once the foundation and headings land.

Keeping this PR open as a reference for the list feature set we need (nested depth, indent/outdent, empty-line markers, Enter-continuation/exit, markdown serialization). Thanks for the detailed direction.

Issue #458, opened 2026-06-28
Shallow `2\pi` inline math still sits visually low


eliotgevers:
### Summary

This is separate from the missing `NSFontAttributeName` issue for iOS inline math attachments.

After adding the active font attribute to the inline math attachment string, larger formulas such as fractions and sums align much better with the surrounding line-height behavior. However, shallow zero-depth formulas such as `$2\pi$` can still look optically low compared with plain text in the same fixed-height row.

This appears to come from the math box reported by RaTeX, not from the missing font attribute. For example, `2\pi` is reported as a shallow formula with height above the baseline and no depth below the baseline. Aligning the formula baseline is technically consistent, but the visible ink can still sit lower than comparable plain text in centered UI rows.

### Reproduction

Minimal Expo/iOS reproduction:

https://github.com/eliotgevers/react-native-enriched-markdown-inline-math-repro

Run:

```sh
npm install
npm run ios
```

The repro uses `react-native-enriched-markdown@0.7.0-nightly-20260627-4e5ceb7c3` so this is tested against the current 0.7 nightly line-height behavior. In 0.7, `ios/renderer/ParagraphRenderer.m` applies `applyBaselineOffset(...)` after `applyLineHeight(...)`, and the helper in `ios/utils/ParagraphStyleUtils.m` uses `NSFontAttributeName` to compute the font line height. That matters because the first issue fixes the missing font attribute, while this issue is about the remaining shallow-formula visual offset after that fix.

The image below highlights the residual shallow-formula case after the font-attribute fix. The red line marks the row center.

<img width="520" alt="Shallow inline math formula 2 pi remains visually low compared with plain text" src="https://github.com/user-attachments/assets/0bc84114-c59a-4120-a990-5c20663a93bc" />

### Expected behavior

Shallow inline formulas should not look visibly lower than plain text when rendered in the same centered row with the same paragraph font size and line height.

### Actual behavior

`$2\pi$` remains visually low compared with plain text, even after the attachment receives the current font attribute. Taller formulas such as `\frac{a}{b}` and `\sum_{i=0}^{n} i` look much better because their math boxes have meaningful height/depth balance.

### Notes

[...]

PR #459, opened 2026-06-28
fix(ios): add font attribute to inline math attachments

### LINKED ISSUE #457: iOS inline math attachments miss the active font attribute
### Summary

Inline math on iOS is rendered as an `NSTextAttachment`. In the current `0.7` nightly, the attachment string is appended without the active `NSFontAttributeName`, while surrounding text has a font attribute from the current block style.

That matters because `0.7` now has paragraph line-height baseline-offset handling. When the math attachment character has no font attribute, it does not participate in that attribution-dependent pass the same way surrounding text does.

The relevant code path still looks like this in `0.7.0-nightly-20260627-4e5ceb7c3`:

```objc
NSAttributedString *attachmentString = [NSAttributedString attributedStringWithAttachment:attachment];
[output appendAttributedString:attachmentString];
```

### Reproduction

Minimal Expo/iOS reproduction:

https://github.com/eliotgevers/react-native-enriched-markdown-inline-math-repro

Run:

```sh
npm install
npm run ios
```

The repro uses `react-native-enriched-markdown@0.7.0-nightly-20260627-4e5ceb7c3` on purpose. Current 0.7 code applies `applyBaselineOffset(...)` after `applyLineHeight(...)` in `ios/renderer/ParagraphRenderer.m`; that helper reads `NSFontAttributeName` in `ios/utils/ParagraphStyleUtils.m` to compute the font line height and add `NSBaselineOffsetAttributeName`. This issue is about the inline math attachment missing that font attribute, not the older pre-0.7 line-height behavior.

The image below compares the current nightly with the same app after applying the small candidate patch.

iOS inline math comparison for missing font attribute on inline math attachments

### Expected behavior

Inline math attachments should carry the current font attribute so they participate in the same font/line-height baseline adjustment as surrounding text.

### Actual behavior

In the current nightly, math rows such as fractions and sums sit visibly lower than the row center when paragraph `lineHeight` is larger than the natural font line height. Adding the current font attribute to the attachment string makes those rows align with the baseline-offset logic that already handles normal text.

Note: shallow formulas such as `$2\pi$` still have a smaller residual visual offset after this patch. I believe that is a separate math-box/optical-centering issue, not the missing font attribute issue described here.

### Suggested fix

Create the attachment string as mutable and add the current font attribute before appending it:

```objc
NSMutableAttributedString *attachmentString =
    [[NSAttributedString attributedStringWithAttachment:attachment] mutableCopy];
if (currentFont) {
  [attachmentString addAttribute:NSFontAttributeName value:currentFont range:NSMakeRange(0, attachmentString.length)];
}
[output appendAttributedString:attachmentString];
```

This keeps the fix inside the default iOS renderer and avoids requiring consumers to add app-specific transforms or special inline-math styling.

### Environment

- `react-native-enriched-markdown@0.7.0-nightly-20260627-4e5ceb7c3`
- `react-native@0.85.3`
- `expo@~56.0.12`
- iOS Simulator

eliotgevers:
## Summary

Fix iOS inline math line-height alignment by preserving the current font attribute on the math attachment string.

Inline math is rendered as an `NSTextAttachment`. The attachment string did not receive `NSFontAttributeName`, even though surrounding text does. In the current `0.7` nightly, paragraph line-height baseline-offset handling depends on font attributes, so the attachment character can be skipped or handled differently from surrounding text.

This PR keeps the change narrow: it does not add a custom placement path for math attachments. It only gives the attachment string the same current font attribute already used by surrounding text.

## Changes

- Convert the inline math attachment string to a mutable attributed string.
- Add `NSFontAttributeName` to the attachment string when `currentFont` is available.
- Leave attachment bounds and block line-height propagation unchanged.

## Reproduction

Minimal reproduction:

https://github.com/eliotgevers/react-native-enriched-markdown-inline-math-repro

The repro uses `react-native-enriched-markdown@0.7.0-nightly-20260627-4e5ceb7c3` and compares the current nightly with this fix in the same iOS app.

## Screenshots

iOS inline math before and after font attribute fix

## Test Plan

- Verified the repro on iOS Simulator with `react-native-enriched-markdown@0.7.0-nightly-20260627-4e5ceb7c3`.
- Compared current nightly against the candidate patch using the same simulator, same markdown, and same crop.
- Verified that the patch improves larger inline math formulas such as fractions and sums.
- Rebuilt a real Expo iOS app using the same patch and confirmed `ENRMMathInlineRenderer.m` compiled with the change.

Note: shallow zero-depth formulas such as `$2\pi$` can still have a smaller residual optical offset after this fix. I am treating that as a separate follow-up issue because it appears to involve math box/baseline semantics rather than the missing font attribute fixed here.

Fixes #457
Related to #458


hryhoriiK97 (approved):
LGTM! Thank you for the fix! 

@eliotgevers  could you add before/after screenshots directly in the PR description? 🙏

eliotgevers:
Done, I added the before/after screenshot directly to the PR description. Thanks again for the review!

PR #460, opened 2026-06-29
feat(input): block-editing pipeline foundation

wildseansy:
## What

First PR in the block-editing series requested in #455 (closed in favor of this rebuild). This adds the **generic block-editing pipeline foundation** for the editable input, mirroring the existing inline-style pipeline (`StyleType → FormattingStore → StyleHandler → InputFormatter → serializer/parser`). No concrete block type is registered yet — with no handler the editor behaves exactly as today (plain paragraphs, all inline styles unaffected). A follow-up PR plugs the first handler (headings, H1–H6) in without touching the orchestrator.

## Abstractions (both platforms)

| Inline (existing) | Block (this PR) |
|---|---|
| `ENRMInputStyleType` / `StyleType` | `ENRMInputBlockType` / `BlockType` (Paragraph default) |
| `ENRMStyleHandler` / `StyleHandler` | `ENRMBlockHandler` / `BlockHandler` — paragraph attributes, markdown line prefix (parser recognition stays in each parser's central map, mirroring the inline pipeline) |
| `ENRMFormattingStore` / `FormattingStore` | `ENRMBlockStore` / `BlockStore` — generic, line-scoped block ranges + `adjustForEdit` |
| formatter style-handler registry | formatter block-handler registry (empty here) + `applyBlockRanges` |
| serializer inline delimiters | block handlers contribute line prefixes (no central switch) |
| parser inline span mapping | iOS: real md4c `enter_block`/`leave_block` (replaces the no-op TODO); Android: block ranges from the shared AST (C++ untouched) |

## Notes addressing the #455 review

- **No monolithic orchestrator logic** — blocks plug in via handlers, like `bold` did for inline.
- **iOS parser** uses md4c `enter_block`/`leave_block` (the existing TODO is removed), not a post-pass.
- **Android** reuses the shared `MarkdownASTNode` the readonly renderer already emits — `MD4CParser.cpp` is unchanged, so readonly parsing is unaffected.
- **No silent serializer fallback** — the block-aware serializer asserts the line-count invariant instead of silently dropping blocks.

## Testing

Both platforms build green (example app, iOS sim + Android emulator). Behavior is inert (byte-for-byte unchanged) with no block handler registered.

https://claude.ai/code/session_01EhuNE6PjRGjroxobhdd72q

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
@@ -65,77 +65,13 @@ class BlockStore {
     removeBlocksOverlapping(start, end)
   }
 
-  /**
-   * Shifts/clips block ranges to follow a text edit, using the same overlap
-   * classification shape as [FormattingStore.adjustForEdit].
-   */
+  /** Shifts/clips block ranges to follow a text edit. See [RangeEditAdjustment]. */
   fun adjustForEdit(
     editLocation: Int,
     deletedLength: Int,
     insertedLength: Int,
   ) {
-    if (deletedLength == 0 && insertedLength == 0) return
-
-    val deleteEnd = editLocation + deletedLength
-    val indexesToRemove = mutableListOf<Int>()
-
-    for ((idx, range) in ranges.withIndex()) {
-      if (deletedLength > 0) {
-        when (classifyOverlap(range.start, range.end, editLocation, deleteEnd)) {
-          EditOverlap.BEFORE_EDIT -> { /* no change */ }
-
-          EditOverlap.AFTER_EDIT -> {
-            range.start = range.start - deletedLength + insertedLength
-            range.end = range.end - deletedLength + insertedLength
-          }
-
-          EditOverlap.FULLY_DELETED -> {
-            indexesToRemove.add(idx)
-          }
-
-          EditOverlap.DELETED_INSIDE -> {
-            range.end = range.end - deletedLength + insertedLength
-          }
-
-          EditOverlap.CLIPPED_END -> {
-            val newEnd = editLocation + insertedLength
-            val newLength = if (newEnd > range.start) newEnd - range.start else 0
-            range.end = range.start + newLength
-            if (newLength == 0) indexesToRemove.add(idx)
-          }
-
-          EditOverlap.CLIPPED_START -> {
-            val charsClipped = deleteEnd - range.start
-            val newStart = editLocation + insertedLength
-            val oldLength = range.length
-            range.start = newStart
-            range.end = newStart + oldLength - charsClipped
-            if (range.length == 0) indexesToRemove.add(idx)
-          }
-        }
-      } else {
-        // Insert-only. Insertion at exactly range.start shifts the block right
-        // (typed characters stay outside it) — same convention as
-        // [FormattingStore]. A concrete block handler re-normalizes its line
-        // bounds on the edit pass, so a leading insert rejoins the block there.
-        when {
-          range.start >= editLocation -> {
-            range.start += insertedLength
-            range.end += insertedLength
-          }
-
-          editLocation < range.end -> {
-            range.end += insertedLength
-          }
-        }
-      }
-    }
-
-    for (idx in indexesToRemove.reversed()) {
-      ranges.removeAt(idx)
-    }
-
-    ranges.removeAll { it.length == 0 }
+    RangeEditAdjustment.adjustForEdit(ranges, editLocation, deletedLength, insertedLength)
   }
 
   /**
@@ -177,27 +113,5 @@ class BlockStore {
       index++
     }
     return index
-  }
-
-  private enum class EditOverlap {
-    BEFORE_EDIT,
-    AFTER_EDIT,
-    FULLY_DELETED,
-    DELETED_INSIDE,
-    CLIPPED_END,
-    CLIPPED_START,
-  }
-
-  private fun classifyOverlap(
-    rangeStart: Int,
-    rangeEnd: Int,
-    editLocation: Int,
-    deleteEnd: Int,
-  ): EditOverlap {
-    if (rangeEnd <= editLocation) return EditOverlap.BEFORE_EDIT
-    if (rangeStart >= deleteEnd) return EditOverlap.AFTER_EDIT
-    if (rangeStart >= editLocation && rangeEnd <= deleteEnd) return EditOverlap.FULLY_DELETED
-    if (rangeStart < editLocation && rangeEnd > deleteEnd) return EditOverlap.DELETED_INSIDE
-    return if (rangeStart < editLocation) EditOverlap.CLIPPED_END else EditOverlap.CLIPPED_START
   }
 }
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
@@ -1,0 +1,193 @@
+package com.swmansion.enriched.markdown.input.formatting
+
+import com.swmansion.enriched.markdown.input.model.BlockRange
+import com.swmansion.enriched.markdown.input.model.BlockType
+import java.util.Collections
+
+/**
+ * Stores the block-level (paragraph-scoped) ranges for the editor, mirroring
+ * [FormattingStore]. Unlike inline ranges, block ranges never overlap: at most
+ * one block covers any given paragraph, and ranges are kept normalized to
+ * whole-line boundaries.
+ */
+class BlockStore {
+  private val ranges = mutableListOf<BlockRange>()
+
+  val allRanges: List<BlockRange> get() = Collections.unmodifiableList(ranges)
+
+  fun setRanges(newRanges: List<BlockRange>) {
+    ranges.clear()
+    ranges.addAll(newRanges.sortedBy { it.start })
+  }
+
+  fun clearAll() {
+    ranges.clear()
+  }
+
+  fun blockRangeContaining(position: Int): BlockRange? = ranges.firstOrNull { position >= it.start && position < it.end }
+
+  /**
+   * Sets/replaces the block on every paragraph the given range touches, expanding
+   * to whole-line boundaries within [text]. Removes any block previously covering
+   * those paragraphs.
+   */
+  fun setBlock(
+    type: BlockType,
+    level: Int,
+    paragraphStart: Int,
+    paragraphEnd: Int,
+    text: CharSequence,
+  ) {
+    val (start, end) = paragraphBounds(paragraphStart, paragraphEnd, text)
+    removeBlocksOverlapping(start, end)
+    if (end <= start) return
+
+    val block = BlockRange(type, start, end, level)
+    ranges.add(sortedInsertionIndex(start), block)
+  }
+
+  /**
+   * Clears any block on the paragraphs the given range touches (reverting them to
+   * the implicit paragraph default).
+   */
+  fun removeBlock(
+    paragraphStart: Int,
+    paragraphEnd: Int,
+    text: CharSequence,
+  ) {
+    val (start, end) = paragraphBounds(paragraphStart, paragraphEnd, text)
+    removeBlocksOverlapping(start, end)
+  }
+
+  /**
+   * Shifts/clips block ranges to follow a text edit, using the same overlap
+   * classification shape as [FormattingStore.adjustForEdit].
+   */
+  fun adjustForEdit(
+    editLocation: Int,
+    deletedLength: Int,
+    insertedLength: Int,
+  ) {
+    if (deletedLength == 0 && insertedLength == 0) return
+
+    val deleteEnd = editLocation + deletedLength
+    val indexesToRemove = mutableListOf<Int>()
+
+    for ((idx, range) in ranges.withIndex()) {
+      if (deletedLength > 0) {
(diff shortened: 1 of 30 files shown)


wildseansy:
To show how this foundation is meant to be *used*, here's the first handler (headings, H1–H6) plugging into it as a reference PR, stacked on this branch so the diff is only the heading delta:

https://github.com/wildseansy/react-native-enriched-markdown/pull/1

It demonstrates the intended extension pattern — a new block type is added by implementing the `BlockHandler` protocol (paragraph styling + markdown line prefix + parser-block matching) and registering it, with no orchestrator changes. Per-level heading sizes are configured via `markdownStyle.h1..h6`, mirroring the readonly renderer. Verified on both platforms (clean editor text, per-level sizing, single-marker round-trip). Lists would follow the same pattern.

wildseansy:
@hryhoriiK97 - base pipeline ready for review.  See https://github.com/wildseansy/react-native-enriched-markdown/pull/1 for how it's used for headings, and https://github.com/wildseansy/react-native-enriched-markdown/pull/2 for the headings fully integrated into the test app

hryhoriiK97 (commented):
@wildseansy thanks for the PR! I'll try to take a look this week if possible. We have a lot on our plate right now, so I can't make any promises.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
```diff
@@ -0,0 +1,203 @@
...
+  private fun sortedInsertionIndex(location: Int): Int {
+    var index = 0
+    for (existing in ranges) {
+      if (existing.start > location) break
+      index++
+    }
+    return index
+  }
+
+  private enum class EditOverlap {
```
hryhoriiK97: `adjustForEdit`, `EditOverlap`, and `classifyOverlap` are copy-pasted from `FormattingStore`. Any future bug fix to overlap logic would need to be applied in both places. Worth extracting into a shared utility that both stores can delegate to.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/model/BlockRange.kt
```diff
@@ -0,0 +1,20 @@
...
+ * A block-level element occupying a paragraph/line range. Mirrors [FormattingRange]
+ * but for block scope: the range always covers whole lines.
+ *
+ * @property type the block kind (PARAGRAPH is the implicit default).
+ * @property start inclusive start offset, in plain-text characters.
+ * @property end exclusive end offset, in plain-text characters.
+ * @property level generic integer payload, 0 by default. Headings use it for the
+ *   H-level (1-6); list items will use it for nesting depth.
+ */
+data class BlockRange(
```
hryhoriiK97: This is a `data class` with mutable `var` fields, which means `equals()`/`hashCode()` depend on mutable state. If a `BlockRange` ever ends up in a set or as a map key, mutations will silently break lookups. Same pattern as `FormattingRange` so not new to this PR, but worth keeping in mind.

packages/react-native-enriched-markdown/ios/input/ENRMInputFormatter.mm
```diff
@@ -189,4 +206,74 @@ - (void)applyFormattingRanges:(NSArray<ENRMFormattingRange *> *)ranges
...
+                       usingBlock:^(id value, NSRange range, BOOL *stop) {
+                         if (value != nil) {
+                           [previouslyClaimedRanges addObject:[NSValue valueWithRange:range]];
+                         }
+                       }];
+  for (NSValue *rangeValue in previouslyClaimedRanges) {
+    NSRange range = rangeValue.rangeValue;
+    [textStorage removeAttribute:ENRMBlockTypeAttributeName range:range];
+    [textStorage removeAttribute:ENRMBlockLevelAttributeName range:range];
+    [textStorage removeAttribute:NSParagraphStyleAttributeName range:range];
```
hryhoriiK97, line 244: The reset pass strips `NSParagraphStyleAttributeName` from previously-claimed ranges, so the "seed from existing attributes" on re-apply will always get the default paragraph style rather than whatever was there before. In practice this is probably fine since `applyWritingDirection` runs after and handlers set absolute values - just something to be aware of.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputEventEmitter.kt
```diff
@@ -185,7 +185,16 @@ class InputEventEmitter(
 
   private fun serializeToMarkdown(): String {
     val plainText = view.text?.toString() ?: ""
-    return MarkdownSerializer.serialize(plainText, view.allFormattingRangesForSerialization())
+    // Each block resolves its markdown line prefix through its registered handler.
```
hryhoriiK97: Nit: this comment narrates what the code does — the function call below is self-explanatory. Consider removing.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
```diff
@@ -1,9 +1,71 @@
...
+    // Inline delimiters never cross a newline, so the line partition is preserved.
+    // If this ever breaks, prefixes would land on the wrong lines. Contract on
+    // both platforms: log and fall back to inline-only output — a library must
+    // not crash the host app over lost block prefixes.
+    if (plainLines.size != markdownLines.size) {
+      Log.e(TAG, "Block serialization line-count invariant violated: plain=${plainLines.size} markdown=${markdownLines.size}")
+      return inlineMarkdown
+    }
+
+    // Plain-text character offset at the start of each line.
```
hryhoriiK97: Nit: the variable name `lineStartOffsets` already says this. Consider removing.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
```diff
@@ -1,9 +1,71 @@
...
+    val lineStartOffsets = IntArray(plainLines.size)
+    var runningOffset = 0
+    for (i in plainLines.indices) {
+      lineStartOffsets[i] = runningOffset
+      runningOffset += plainLines[i].length + 1 // +1 for the '\n' separator
```
hryhoriiK97: Nit: obvious from context, consider removing.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
```diff
@@ -1,9 +1,71 @@
...
+    }
+
+    for (blockRange in blockRanges) {
+      val prefix = blockPrefixProvider(blockRange)
+      if (prefix.isEmpty()) continue
+
+      for (lineIndex in plainLines.indices) {
+        val lineStart = lineStartOffsets[lineIndex]
+        val lineEnd = lineStart + plainLines[lineIndex].length
+        // A block claims a line if their ranges intersect (block ranges are
```
hryhoriiK97: Nit: the `if` condition is clear enough on its own. Consider removing.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -863,6 +870,26 @@ - (void)showLinkPrompt
...
 
+/// Serializes plain text with inline + block formatting, resolving each block's
+/// markdown line prefix through its registered handler. With no block handlers
+/// registered the prefix provider is never consulted productively and output
+/// equals the inline-only serialization.
+- (NSString *)serializeText:(NSString *)text
+                     ranges:(NSArray<ENRMFormattingRange *> *)ranges
+                blockRanges:(NSArray<ENRMBlockRange *> *)blockRanges
+{
+  // The provider runs synchronously inside the serializer call, so capturing
```
hryhoriiK97: Nit: over-explaining a trivial local capture. Consider removing.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
wildseansy: Extracted into a shared unit on both platforms in 5019270: `RangeEditAdjustment` (Kotlin, operates on a `MutableRangeBounds` interface both range models implement) and `ENRMRangeEditAdjustment` (iOS). Both stores now delegate; the duplicated `EditOverlap`/`classifyOverlap` copies are gone.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/model/BlockRange.kt
wildseansy: Converted both `BlockRange` and `FormattingRange` to plain classes in 5019270 — bounds mutate as edits shift ranges, so identity semantics are the safe default. No call site relied on structural equality (`removeRange` receives the store's own instance).

packages/react-native-enriched-markdown/ios/input/ENRMInputFormatter.mm
wildseansy: Good catch — the comment overpromised. Reworded in 5019270: seeding only matters for newly claimed paragraphs; re-claimed ones read the default after the reset pass, which is fine since handlers set absolute values and `applyWritingDirection` re-derives direction afterwards.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputEventEmitter.kt
wildseansy: Removed in 5019270.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
wildseansy: Removed in 5019270 (also from the iOS twin).

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
wildseansy: Removed in 5019270.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
wildseansy: Removed in 5019270 (also from the iOS twin).

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Removed in 5019270.

wildseansy:
@hryhoriiK97 thanks for the review — everything is addressed, replies on each thread. Summary of the updates:

**`5019270` — your review:**
- Extracted the duplicated `adjustForEdit`/`EditOverlap`/`classifyOverlap` logic into one shared unit per platform: `RangeEditAdjustment` (Kotlin, via a small `MutableRangeBounds` interface both range models implement) and `ENRMRangeEditAdjustment` (iOS). `FormattingStore` and `BlockStore` now delegate on both platforms — net −91 lines.
- Converted `BlockRange` **and** `FormattingRange` from data classes to plain classes so equals/hashCode don't depend on mutable bounds. Verified no call site relied on structural equality.
- Fixed the misleading "seed from existing attributes" comment in `applyBlockRanges`.
- Removed all the narrating comments you flagged (plus their identical twins in the iOS serializer).

**`c3d7992` — earlier self-review pass, in case you're re-reviewing the whole diff:**
- Dropped the unused handler-matching API (`matchesNodeType` / `matchesMd4cBlockType` / `allBlockHandlers`); parser recognition lives in each parser's central map, mirroring the inline pipeline. (The extension-pattern note in my earlier comment above predates this — a new block type now adds one map entry + one handler.)
- iOS parses markdown once per import instead of twice (block ranges come from the same md4c run as inline spans).
- iOS block pass now strips its previous attributes before re-applying, so removed blocks can't leave stale paragraph styling.
- Serializer line-count invariant no longer crashes release Android — both platforms log + fall back to inline-only output.

Verified: ktlint + `compileDebugKotlin` green, iOS builds clean via xcodebuild, and the maestro smoke suite passes on functional assertions (screenshot-threshold diffs on my machine are iOS-runtime font drift vs the 26.3 baselines, unrelated to this diff).

hryhoriiK97 (approved):
LGTM! Thank you!

@wildseansy are you planning to open the heading handler PR (from wildseansy#1) as a follow-up?

wildseansy:
@hryhoriiK97 - yep.  I'll re-assign base to `software:main`.  Thanks for the review!

wildseansy:
@hryhoriiK97 - heading blocks submitted: https://github.com/software-mansion/react-native-enriched-markdown/pull/497.  Subsequent PR here that adds into the toolbar: https://github.com/wildseansy/react-native-enriched-markdown/pull/2

PR #461, opened 2026-06-29
feat: copyToClipboard ref method for EnrichedMarkdownTextInput

### LINKED ISSUE #341: Ref method for copy to clipboard (Input)
There's copy options which it requires selecting text and pressing the context menu. In my app where i render content as rich markdown using the library, i have a copy button that should copy the rich content but it doesn't work because there's no ref method for copying the content.

Expose a copyToClipboard method from ref so i can copy the content exactly as if user select all the content and press copy in the context menu.

What i have now is a copy button that copies the raw markdown or plain text (not ideal)

mvanhorn:
### What/Why?

Resolves #341. An app that renders content through `EnrichedMarkdownTextInput` wanted a copy button that copies the rich content, but the only way to copy was for the user to manually select text and use the context menu. From JavaScript the app could reach the raw Markdown or plain text, which loses the formatting the user sees on screen.

This adds a `copyToClipboard` imperative ref method that places the input's full rich content on the system clipboard, producing the same result as selecting all text and pressing the context menu's copy action. The method is a no-op when the input is empty and leaves the current selection untouched, so it is safe to call from a toolbar button at any time.

The command follows the existing `removeLink`/`setValue` pattern through the codegen command list and the `useImperativeHandle` block. On Android it writes the whole `Editable` to the `ClipboardManager`, which keeps the inline styles available to rich paste targets without moving the caret. On iOS and macOS it reuses the exact pasteboard format already produced by the user-triggered copy action, writing both plain text and the existing private Markdown pasteboard type so a later paste back into the input restores the formatting.

### Testing

`yarn typecheck` and `yarn lint` pass, and `clang-format --Werror` is clean on the changed Objective-C++ files. I also confirmed that React Native codegen regenerates the native command dispatchers so `copyToClipboard` routes to the new native methods on both platforms. Documentation in `docs/INPUT.md` and `docs/API_REFERENCE.md` is updated.

### PR Checklist

Native compile, running the example app, and the E2E suite were not performed in this environment, so a maintainer should verify the on-device behavior.

Fixes #341


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -411,6 +413,17 @@ class EnrichedMarkdownTextInputView(
     }
   }
 
+  // Copies the whole rich content to the system clipboard, matching the
+  // result of selecting all text and pressing the context menu's copy action.
+  // Writing the full Editable directly preserves inline styles for rich paste
+  // targets without disturbing the current selection.
```
hryhoriiK97: The comment says "Writing the full Editable directly preserves inline styles for rich paste targets" — but `ClipData.newPlainText(null, content)` writes plain text only, it doesn't preserve inline styles. The comment should accurately describe what the method does.

Note: on iOS, copy/paste preserves formatting via a private pasteboard type (`kENRMMarkdownPasteboardType`). Android doesn't have this round-trip yet — system copy/paste is plain text only. This is a pre-existing gap, not introduced by this PR, but worth tracking as a follow-up if rich paste on Android is desired.

hryhoriiK97 (approved):

hryhoriiK97 (changes requested):
This doesn't compile — `_textView` is typed as `ENRMPlatformTextView` (`UITextView`/`NSTextView`), not `ENRMInputTextView`, so the compiler can't find `copyEntireContents`.

Rather than downcasting, I'd suggest removing `copyEntireContents` from `ENRMInputTextView` entirely and doing the pasteboard write directly here in `EnrichedMarkdownTextInput.mm`. This matches how `requestMarkdown:` (line ~902) already serializes full content in this file. Something like:

```objc
- (void)copyToClipboard
{
  NSString *plainText = ENRMGetPlainText(_textView);
  if (plainText.length == 0) {
    return;
  }
  NSString *markdown = [ENRMMarkdownSerializer serializePlainText:plainText
                                                           ranges:[self allRangesIncludingTransient]];
#if TARGET_OS_OSX
  NSPasteboard *pasteboard = [NSPasteboard generalPasteboard];
  [pasteboard clearContents];
  NSMutableArray *types = [NSMutableArray arrayWithObject:NSPasteboardTypeString];
  if (markdown.length > 0) {
    [types addObject:kENRMMarkdownPasteboardType];
  }
  [pasteboard declareTypes:types owner:nil];
  [pasteboard setString:plainText forType:NSPasteboardTypeString];
  if (markdown.length > 0) {
    [pasteboard setString:markdown forType:kENRMMarkdownPasteboardType];
  }
#else
  UIPasteboard *pasteboard = [UIPasteboard generalPasteboard];
  NSMutableDictionary *items = [NSMutableDictionary dictionary];
  items[UTTypePlainText.identifier] = plainText;
  if (markdown.length > 0) {
    items[kENRMMarkdownPasteboardType] = markdown;
  }
  pasteboard.items = @[ items ];
#endif
}

**Comment by eszlamczyk** [comment]
Hi @mvanhorn, we want to add this before the release so Im gonna finish the work myself (most of the stuff is good, just need some example app enhancements and fixing in the compilation errors)

**Comment by eszlamczyk** [comment]
### Android

https://github.com/user-attachments/assets/863501fe-20b4-4e3d-973d-f3e15d74a806

### IOS
https://github.com/user-attachments/assets/4002c6c8-4549-4803-b19d-5123ed6d076b

File: packages/react-native-enriched-markdown/src/EnrichedMarkdownTextInput.tsx [hunk 2]
```diff
@@ -109,6 +109,19 @@ export interface EnrichedMarkdownTextInputInstance {
...
+   * result of selecting all text and pressing the context menu's copy action.
+   * The selection is left unchanged, and calling it on an empty input is a
+   * no-op.
+   *
+   * On iOS and macOS the clipboard receives both plain text and a private
+   * markdown pasteboard type, so pasting back into an
+   * `EnrichedMarkdownTextInput` restores the formatting; external apps
+   * receive plain text only. On Android the clipboard receives plain text
+   * only — inline styles are not preserved for any paste target.
+   */
```
hryhoriiK97: Nit: We can drop this comment since this info is already covered in API_REFERENCE.

hryhoriiK97 (approved):
LGTM! I've left one nit comment 👍

PR #469, opened 2026-06-29
fix: resolve modal flicker on Android after setting raw markdown

hryhoriiK97:
### What/Why?

Fixes a visual flicker on Android where the "Set Raw Markdown" modal briefly reappears after being dismissed.

The root cause was that heavy work (`setValue` + markdown preview re-render) ran in the same frame as the modal dismiss, causing Android to briefly redraw the modal overlay during the animation. Additionally, the modal was nested inside a `ScrollView`, which amplified the issue during re-renders.

Changes:
- Dismiss the modal first, then defer `setValue`/`setMarkdown` via `requestAnimationFrame`
- Move the modal outside the `ScrollView` (matching the pattern used by other modals in the app)
- Use `fade` animation on Android instead of `slide` to avoid animation overlap

#### Screenshots

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/21e2142a-a076-48ba-9f2e-1734625a5b9a" width=300 /> | <video src="https://github.com/user-attachments/assets/4916f759-595f-46b2-b518-d32bf4979099" width=300 /> |

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/apps/react-native-example/src/screens/playground/PlaygroundScreen.tsx
+++ b/apps/react-native-example/src/screens/playground/PlaygroundScreen.tsx
@@ -11,6 +11,7 @@ import {
   Image,
   Modal,
   TextInput,
+  Keyboard,
 } from 'react-native';
 import { useHeaderHeight } from '@react-navigation/elements';
 import {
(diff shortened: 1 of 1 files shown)


eszlamczyk (approved):
Good catch!

PR #476, opened 2026-07-01
fix(android): emit list item markers when parsing input

multiplechuks:
### What/Why?

- Added a walkListItems helper.
- Wired UnorderedList and OrderedList node types into walkNode.
- Reconstructs list item markers (-, 1., and [x]/[ ] for task items) when walking the AST back into markdown.

InputParser.walkNode did not handle list nodes, so markdown parsing on Android silently dropped list markers. As a result, unordered lists, ordered lists, and task lists were rendered as plain paragraphs, losing their original markdown prefixes. This change preserves the correct markdown syntax when converting the parsed AST back into plain text

### Testing
- Manual testing as shown in the gif below
- All Maestro E2E tests ran

| Before | After |
| - | - |
|  | after |

### PR Checklist

- [x] E2E tests are passing

### NB
No new tests were added. Since this change affects the rendered list output on the input component, a Maestro test may be appropriate if you'd like one. I'm happy to add it if needed

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/InputParser.kt
@@ -44,6 +44,25 @@ object InputParser {
     return ParseResult(plainText.toString(), ranges)
   }
 
+  private fun walkListItems(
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (commented):
Hi @multiplechuks! Thanks for the PR, but we want to proceed with the proper list handling introduced in #545.

PR #482, opened 2026-07-02
fix: reset cached style state in Fabric prepareForRecycle to stop stale-markdown flicker

### LINKED ISSUE #110: Cached flicker
**Describe the bug**
Sometimes I mount a new markdown component, and on the first frame, I get a flicker where it shows a previously-rendered markdown from an entirely different component.

**To Reproduce**

no clear steps for now, just rendering different markdown blocks and toggling from one screen to another. I change the `key` on the root screen.

I wanted to share to see if this is a bug with a possibly-known answer or not.

**Expected behavior**

Each markdown block should just render what I give it. It feels as though there is a cache somewhere storing previous markdown instances globally; is this possible, or am I totally off?

**Screenshots**

https://github.com/user-attachments/assets/33e4806a-1d13-4bb8-897e-5378f1b35417

iOS, `0.3.0-nightly-20260221-57a898238`

**Additional context**

mvanhorn:
### What/Why?

On the Fabric (new architecture) iOS path, recycled `EnrichedMarkdown` / `EnrichedMarkdownText` views kept cached style and md4c-flag state from their previous mount, so a recycled view briefly showed stale markdown (a flicker) before the next props update repainted it. `prepareForRecycle` now resets the cached style state back to the generated-props baseline, and resets `_md4cFlags` from that same baseline (not the library default), so a recycled view with `md4cFlags.latexMath` disabled no longer parses `$...$` as math. Fixes #110.

### Testing

Change is confined to the iOS `prepareForRecycle` reset path on the Fabric renderer. Verified by code inspection that the reset now mirrors the reset `_props` baseline so the subsequent `updateProps` diff applies the incoming flags correctly. On-device iOS/Android runs and the example app were not exercised in this environment, so the iOS/Android build-and-run checklist items are left unchecked for a maintainer with a simulator to confirm.

### PR Checklist

- [x] Updated documentation/README if applicable (none needed for this fix)
- Ran example app to verify changes: not performed in this environment (no iOS simulator); reset path verified by inspection and review
- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
@@ -949,8 +949,16 @@ static char kENRMSegmentFadeAnimatorKey;
 
   _cachedMarkdown = nil;
   _renderedMarkdown = nil;
+  _config = nil;
+  _md4cFlags = [ENRMMd4cFlags defaultFlags];
+  _maxFontSizeMultiplier = 0;
+  _allowTrailingMargin = NO;
   _streamingAnimation = NO;
   _tableStreamingMode = ENRMTableStreamingModeProgressive;
+  _renderedStyleFingerprint = 0;
+  _pendingStyleFingerprint = 0;
+  _contextMenuItemTexts = nil;
+  _contextMenuItemIcons = nil;
   _dirtyFlags = ENRMDirtyNone;
 
   [super prepareForRecycle];
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
@@ -931,7 +931,8 @@ static char kENRMSegmentFadeAnimatorKey;
 
 - (void)prepareForRecycle
 {
-  _props = std::make_shared<const EnrichedMarkdownProps>();
+  const auto resetProps = std::make_shared<const EnrichedMarkdownProps>();
+  _props = resetProps;
   [_renderCoordinator invalidate];
 
   for (RCTUIView *segment in _segmentViews) {
@@ -951,6 +952,11 @@ static char kENRMSegmentFadeAnimatorKey;
   _renderedMarkdown = nil;
   _config = nil;
   _md4cFlags = [ENRMMd4cFlags defaultFlags];
+  _md4cFlags.underline = resetProps->md4cFlags.underline;
+  _md4cFlags.superscript = resetProps->md4cFlags.superscript;
+  _md4cFlags.subscript = resetProps->md4cFlags.subscript;
+  _md4cFlags.latexMath = resetProps->md4cFlags.latexMath;
+  _md4cFlags.highlight = resetProps->md4cFlags.highlight;
   _maxFontSizeMultiplier = 0;
   _allowTrailingMargin = NO;
   _streamingAnimation = NO;
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
@@ -67,6 +67,7 @@ typedef NS_OPTIONS(NSUInteger, ENRMDirtyFlags) {
 static char kENRMSegmentFadeAnimatorKey;
 
 @interface EnrichedMarkdown () <RCTEnrichedMarkdownViewProtocol, UITextViewDelegate>
++ (ENRMMd4cFlags *)flagsFromProps:(const EnrichedMarkdownMd4cFlagsStruct &)props;
 - (void)emitLinkPress:(NSString *)url;
 - (void)emitLinkLongPress:(NSString *)url;
 - (void)emitTaskListItemPress:(NSInteger)index checked:(BOOL)checked text:(NSString *)text;
@@ -123,6 +124,17 @@ static char kENRMSegmentFadeAnimatorKey;
   return concreteComponentDescriptorProvider<EnrichedMarkdownComponentDescriptor>();
 }
 
++ (ENRMMd4cFlags *)flagsFromProps:(const EnrichedMarkdownMd4cFlagsStruct &)props
+{
+  ENRMMd4cFlags *flags = [ENRMMd4cFlags defaultFlags];
+  flags.underline = props.underline;
+  flags.superscript = props.superscript;
+  flags.subscript = props.subscript;
+  flags.latexMath = props.latexMath;
+  flags.highlight = props.highlight;
+  return flags;
+}
+
 - (instancetype)initWithFrame:(CGRect)frame
 {
   if (self = [super initWithFrame:frame]) {
@@ -131,7 +143,7 @@ static char kENRMSegmentFadeAnimatorKey;
 
     self.backgroundColor = [RCTUIColor clearColor];
     _parser = [[ENRMMarkdownParser alloc] init];
-    _md4cFlags = [ENRMMd4cFlags defaultFlags];
+    _md4cFlags = [EnrichedMarkdown flagsFromProps:defaultProps->md4cFlags];
     _segmentViews = [NSMutableArray array];
     _segmentSignatures = [NSMutableArray array];
     _dirtyFlags = ENRMDirtyNone;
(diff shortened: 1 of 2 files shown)


packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
eszlamczyk: `defaultFlags` sets `latexMath = YES`, but the C++ prop baseline is `latexMath{false}`
  (Props.h:5205). Your recycle path reconciles `_md4cFlags` back to the prop baseline via
  `_md4cFlags.latexMath = resetProps->md4cFlags.latexMath`, but `init` skips that override, so
  after init `_md4cFlags.latexMath == YES` while `_props` (what updateProps diffs against) says
  false.

  The JS wrapper always sends a concrete value (EnrichedMarkdownText.tsx:147: `latexMath ?? true`),
  so on the **FIRST** updateProps oldViewProps is the C++ default `false`:

  • user leaves default -> JS sends true  -> new(true)  != old(false) -> fires -> latexMath = true  (good)
  • user sets latexMath: false -> JS sends false -> new(false) == old(false) -> skipped -> _md4cFlags.latexMath stays YES from init -> view parses `$…$` as math anyway (bad)

  So on first mount, explicitly disabling latexMath has no effect — same failure shape as #110,
  just on the init path instead of recycle. init should apply the same prop-baseline mapping as
  the reset (ideally via a shared `flagsFromProps:` helper used by `init/updateProps/recycle`). 

Same applies to `EnrichedMarkdown.mm`

packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
```diff
@@ -639,10 +640,25 @@ - (void)prepareForRecycle
   _forceHeightUpdateOnNextRender = NO;
   _cachedMarkdown = nil;
   _renderedMarkdown = nil;
+  _config = nil;
+  _md4cFlags = [ENRMMd4cFlags defaultFlags];
+  _md4cFlags.underline = resetProps->md4cFlags.underline;
+  _md4cFlags.superscript = resetProps->md4cFlags.superscript;
+  _md4cFlags.subscript = resetProps->md4cFlags.subscript;
+  _md4cFlags.latexMath = resetProps->md4cFlags.latexMath;
+  _md4cFlags.highlight = resetProps->md4cFlags.highlight;
```
eszlamczyk: This 5-field mapping now lives in two places (updateProps, here) (and assuming issue 1 fixed in recycle as well) . That duplication is why this bug class recurs. Suggest a single helper, e.g. `+ (ENRMMd4cFlags *)flagsFromProps:(const …Md4cFlagsStruct &)`, used by all three sites so they can't drift.

packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
```diff
@@ -639,10 +640,25 @@ - (void)prepareForRecycle
...
+  _maxFontSizeMultiplier = 0;
+  _lastElementMarginBottom = 0;
+  _allowTrailingMargin = NO;
+  _renderedStyleFingerprint = 0;
+  _pendingStyleFingerprint = 0;
+  _contextMenuItemTexts = nil;
+  _contextMenuItemIcons = nil;
```
eszlamczyk, line 649: Incomplete reset — same failure mode as the fields above. These are set in updateProps only via `new != old`, so after recycle they keep the previous mount's value whenever the next mount's prop equals the default, and get baked into the synchronous render at the end of updateProps:

```
  _lineBreakStrategy        (updateProps:587, init default NSLineBreakStrategyNone)
  _writingDirectionMode     (updateProps:594, init default FirstStrong)
  _fontScaleObserver.allowFontScaling (updateProps:478 — observer persists across recycle)
  _accessibilityLabels      (updateProps:544 — labels object isn't reset, only elements/info)
```

_lineBreakStrategy and _writingDirectionMode are the strongest: they clearly drive the render and have non-zero init defaults. Suggest resetting them here.

packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
```diff
@@ -949,8 +950,21 @@ - (void)prepareForRecycle
...
+  _md4cFlags.highlight = resetProps->md4cFlags.highlight;
+  _maxFontSizeMultiplier = 0;
+  _allowTrailingMargin = NO;
   _streamingAnimation = NO;
   _tableStreamingMode = ENRMTableStreamingModeProgressive;
+  _renderedStyleFingerprint = 0;
+  _pendingStyleFingerprint = 0;
+  _contextMenuItemTexts = nil;
+  _contextMenuItemIcons = nil;
   _dirtyFlags = ENRMDirtyNone;
```
eszlamczyk, line 969: Incomplete reset —-same failure mode as the fields this PR does reset. These are applied in
updateProps only via `new != old`, so after recycle they keep the previous mount's value
whenever the next mount's prop equals the default:

```
_lineBreakStrategy     (updateProps:889, init default NSLineBreakStrategyNone)
_writingDirectionMode  (updateProps:895, init default ENRMWritingDirectionModeFirstStrong)
_fontScaleObserver.allowFontScaling (updateProps:765 — observer persists across recycle)
_accessibilityLabels   (updateProps:843, diff-gated on ENRMAccessibilityLabelsChanged)
```

_lineBreakStrategy and _writingDirectionMode are the strongest - they feed the synchronous
render and have non-zero init defaults. Suggest resetting these here.


(Element-level a11y - _accessibilityElements etc. - isn't listed because the container doesn't own it; it's per-segment anddiscarded with the segments above.)

eszlamczyk (changes requested):
Hi @mvanhorn 

Thanks for chasing this down - root cause is right. On Fabric, updateProps applies most props only when `new != old`, so any per-mount state not reset in prepareForRecycle survives into the next mount and gets baked into the synchronous render at the end of updateProps whenever the recycled view's incoming prop equals the default. The md4c-flag and style-fingerprint resets
here correctly fix that for those fields.

Two things I'd like resolved before merge (details inline):

1. Incomplete reset — the fix stops short of its own logic. Several sibling fields (_lineBreakStrategy, _writingDirectionMode, _fontScaleObserver.allowFontScaling, _accessibilityLabels) have the identical failure mode and sit two lines from what's already being reset. At minimum I'd reset the render-feeding ones (_lineBreakStrategy, _writingDirectionMode) in both components, or note explicitly why they're out of scope — otherwise #110's class can still reproduce.

2. init vs recycle disagree on the md4c baseline - a live first-mount bug. `defaultFlags` seeds latexMath=YES while the C++ prop baseline is false (Props.h:5205). The recycle path reconciles `_md4cFlags` back to that baseline; init doesn't. Since the JS wrapper always sends a concrete value (EnrichedMarkdownText.tsx:147: `latexMath ?? true`), setting `latexMath: false` on first mount gives new(false) == old(false) → diff skipped → `_md4cFlags` stays YES from init → `$…$` still renders as math. So disabling latexMath on first mount silently doesn't work. Fix: have init seed from the prop baseline like recycle does (a shared `flagsFromProps:` helper for init + recycle keeps them from drifting).

mvanhorn:
Went through all three in 31393bb. Extracted a +flagsFromProps: helper (per component, on the concrete struct type) and routed init, updateProps, and the recycle reset through it so the five-field mapping cannot drift between sites. init now derives from the prop baseline instead of defaultFlags, which fixes the first-mount case you flagged: latexMath: false on a fresh mount is honored now instead of being skipped by the new != old diff and staying YES from init. I also reset _lineBreakStrategy and _writingDirectionMode to their init defaults (NSLineBreakStrategyNone / FirstStrong) in the recycle path of both components, since they feed the synchronous render and were keeping the previous mount's value. I left the font-scale observer and the accessibility-labels object out for now, matching your read that they are weaker. clang-format, typecheck, lint, and the library build are clean; I could not run the native iOS build or E2E locally (no CocoaPods or simulator), so those lean on CI.

eszlamczyk (approved):
@mvanhorn Thanks for your contribution, gonna clean up some minor stuff and merge it today

mvanhorn:
Thanks for cleaning it up and landing it, @eszlamczyk - stale marks in recycled Fabric views were a nasty one to chase.

PR #488, opened 2026-07-02
fix(expo): build config plugin separately from main lib

### LINKED ISSUE #487: [0.7.x] Expo config plugin fails with `Cannot require() ES Module` on Node 22+
**Describe the bug**

Since v0.7.0, using `react-native-enriched-markdown` as an Expo config plugin crashes during `expo prebuild` with:

```
PluginError: Unexpected module status 0. Cannot require() ES Module
/node_modules/react-native-enriched-markdown/lib/module/plugin/withReactNativeEnrichedMarkdown.js
because it is not yet fully loaded.
(From react-native-enriched-markdown/app.plugin.js in non-loader-hook thread)
```

`app.plugin.js` uses `require()` to load the compiled plugin, but the only build output is `lib/module/` which is an ES Module (`"esm": true` in bob config). Expo resolves config plugins synchronously via CommonJS, which fails on Node 22+.

**To Reproduce**

1. Install `react-native-enriched-markdown` >= 0.7.0 in an Expo project
2. Add the plugin to `app.json`:
   ```json
   {
     "plugins": [["react-native-enriched-markdown", { "enableMath": false }]]
   }
   ```
3. Run `npx expo prebuild --no-install`
4. See the `PluginError: Unexpected module status 0` error

**Expected behavior**

`expo prebuild` completes successfully and the plugin applies the math configuration to the native project.

- Node.js: 22+
- Expo SDK: 52 / 53
- `react-native-enriched-markdown`: 0.7.0

**Additional context**

Root cause: `app.plugin.js` in v0.7.0 does:
```js
module.exports = require('./lib/module/plugin/withReactNativeEnrichedMarkdown');
```
`lib/module/` is an ESM build — `require()` cannot load it in Node 22+.

Fix: compile `src/plugin/` separately to CommonJS via a dedicated `plugin/tsconfig.build.json` and point `app.plugin.js` at `./plugin/build/withReactNativeEnrichedMarkdown`. This is the pattern used by Expo's own packages.

huextrat:
### What/Why?

fixes: https://github.com/software-mansion/react-native-enriched-markdown/issues/487

- Add dedicated tsconfig for compiling the Expo config plugin
- Update plugin entry point to use compiled output
- Include plugin build dir in package files and clean script
- Ignore generated plugin build artifacts in git
- 
### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (approved):
@huextrat thank you for the fix! I will make a new release with it today.

PR #497, opened 2026-07-03
feat(input): heading blocks (H1-H6) on the block pipeline

wildseansy:
## What

Second PR in the block-editing series: the first concrete block type — headings H1–H6 — plugging into the pipeline foundation merged in #460. Follows the extension pattern exactly: one entry in each parser's central block map + one registered handler, no orchestrator changes.

## How

**Both platforms**
- `HeadingBlockHandler` / `ENRMHeadingBlockHandler` — owns heading paragraph styling (per-level font size/weight/color) and the markdown line prefix (`"# "`…`"###### "`). One handler instance serves all six levels, registered under each `HEADING_n` key.
- `BlockType.HEADING_1..6` / `ENRMInputBlockTypeHeading1..6`; parsers map heading nodes centrally — Android `nodeTypeToBlockType` reads the AST node's `level` attribute, iOS `kSupportedBlocks` + `resolveBlockLevel` reads `MD_BLOCK_H_DETAIL.level`.
- Per-level `h1`–`h6` style props on the input, mirroring the readonly renderer's `markdownStyle` so heading sizing is configured consistently across read and edit views (normalized with complete defaults on the JS side).
- `BlockStore` gains `normalizeToLineBounds` (idempotent): after an edit is adjusted, ranges re-snap to whole-line bounds — re-absorbs characters typed at a block's edges, clips a newline-split heading to its first line, and drops blocks orphaned by line joins.

## Testing

- Example app builds green on both platforms (iOS sim via xcodebuild, Android `compileDebugKotlin`), ktlint/clang-format clean.
- Manual: type `# ` prefixes, import markdown with H1–H6, edit at heading boundaries, serialize round-trip.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #498, opened 2026-07-04
feat(input): import block ranges on paste

wildseansy:
## What

Follow-up to [the paste gap flagged in #497 review](https://github.com/software-mansion/react-native-enriched-markdown/pull/497#pullrequestreview-4629282365): pasting markdown containing headings lost the heading, because the paste path only round-tripped through `InputParser` → `FormattingStore`, never `BlockStore`.

> ~~Stacked on #497~~ — #497 merged; rebased onto `main`, single commit.

## How

**Both platforms** — paste now imports parsed block ranges into `BlockStore` alongside inline ranges, and handles mixed content (multiple headings of different levels + paragraphs + inline styles) in a single paste. Each pasted block is re-anchored at the insertion offset and inserted via the store's paragraph-scoped `setBlock`/`setBlockType:forParagraphRange:`, which snaps it to line bounds and evicts whatever block the paste landed on — so pasting a heading mid-paragraph promotes that line, and pasting a paragraph into a heading joins it, consistently with the editing pipeline's line-bound invariants.

**iOS**
- `replaceTextInRange:withText:formattingRanges:` gains a `blockRanges:` designated variant (existing callers keep the inline-only signature); `pasteMarkdown` passes `parsed.blockRanges` through it.
- `copyToClipboard` now serializes with block prefixes via the shared `serializeText:ranges:blockRanges:` helper — it was the one remaining inline-only serialization, so a copy → paste round trip through the JS command dropped `#` markers.

**Android**
- There was no markdown paste path at all — copy-as-markdown produced a plain-text clip that pasted back literally. New `MarkdownClipboard` tags our clips with a vendor MIME type (the analog of iOS's `kENRMMarkdownPasteboardType`), `onTextContextMenuItem` intercepts paste for those clips only, and the new `pasteMarkdown` mirrors the iOS flow (replace selection → adjust + normalize stores → add shifted inline ranges → `setBlock` each pasted block → apply + detector pass). External plain-text clips keep default paste handling, matching iOS.
- `markdownForSelectedRange` now clips and serializes block ranges through the handler's `markdownLinePrefix`, matching the iOS serializer parity.

## Testing

- New e2e flow `enrichedMarkdownInput/flows/basic/block_elements/paste_markdown_test.yaml` — seeds H1 + bold paragraph + H2, copies the input as a tagged markdown clip, pastes it back, screenshot-asserts the editor (pasted H1 promotes the merged caret line, H2 lands intact, bold preserved). iOS golden captured on the CI device model; `ios-only` until an Android golden is captured on the CI AVD.

- Android: `ktlintCheck` + `compileDebugKotlin` green; iOS: example app builds via xcodebuild (iPhone 17 Pro sim), clang-format clean.
- Manual flow: select heading + body + subheading → copy as markdown → paste elsewhere in the input → all block levels and inline styles restored; paste from external apps unchanged.


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/MarkdownClipboard.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.content.Context
+
+/**
+ * Clipboard round-trip for markdown content, mirroring iOS's
+ * kENRMMarkdownPasteboardType: clips created here carry a vendor MIME type so
+ * paste can distinguish our markdown from arbitrary external text (which keeps
+ * pasting as plain text).
+ */
+object MarkdownClipboard {
+  const val MIME_TYPE = "text/vnd.swmansion.enriched-markdown"
```
hryhoriiK97: iOS uses `com.swmansion.enriched-markdown.markdown` (reverse-domain), so `text/vnd.com.swmansion.enriched-markdown` would be more consistent with the iOS naming. Could we align?

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -455,6 +455,55 @@ class EnrichedMarkdownTextInputView(
     clipboard.setPrimaryClip(ClipData.newPlainText(null, content))
   }
 
+  // Intercepts paste when the clip was produced by copyAsMarkdown, so formatting
+  // and block ranges survive the round trip. External clips keep default handling.
+  override fun onTextContextMenuItem(id: Int): Boolean {
```
hryhoriiK97, line 470: iOS overrides `copy:` to always store markdown alongside plain text, so regular copy→paste preserves headings transparently. Android only stores the vendor MIME via the explicit "Copy as Markdown" context menu — regular system copy loses headings. This should be aligned in this PR so the paste fix works end-to-end — the building blocks are already here (`MarkdownClipboard.newMarkdownClip`, `markdownForSelectedRange`), just needs the copy/cut intercept wired up.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/MarkdownClipboard.kt
wildseansy: good catch

wildseansy: Done in e458d2e — renamed to `text/vnd.com.swmansion.enriched-markdown`.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
wildseansy: Done in e458d2e. `onTextContextMenuItem` now also intercepts `android.R.id.copy`/`cut`: default handling runs first, then the clip is re-tagged with the selection serialized to markdown, so a regular system copy → paste round trip preserves headings. `markdownForSelectedRange` moved from `InputContextMenu` to the view for reuse (and parity with iOS, where it lives on the input). Also aligned the `copyToClipboard` command the same way — it was the remaining plain-text-only copy path on Android while iOS stores markdown under its pasteboard type.

Also added an e2e flow in bdc29b8 (`enrichedMarkdownInput/flows/basic/block_elements/paste_markdown_test.yaml`): seeds H1 + bold + H2, copies via the command, pastes back, screenshot-asserts the editor. Golden captured on the CI device model for iOS; tagged `ios-only` until an Android golden is captured on the Pixel9 CI AVD (`yarn test:e2e:android:update-screenshots`) — Android behavior verified manually on an emulator (paste restores `#`/`##`/bold).

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/MarkdownClipboard.kt
```diff
@@ -0,0 +1,35 @@
...
+object MarkdownClipboard {
+  const val MIME_TYPE = "text/vnd.com.swmansion.enriched-markdown"
+
+  fun newMarkdownClip(markdown: String): ClipData =
+    ClipData(
+      ClipDescription("Markdown", arrayOf(MIME_TYPE, ClipDescription.MIMETYPE_TEXT_PLAIN)),
+      ClipData.Item(markdown),
+    )
+
+  /** Returns the clipboard's markdown, or null if the clip wasn't created by [newMarkdownClip]. */
```
hryhoriiK97, line 33: `newMarkdownClip` stores the markdown-syntax string as the `ClipData.Item` text, but Android's `ClipData.Item` has a single `text` field shared across all declared MIME types. Unlike iOS's `UIPasteboard` where `kUTIPlainText` and `kENRMMarkdownPasteboardType` each map to their own value, here both `text/plain` and the vendor MIME type resolve to the same markdown string.

So when an external app pastes after copying from our input, it gets raw syntax (e.g. `"**hello** world"`, `"# Hello"`) instead of clean plain text. The `super.onTextContextMenuItem(copy)` call in `EnrichedMarkdownTextInputView` puts clean text on the clipboard first, but `setPrimaryClip(newMarkdownClip(...))` immediately replaces it.

iOS doesn't have this problem because `copyItemsToPasteboard` sets `plainText` and `markdown` as separate pasteboard items keyed by different UTIs.

wildseansy: Done in 1591162. Markdown now travels in the `ClipDescription` extras (`PersistableBundle`) while the item text stays the clean plain-text selection — external paste targets get `hello world`, our paste reads the markdown from the extras and restores formatting + blocks. Closest Android analog to iOS's per-UTI pasteboard items, and avoids both the shared-text problem and multi-item concatenation quirks. `copyAsMarkdown` intentionally keeps markdown as the item text — there the syntax is what the user asked to copy. Re-verified the paste round trip on an emulator after the change (headings + bold restored).

hryhoriiK97 (approved):

PR #499, opened 2026-07-04
feat(input): heading toolbar support — toggleHeading command, active-state, empty-line persistence

wildseansy:
Third piece of the heading/block series, building on #460 and #497.

## What

- **`toggleHeading(level)` command** (JS → both platforms), routed through a generic `toggleBlockType` that mirrors the inline `toggleInlineStyle` path — future block toggles (lists, blockquotes) reuse it. Toggling the active level reverts the paragraph; a multi-paragraph selection sets one block per paragraph (blocks are single-paragraph by the store's line normalization).
- **`onChangeState` gains `heading.level`** (0 = plain paragraph, 1–6 = H1–H6) for toolbar active-state, emitted only when the value changes — same dedupe as the inline flags.
- **Empty-line heading persistence**: deleting a heading's text keeps the line a heading via a zero-length anchor in the BlockStore; typing attributes stay heading-sized so the caret and next characters render correctly. A line merge (Backspace at line start) prunes headings no longer anchored at a line start — the prune runs before line normalization so a merged range can't grow over the line it merged into.
- **Shared heading defaults** (`headingDefaults.ts`): size/weight/color defaults now come from one module used by both the read-only renderer and the input, replacing the input's duplicated table and the iOS hardcoded scale fallback.
- Example app: H1–H6 toolbar buttons (horizontal scroller); INPUT.md/API_REFERENCE.md updated.

## Note on renderer defaults

Sharing the defaults surfaces one deliberate change: default heading `fontWeight` moves from `''` (inherit) to `'bold'` in the **read-only renderer** as well, so read and edit views match. If you'd rather keep the renderer's current default untouched, happy to flip the shared default to `''` instead — one-line change in `headingDefaults.ts`.

## Verification

- Android: ktlint + `compileDebugKotlin` green.
- iOS: example app `xcodebuild` BUILD SUCCEEDED; clang-format clean.
- TypeScript green.
- Simulator (iPhone 17): typed text → H1 toggle renders + button active-state; switch H1→H2; toggle-off reverts to paragraph; Enter after a heading yields a plain body line; `getMarkdown()` returns `# Hello heading\nbody line`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #509, opened 2026-07-05
feat(input): nested unordered (bullet) lists on the block pipeline

wildseansy:
Fourth piece of the block series, building on #460 / #497 / #499.

## What

- **`UNORDERED_LIST_ITEM` block type** on both platforms. The 0-based nesting depth rides in the block range's generic `level` payload — one handler instance serves every depth, mirroring how the heading handler serves all six levels. The parser derives depth from AST `UL` nesting (never from counting leading spaces); the serializer emits `- ` indented two spaces per depth.
- **Commands**: `toggleUnorderedList()`, `indentList()`, `outdentList()` (clamped to depth 5). Hardware/software **Tab** indents, **Backspace at an item's start** outdents and removes the bullet at depth 0.
- **Enter behavior is handler-driven**: a new `continuesOnNewline` on the block handler (default NO/false) makes a non-empty item continue as a sibling at the same depth, while Enter on an empty item exits the list. Headings keep reporting false — no per-type branching in the orchestrators.
- **Empty-line persistence generalizes the heading anchor machinery** from #499/#505: `BlockType.ANCHORED` (Android) and `ENRMBlockTypePersistsWhenEmpty` (iOS) replace the heading-only checks in the block store's adjust/normalize and the views' prune pass, so an emptied bullet line keeps its marker exactly like an emptied heading keeps its size.
- **Rendering**: Android draws the marker with a `LeadingMarginSpan` (`InputBulletSpan`; dot/ring/square by depth); iOS reserves the indent column via the paragraph style and draws the glyph in the layout manager, with an explicit empty-line marker for anchor-only lines.
- **State**: `onChangeState` and the context-menu `styleState` gain `unorderedList: { isActive, depth }`, matching the `{ isActive }` shape of the other entries.
- **`listItemSpacing` prop** adds vertical spacing above items (iOS `paragraphSpacingBefore`, Android `LineHeightSpan` on the item's first character).
- Example toolbar gains bullet / indent / outdent buttons; INPUT.md / API_REFERENCE.md updated; maestro flow + baseline added.

## Paste

External plain-text clips now route through `pasteMarkdown` on both platforms, so pasted markdown syntax (bullets, headings, inline styles) formats on insert instead of landing literal — this also closes the paste gap flagged in #497 (block ranges now import on paste). Syntax-free text pastes unchanged; Android's "Paste as plain text" menu item keeps the literal default as an escape hatch. Internal copy/cut/paste keeps using the tagged clipboard round-trip.

## Design note: the Android ZWSP anchor

An empty bullet line on Android has nothing to hang the marker on: a `LeadingMarginSpan` applies neither its indent nor its `drawLeadingMargin` to an empty paragraph, so a just-toggled or just-continued empty item would render flush-left with no bullet. The view therefore anchors the caret's empty list line with a single zero-width space, and strips it the moment the line gains real content or stops being a list item.

Containment, since an invisible character in user text deserves scrutiny:

- The anchor is settled **before** block spans are stamped, so exactly one marker lands per line (stamping around an unsettled anchor produced a doubled bullet during on-device testing).
- The sync is re-entrancy-guarded — its insert/delete + `setSelection` loop back through the text/selection callbacks.
- It never escapes the editor: the serializer strips ZWSP, so markdown round-trips clean, and empty-item detection (Enter-to-exit) counts non-ZWSP characters.
- iOS needs no counterpart: the paragraph style carries the indent and the layout manager draws the empty-line glyph explicitly.

If you'd rather avoid the sentinel character entirely, the alternative we saw was a custom cursor/margin drawable for empty list paragraphs — considerably more machinery for the same visual; happy to explore it if you prefer.

## Depth invariant

Manual testing surfaced two round-trip holes, fixed in 9ed2037:

- The editor could hold depth ancestry markdown cannot represent (a leading indented item, or an item nested more than one level past its predecessor) — md4c flattens those on re-parse. The stores now clamp each item to `previous adjacent item's depth + 1` (0 with no item above) during line normalization, so every editor state serializes losslessly. Indenting a list's first item is a visual no-op, matching Notion/Docs.
- An empty item serialized as a marker-only line (`  - `), which re-parses as a **setext-heading underline** for the previous line — the prior item rendered bold downstream. Empty list lines now serialize bare; empty `# ` headings keep their prefix (valid ATX).

## Verification

- Android: ktlint + `compileDebugKotlin` green. TypeScript green.
- iOS: example app `xcodebuild` BUILD SUCCEEDED; clang-format clean.
- Simulator (iPhone 17): toolbar toggle renders the bullet + active state; Enter continues the list; indent nests with the ring glyph; double-Enter on an empty item exits to a flush-left paragraph; markdown import of a nested list round-trips; `getMarkdown()` returns `- First item\n  - Second item\n\nout of list`.
- The maestro flow now scrolls the toolbar before tapping the list buttons — they sit past H1–H6 and were offscreen on phone-width viewports.


hryhoriiK97 (commented):
@wildseansy thanks for working on this - really appreciate the effort on these block-editing features! ❤️ 

I took a look at the PR and my main concern is the size - +1,693 lines across 36 files is a lot to review in one go, especially around the keyboard behavior and the ZWSP anchor logic on Android. I'd feel much more confident reviewing this in smaller pieces.

Also - are you planning to handle ordered lists as well? Most of this infrastructure (anchor machinery, indent/outdent, Enter continuation, parser/serializer) would be shared, so splitting now would keep the ordered-list follow-up small.

How about stacking it like the heading series (#460 → #497 → #499):

1. **Anchor machinery refactor** ← targets `main` — swap the heading-only checks for the generalized `ENRMBlockTypePersistsWhenEmpty` / `BlockType.ANCHORED`. Small, no new features.
2. **List rendering** ← targets PR1 — the block type, handler, bullet spans, layout manager drawing, `listItemSpacing`.
3. **Parser + serializer** ← targets PR2 — depth from AST, `- ` prefix emission. Easy to verify with round-trip tests.
4. **Commands + keyboard** ← targets PR3 — toggle/indent/outdent, Enter/Backspace/Tab, the ZWSP sync. This is the trickiest part and deserves focused attention.
5. **State + toolbar + docs + e2e** ← targets PR4 — wire up `onChangeState`, example app buttons, docs, maestro flow.

Ideally open all 5 PRs at once so I can checkout the last branch to test the full feature end-to-end, while each PR's diff on GitHub stays small and focused for review. We'd merge bottom-up as before.

In the meantime, over the next few days I'm going to focus on testing the heading feature that just landed - catching potential bugs and doing any refactoring needed there. Once that's solid, I'll be ready to move on to reviewing the list PRs.

What do you think?

wildseansy:
@hryhoriiK97 - sounds good, I can break it up.  I was originally going to do ordered lists as a follow up, but will consider in this set of changes

wildseansy:
Split as requested — and yes on ordered lists: the shared machinery made them cheap, so they're **included in the stack** rather than a follow-up.

1. #510 — anchor machinery refactor (no behavior change)
2. #511 — list block types + marker rendering (bullets **and** numbers, `listItemSpacing`)
3. #512 — parser + serializer (AST-derived depth, computed ordinals, uniform 3-space nesting indent)
4. #513 — commands + keyboard + the ZWSP/empty-anchor logic (the part that deserves the closest look)
5. #514 — state + toolbar + docs + e2e; checkout `feat/list-5-state` to test everything end-to-end

GitHub can't base a fork PR on another fork branch, so each PR targets `main` and shows a cumulative diff — every stage is exactly **one commit** on the previous branch, so each PR's own change is its last commit. Merging bottom-up collapses the next PR's diff to just its own commit, same as the heading series.

Closing this PR in favor of the stack. No rush on review — sounds good to let the heading feature settle first.

PR #510, opened 2026-07-05
refactor(input): generalize empty-line anchor machinery beyond headings (lists 1/5)

wildseansy:
**The list stack** (per [the review ask](https://github.com/software-mansion/react-native-enriched-markdown/pull/509#pullrequestreview-4631629409), now including **ordered lists** — the shared machinery made them cheap to add):

1. #510 — anchor machinery refactor (no behavior change)
2. #511 — list block types + marker rendering
3. #512 — parser + serializer
4. #513 — commands + keyboard + ZWSP/empty anchors
5. #514 — state + toolbar + docs + e2e

GitHub can't base a fork PR on another fork branch, so every PR targets `main` and the diff shown for PR *N* is cumulative — **this PR's own change is its last commit** (each stage is exactly one commit on the previous branch). We'll merge bottom-up; after each merge the next PR's diff collapses to just its own commit. Checkout `feat/list-5-state` to test the full feature end-to-end.

## This PR

Introduces `BlockType.ANCHORED` (Android) and `ENRMBlockTypePersistsWhenEmpty` (iOS) and routes the block store's adjust/normalize passes, the formatter's zero-length anchor stamping, and the views' orphan-anchor prune through them instead of heading-only checks. Pure refactor — headings are the only anchored type until #511 adds the list types.

Verified: ktlint + compileDebugKotlin green, TS green, example iOS build green (whole stack).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #512, opened 2026-07-05
feat(input): parse and serialize lists (lists 3/5)

wildseansy:
**The list stack** (per [the review ask](https://github.com/software-mansion/react-native-enriched-markdown/pull/509#pullrequestreview-4631629409), now including **ordered lists** — the shared machinery made them cheap to add):

1. #510 — anchor machinery refactor (no behavior change)
2. #511 — list block types + marker rendering
3. #512 — parser + serializer
4. #513 — commands + keyboard + ZWSP/empty anchors
5. #514 — state + toolbar + docs + e2e

GitHub can't base a fork PR on another fork branch, so every PR targets `main` and the diff shown for PR *N* is cumulative — **this PR's own change is its last commit** (each stage is exactly one commit on the previous branch). We'll merge bottom-up; after each merge the next PR's diff collapses to just its own commit. Checkout `feat/list-5-state` to test the full feature end-to-end.

## This PR (last commit)

Parsers map list items to their block type with depth derived from AST UL/OL nesting (never from counting leading spaces); iOS tags `MD_BLOCK_LI` itself since md4c emits inner paragraphs only for *loose* lists, and clips each item's range to its own first line so nested items keep their own depth.

Serializers emit `- ` / computed `N. ` via the handler's line prefix, indented a uniform three spaces per depth so ordered and bullet nesting round-trip in any combination. An empty item's line serializes bare: a marker-only line (`  - `) re-parses as a **setext-heading underline** for the previous line. Ordered markers always use computed ordinals, so a pasted `5.` start value normalizes to `1.` — noted in the docs.

Easy to verify with round-trips: `1. alpha\n   1. beta\n2. gamma` ↔ editor.

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
@@ -198,11 +198,18 @@ class EnrichedMarkdownTextInputManager :
     if (view == null || value == null) return
 
     val style = MarkdownStyleParser.parse(value)
-    view.setAutoLinkStyle(style)
-    val changed = view.formatter.updateStyle(style)
+    val changed = view.setMarkdownStyleFromProps(style)
     if (changed) {
       view.applyFormatting()
     }
+  }
+
+  @ReactProp(name = "listItemSpacing", defaultInt = 0)
+  override fun setListItemSpacing(
+    view: EnrichedMarkdownTextInputView?,
+    value: Int,
+  ) {
+    view?.setListItemSpacingFromProps(value.toFloat())
   }
 
   @ReactProp(name = "color", customType = "Color")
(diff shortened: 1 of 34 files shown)


wildseansy:
**Note on ordered-list start numbers (future extension point):**

Both parsers in this PR normalize ordered lists to start at 1 — the store recomputes each item's ordinal from its adjacent same-depth siblings, so an imported `5. first item` serializes back as `1. first item`. This matches the existing renderer behavior (`BlockStyleContext` counts from 1 and the AST doesn't carry md4c's `MD_BLOCK_OL_DETAIL.start`), so the editor and `EnrichedMarkdownText` agree with each other, but both diverge from CommonMark, which honors the start number.

Flagging it here because of react-native-streamdown: LLM output routinely contains continuation lists (`6. ...` mid-response or lists split across chunks), and today those render as `1.`. If start-number support lands library-wide someday, this PR's machinery is the natural seam — plumb `MD_BLOCK_OL_DETAIL.start` through the parsers and seed the store's per-depth ordinal counters with it instead of 1. No change proposed for this stack; just documenting the extension point.

PR #535, opened 2026-07-10
fix: autolink detection while working with ref

### LINKED ISSUE #516: Bug Report - Setting value through ref causes extra delimiters to appear in the text input for ios
🐛 Bug report — [iOS] Combined bold+italic (***) wrapping a strikethrough leaves a literal ***

Bug summary
On iOS, when an EnrichedMarkdownTextInput is given markdown where a combined bold+italic run (***) directly wraps a strikethrough (~~…~~), the closing *** is not consumed — it shows up as literal text in the input. The word itself is still styled, but a stray *** is left after it.

This is the exact markdown the input serializes for a bold+italic+strikethrough selection, so its own output does not survive a round-trip (serialize → set back via defaultValue/setValue).

Input markdown:

***~~tes~~***
Expected: the word tes rendered as bold + italic + strikethrough.

Actual (iOS): tes is styled, but a literal *** appears after it (visible content becomes tes***).

Image

Android: renders correctly — no stray ***.

The trigger is specific to *** immediately adjacent to ~~. In isolation, **bold**, *italic*, ~~strike~~, and even ***bolditalic*** (without an adjacent strikethrough) all work fine.

Library version
0.7.3

Environment info
Expo (bare workflow) project, so react-native info isn't available — output of npx expo-env-info:

  expo-env-info 2.1.0 environment info:
    System:
      OS: macOS 26.4.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.19.4
      Yarn: 4.12.0
      npm: 10.8.2
      Watchman: 2026.03.30.00
    Managers:
      CocoaPods: 1.16.2
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.4, iOS 26.4, macOS 26.4, tvOS 26.4, visionOS 26.4, watchOS 26.4
    IDEs:
      Xcode: 26.4/17E192
    npmPackages:
      expo: ^55.0.0 => 55.0.26
      react: 19.2.0 => 19.2.0
      react-native: 0.83.6 => 0.83.6
    Expo Workflow: bare
Reproduced on the iOS 26.4 simulator (iPhone 17). New Architecture enabled.

Steps to reproduce

- Render an EnrichedMarkdownTextInput seeded with the combined markdown:
- import { EnrichedMarkdownTextInput } from "react-native-enriched-markdown";
- export default function App() {return <EnrichedMarkdownTextInput defaultValue="***~~tes~~***" />;}
(The same happens via the imperative API: ref.current?.setValue("***~~tes~~***").)
- Run on iOS and look at the input.
- Result: the word is styled (bold + italic + strikethrough) followed by a literal ***; the input's content is tes*** instead of tes. On Android the same input renders correctly.
- The same is observable Any time more than 1 style is applied to the input
- e.g Bold + italic will return test*
- Bold + strikethrough will return test**
- strikethrough + italic will return test*
- Similarly an issue can be observed with links, as they seem to get double wrapped, so:
```[Testing](http://example.com) turns into [Testing](https://[Testing](http://example.com))```

Reproducible example repository
Reproduces with the single component above in a fresh app — no repository needed:

<EnrichedMarkdownTextInput defaultValue="***~~tes~~***" />

### PR DESCRIPTION
**Pull request by eszlamczyk**
### What/Why?

Fixes #516 (link corruption on iOS).

When the input's plain text contained link-like tokens, the iOS auto-link detector (`ENRMAutoLinkDetector`) corrupted them on serialize. Typing a markdown link like `[Testing](http://example.com)` produced this from `getMarkdown()`:

```
[[Testing](http://example.com)](https://[Testing](http://example.com))
```

Two defects, both in `ENRMAutoLinkDetector.mm`:

1. **Over-capture of the URL** - `tryMatchWord:` only *tested* for a regex match, then returned `normalizeUrl:` of the **entire** whitespace-delimited token. For the token `[Testing](http://example.com)` that yields `https://[Testing](http://example.com)` - the exact garbage destination from the issue.
2. **Whole-token attribution** - the auto-link attribute was applied over the full token range, so `allRangesIncludingTransient` serialized an auto-link *wrapping* the markdown syntax that was already there, giving the double-wrap above.

The existing manual-link guard didn't catch this because it only checked the token's *first* character against the formatting store.

Fix:

- `tryMatchWord:matchedRange:` now uses `firstMatchInString:` and returns **only the matched URL sub-range** (custom regexes keep their whole-token contract).
- `detectNewLinkAtRange:…matchedRange:` applies the link to that sub-range only, checks the manual-link store across the **whole** slice, and **skips tokens carrying markdown link syntax** (`[…` / `…](…`) so the parser owns them.
- `processWord:` emits the actual linked slice to `onLinkDetected` instead of the whole token.

This is iOS-only; Android was already correct.

### Testing

Using the example app the cases in table below were tested:

| Typed/set via ref input | Before (iOS) | After (iOS) |
| - | - | - |
| `[Testing](http://example.com)` | `[[Testing](http://example.com)](https://[Testing](http://example.com))` | `[Testing](http://example.com)` |
| `see:http://example.com done` | over-captured whole token | `see:[http://example.com](http://example.com) done` |
| `http://example.com` | correct | correct (still auto-links) |

*On android every case worked before the fix*

Added an E2E regression flow: `.maestro/enrichedMarkdownInput/flows/basic/inline_elements/link_autolink_roundtrip_test.yaml`
(passing).

### Screenshots

| Before | After |
| :---: | :---: |
| <video src="https://github.com/user-attachments/assets/10dbfcd9-a40d-4dc1-9693-e3f0f2215f72" width="300" style="max-width: 300px;" controls></video> | <video src="https://github.com/user-attachments/assets/0bc580a5-85a9-41a5-a4c0-d0093393fb6c" width="300" style="max-width: 300px;" controls></video> |
| <video src="https://github.com/user-attachments/assets/f3a5ef2d-2c38-44b0-8db8-33caf5e9ada2" width="300" style="max-width: 300px;" controls></video> | <video src="https://github.com/user-attachments/assets/16ef74eb-9bac-485e-8903-31fb6ab3313d" width="300" style="max-width: 300px;" controls></video> |
| <video src="https://github.com/user-attachments/assets/f4f6ff17-7a49-42f4-aec4-71ad3e1553e6" width="300" style="max-width: 300px;" controls></video> | <video src="https://github.com/user-attachments/assets/bc782848-5303-4e26-b8c4-3f0c3b6cd933" width="300" style="max-width: 300px;" controls></video> |**

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/input/ENRMAutoLinkDetector.mm
+++ b/packages/react-native-enriched-markdown/ios/input/ENRMAutoLinkDetector.mm
@@ -134,39 +134,38 @@ static NSAttributedStringKey const ENRMAutomaticLinkAttributeName = @"ENRMAutoma
   NSRange localMatch = NSMakeRange(NSNotFound, 0);
   NSString *matchedUrl = [self tryMatchWord:word matchedRange:&localMatch];
 
-  if (matchedUrl != nil && localMatch.location != NSNotFound) {
-    NSRange matchRange = NSMakeRange(wordRange.location + localMatch.location, localMatch.length);
-
-    // A manual link overlapping the matched slice takes precedence — test the
-    // whole slice, not just its first character.
-    ENRMFormattingStore *store = _formattingStore;
-    if (store != nil && [store isStyleActive:ENRMInputStyleTypeLink inRange:matchRange]) {
-      [self clearAutoLinkInRange:wordRange];
-      return nil;
-    }
-
-    NSRange effectiveRange;
-    id existing = [textStorage attribute:ENRMAutomaticLinkAttributeName
-                                 atIndex:matchRange.location
-                          effectiveRange:&effectiveRange];
-    BOOL alreadyApplied = [matchedUrl isEqual:existing] && NSEqualRanges(effectiveRange, matchRange);
-
-    if (!alreadyApplied) {
-      [textStorage beginEditing];
-      [textStorage removeAttribute:ENRMAutomaticLinkAttributeName range:wordRange];
-      [textStorage addAttribute:ENRMAutomaticLinkAttributeName value:matchedUrl range:matchRange];
-      [self applyVisualStylingToRange:matchRange];
-      [textStorage endEditing];
-    }
-
-    if (outMatchedRange != NULL) {
-      *outMatchedRange = localMatch;
-    }
-  } else {
+  if (matchedUrl == nil || localMatch.location == NSNotFound) {
     [self clearAutoLinkInRange:wordRange];
     return nil;
   }
 
+  NSRange matchRange = NSMakeRange(wordRange.location + localMatch.location, localMatch.length);
+
+  // A manual link overlapping the matched slice takes precedence — test the
+  // whole slice, not just its first character.
+  ENRMFormattingStore *store = _formattingStore;
+  if (store != nil && [store isStyleActive:ENRMInputStyleTypeLink inRange:matchRange]) {
+    [self clearAutoLinkInRange:wordRange];
+    return nil;
+  }
+
+  NSRange effectiveRange;
+  id existing = [textStorage attribute:ENRMAutomaticLinkAttributeName
+                               atIndex:matchRange.location
+                        effectiveRange:&effectiveRange];
+  BOOL alreadyApplied = [matchedUrl isEqual:existing] && NSEqualRanges(effectiveRange, matchRange);
+
+  if (!alreadyApplied) {
+    [textStorage beginEditing];
+    [textStorage removeAttribute:ENRMAutomaticLinkAttributeName range:wordRange];
+    [textStorage addAttribute:ENRMAutomaticLinkAttributeName value:matchedUrl range:matchRange];
+    [self applyVisualStylingToRange:matchRange];
+    [textStorage endEditing];
+  }
+
+  if (outMatchedRange != NULL) {
+    *outMatchedRange = localMatch;
+  }
   return matchedUrl;
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by hryhoriiK97 — APPROVED** [review]
LGTM! I've just left one nit comment

File: packages/react-native-enriched-markdown/ios/input/ENRMAutoLinkDetector.mm [hunk 1]
```diff
@@ -107,61 +123,84 @@ - (nullable NSString *)detectNewLinkAtRange:(NSRange)wordRange inText:(NSString
...
-    }
+  // Tokens carrying markdown link syntax (e.g. "[label](url)") belong to the
+  // parser, not the auto-linker. Auto-linking them re-wraps the destination and
+  // yields a corrupted URL like "https://[label](url)" on serialize.
+  if ([word hasPrefix:@"["] || [word rangeOfString:@"]("].location != NSNotFound) {
+    [self clearAutoLinkInRange:wordRange];
+    return nil;
   }
 
-  NSString *matchedUrl = [self tryMatchWord:word];
```
hryhoriiK97, line 118: Nit (non-blocking): the `if/else` at the end reads slightly inside-out - `return matchedUrl` is only reachable from the `if` branch since the `else` returns early. Flipping to a guard clause would flatten the indentation and make the happy path read linearly:

```objc
if (matchedUrl == nil || localMatch.location == NSNotFound) {
    [self clearAutoLinkInRange:wordRange];
    return nil;
}

NSRange matchRange = NSMakeRange(wordRange.location + localMatch.location, localMatch.length);
// ... happy path continues flat ...
return matchedUrl;
```

PR #536, opened 2026-07-11
fix(ios): visible source fallback when RaTeX cannot parse a math span

EthanCai9568:
### What/Why?

When `RaTeXEngine.parse` throws, the iOS math pipeline currently swallows the failure:

- `ENRMMathInlineAttachment.prepareIfNeeded` leaves `_renderResult` nil → zero-size bounds + nil image → the span **silently vanishes** from the paragraph (no error, no trace — prose closes up seamlessly).
- `ENRMMathContainerView.applyLatex:` has the same shape for display blocks: nil render result → zero intrinsic size → the block collapses into an invisible bar.

Malformed LaTeX (`$x^$`, `$\frac{1}$`, fused `$x+1$$y+2$`) is routine in LLM/OCR-produced markdown, and invisible content loss is the worst possible failure mode for it. This PR makes parse failures render the **original source, delimiters included**, instead:

- inline: body style (ambient font size + configured math color), baseline-aligned, no loud error UI;
- display: math container style, wraps to the available width, horizontal scrolling disabled in fallback mode;
- copy actions and the accessibility label already read `cachedLatex` and keep working unchanged.

### Testing

Verified on a production K12 app (iPhone simulator dev build) that has been running this change via `patch-package` against 0.7.3 (the touched files are byte-identical to current `main`):

- `$x^$` and `$x+1$$y+2$` (md4c fuses the glued pair into one span whose body contains `$$`) previously vanished; with the patch they render as visible source text.
- Block-level `$$\frac{1}$$` previously collapsed; now renders `$$\frac{1}$$` wrapped inside the math container.
- Regression sweep: root-level `$$` blocks (`cases`, `pmatrix`, quadratic formula), inline math, CJK-adjacent math — unchanged.
- Corpus datapoint: 4,977 real-world math spans through this RaTeX build — 92.4% parse OK; the 7.6% failures are exactly the population this fallback makes visible.

Repro snippets for the example app:

```
坏公式 $x^$ 应当看得见这行字后面有东西。
相邻：$x+1$$y+2$ 尾。
前文。\n\n$$\frac{1}$$\n\n后文。
```

Note: iOS only. `MathInlineSpan.kt` / `MathContainerView.kt` on Android have the same silent-failure shape, but we ship iOS-only and did not device-verify an Android fix, so it is deliberately out of scope here.

### PR Checklist

- [ ] Code compiles and runs on Android (Android sources untouched — iOS-only change)
- [ ] Updated documentation/README if applicable (no documented behavior contract changes)
- [ ] Ran example app to verify changes (verified in a production app via patch-package instead — see Testing)
- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/attachments/ENRMMathInlineAttachment.m
+++ b/packages/react-native-enriched-markdown/ios/attachments/ENRMMathInlineAttachment.m
@@ -14,13 +14,31 @@
 
 - (void)prepareIfNeeded
 {
-  if (_renderResult)
+  if (_renderResult || _fallbackSource)
     return;
 
   RCTUIColor *color = self.mathTextColor ?: [RCTUIColor blackColor];
   ENRMRaTeXRenderResult *result = [ENRMRaTeXBridge parse:self.latex displayMode:NO fontSize:self.fontSize color:color];
-  if (!result)
+  if (!result) {
+    // RaTeX parse failure: fall back to the original source, delimiters
+    // included, in body style — the formula stays visible and copyable
+    // instead of collapsing into an invisible zero-size box.
+    CGFloat size = self.fontSize > 0 ? self.fontSize : 16.0;
+    UIFont *font = [UIFont systemFontOfSize:size];
+    NSString *source = [NSString stringWithFormat:@"$%@$", self.latex ?: @""];
+    _fallbackSource = [[NSAttributedString alloc] initWithString:source
+                                                      attributes:@{
+                                                        NSFontAttributeName : font,
+                                                        NSForegroundColorAttributeName : color,
+                                                      }];
+    CGRect bounds = [_fallbackSource boundingRectWithSize:CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX)
+                                                  options:NSStringDrawingUsesLineFragmentOrigin
+                                                  context:nil];
+    _mathAscent = font.ascender;
+    _mathDescent = -font.descender;
+    _cachedSize = CGSizeMake(ceil(bounds.size.width), ceil(_mathAscent + _mathDescent));
     return;
+  }
 
   _renderResult = result;
   _mathAscent = result.ascent;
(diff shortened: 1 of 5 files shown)


packages/react-native-enriched-markdown/ios/attachments/ENRMMathInlineAttachment.m
```diff
@@ -14,13 +14,31 @@ @implementation ENRMMathInlineAttachment
...
+    // included, in body style — the formula stays visible and copyable
+    // instead of collapsing into an invisible zero-size box.
+    CGFloat size = self.fontSize > 0 ? self.fontSize : 16.0;
+    UIFont *font = [UIFont systemFontOfSize:size];
+    NSString *source = [NSString stringWithFormat:@"$%@$", self.latex ?: @""];
+    _fallbackSource = [[NSAttributedString alloc] initWithString:source
+                                                      attributes:@{
+                                                        NSFontAttributeName : font,
+                                                        NSForegroundColorAttributeName : color,
+                                                      }];
```
eszlamczyk: I think it'd be great if we extracted this fallback-string construction into a shared helper, as basicly the same thing is duplicated in `ios/views/ENRMMathContainerView.m:110-118`. Only the delimiter ($ vs $$), font-size source, and color default differ.

Since we already keep small helpers in ios/utils/, something like:
```objective-c
// utils/ENRMMathFallback.h
static const CGFloat kENRMMathFallbackDefaultFontSize = 16.0;

NSAttributedString *ENRMMathFallbackString(NSString *latex,
                                           NSString *delimiter,
                                           CGFloat fontSize,
                                           RCTUIColor *color);
```

would let both call sites collapse to a one-liner:
```objective-c
_fallbackSource = ENRMMathFallbackString(self.latex, @"$", self.fontSize, color);
```
```objective-c
_mathView.fallbackSource = ENRMMathFallbackString(latex, @"$$", config.mathFontSize, config.mathColor);
```

This also names the 16.0 magic number and lets the "never render as an invisible box, typeset the source instead" note live in one place instead of being repeated across the shared header and both .m files.

eszlamczyk (changes requested):
Hi @EthanCai9568 thanks for the pull request, code is verified and working! I just have small issue with it. When you fix this we should be able to merge it.

eszlamczyk (approved):

PR #537, opened 2026-07-11
feat(ios): render in-paragraph display math instead of dropping its delimiters

EthanCai9568:
### What/Why?

`MarkdownNodeTypeLatexMathDisplay` has no case in `RendererFactory`, so display math **inside a paragraph** (mid-paragraph `$$x$$`, or `$$…$$` adjacent to a soft break) falls through the nil-renderer child dump: the BODY leaks as plain text and the `$$` delimiters are silently dropped — e.g.

```
新的积是
$$(a\times 10)\times (b\div 10)$$
也就是 …
```

renders the literal characters `(a\times 10)\times (b\div 10)` with no math and no `$$`.

This PR routes the node to `ENRMMathInlineRenderer`, typesetting it as an inline attachment. Root-level display blocks are unaffected (`SegmentRenderer` intercepts them before the factory is consulted).

### Testing

Device-verified (production app via patch-package, files byte-identical to `main`): the snippet above and mid-paragraph `段中 $$x+1$$ 嵌入文本继续。` now typeset; standalone `$$` blocks unchanged.

### PR Checklist

- [x] Code compiles and runs on Android (Android sources untouched — iOS-only change)
- [x] Ran example app to verify changes (verified in a production app via patch-package instead — see Testing)
- [x] E2E tests are passing


eszlamczyk (commented):
Thanks for this, @EthanCai9568 - and good catch. The old behavior here (silently dropping the $$…$$ delimiters and leaking the body as plain text) is a genuine bug, and "render the math instead of dropping it" is the right instinct.

Digging in, though, I think this is fixing the symptom not the problem itself. the proper fix lives one layer down, so I'm going to take it from here.

Why it happens: md4c parses all math — both `$…$` and `$$…$$` — as inline spans, so display math always lands nested inside a Paragraph, even when it's on its own line. We already compensate for that in the shared C++ core: `promoteDisplayMathFromParagraphs` in `packages/core/cpp/parser/MD4CParser.cpp` hoists display-math nodes out of their paragraph up to the top level so the renderers treat them as blocks. The gap is that it currently only handles display math at the end of a paragraph. When there's text on both sides:
```
text
$$
x = y
$$
text
```
the node stays buried mid-paragraph and falls through to the renderer factory — which is exactly the case you patched.

What I'll do: generalize that core promotion pass to also split a paragraph around interior isolated display math. Fixing it in the core means it renders as a real block — its own line, centered, honoring the block mathStyle config — on iOS, Android, and web all at once, rather than each platform separately rendering it inline with inline-math styling.

Why your change is still useful: your renderer-factory case is exactly the right fallback for genuinely mid-line display math like a $$x$$ b, where the math really is inline and dropping it would still be a bug. So I'll keep it (with attribution) and layer the core fix on top of it, rather than closing this out.

I'll push the core change onto this branch shortly. Thanks again for surfacing it!

hryhoriiK97 (approved):
LGTM! @EthanCai9568 thank you for the contribution! This change will be available in tomorrow's nightly release

Issue #380, opened 2026-06-05
Support full Markdown image display and original image saving


wekor:
Markdown images are currently clipped when rendered, and the saved image in the photo album is also clipped. Please support displaying images with their original aspect ratio using a fit/contain mode, and save the original image source instead of the cropped rendered view.

eszlamczyk:
Hi @wekor thanks for the bug report, the part of the fit/contain mode was adressed in pull request, the original image saving is a biger problem and It will be adressed in other pull request!

PR #539, opened 2026-07-15
feat: add maxHeight, aspectRatio and resizeMode to block image styling

### LINKED ISSUE #195: Add support for block images that aren't a fixed height and aspect ratio
The image style only accepts a fixed height value. This makes it impossible to render images responsively. There's no way to say "fit this image proportionally within a maximum height" or "render at a specific aspect ratio."

It's also not possible to control how the image fills its container — whether it should be letterboxed, cropped to fill, or stretched.

Three additions to the image style config:

`maxHeight`: this would replace the current height as the primary sizing knob for most use cases.

`aspectRatio`: a width / height ratio (e.g. 16/9, 4/3). The image fills the available width and its height is derived from the ratio, ignoring height/maxHeight when set.

`resizeMode`: controls how the image fills its bounds, analogous to React Native's existing resizeMode / CSS object-fit. Values: contain (scale to fit, letterbox), cover (scale to fill, crop), stretch (fill exactly, distort), center (no scaling).

Ideally these would have a clear precedence order, e.g. aspectRatio > maxHeight > height, with resizeMode applying independently on top.

Setting a large fixed height (e.g. 400) as a workaround, but this breaks images that are wider than tall — they get extra whitespace.

**Additional context**
The current default is height: 200 with no resizeMode control. maxHeight + resizeMode: contain would cover the vast majority of article/blog rendering use cases out of the box. aspectRatio + resizeMode: cover is the standard pattern for banner/hero images. These are all well-established concepts in both React Native and CSS, so the mental model would be immediately familiar to most users.

eszlamczyk:
### What/Why?

Closes #195.
Adresses #380

Block images previously supported only a fixed `markdownStyle.image.height` (default `200`): the box was always `containerWidth × height`, scaled to width and clipped. This made responsive rendering impossible - no way to fit an image proportionally within a max height, pin an aspect ratio, or control fill behavior.

This PR adds three sizing knobs to `markdownStyle.image`, mirroring RN `resizeMode` / CSS `object-fit`:

- `maxHeight` - image is fitted proportionally within a max height (intrinsic aspect ratio preserved)
- `aspectRatio` - width/height ratio (e.g. `16/9`); fills available width, height derived from the ratio
- `resizeMode` - `contain` | `cover` | `stretch` | `center` | `none` (`center` is scale-down-only, never upscales)

**Precedence:** `aspectRatio` > `maxHeight` > `height`, with `resizeMode` applying independently on top.

**Backward compatibility:** when neither `maxHeight` nor `aspectRatio` is set, block images take the exact legacy fill-width code path and `resizeMode` is a no-op - existing `height`-only users see zero pixel change.

Implementation notes:

- **iOS / Android / Web** share one geometry model: box derivation from the precedence rule + per-mode transform of the source within the box (manual canvas math - block images are text attachments/spans, not image views). Web maps to `objectFit` / `aspect-ratio` / `max-height` CSS.
- **Async intrinsic size:** `maxHeight` boxes can only be resolved once the image loads, so the container height must settle afterwards. This required re-measure infrastructure on both platforms:
  - **Android:** new `MarkdownTextState` (C++) wired into the text and GFM container shadow nodes, mirroring the input's `forceHeightRecalculationCounter` → `dirtyLayout()` mechanism; `ImageSpan` notifies the owning view after a post-load reflow, which re-stores the measurement and bumps state only when the height actually changed. `MeasurementStore` also resolves image boxes width-/cache-aware before building measurement layouts.
  - **iOS:** the measurement style fingerprint now includes `maxHeight`/`aspectRatio` (stale `MeasurementCache` hits caused wrong container heights when only those fields changed), and a new `ENRMImageLayoutObserver` protocol lets block attachments notify the hosting component (`EnrichedMarkdownText` / `EnrichedMarkdown`) to purge stale cache entries and request a height update after the image settles.
- **Border radius** is always applied to the final visible region (box ∩ drawn image rect), so letterboxed modes (`contain`/`center`) round the image itself and cropping modes (`cover`/`none`) round the box edges.
- Processed-image cache keys on both platforms now include `resizeMode` and the resolved box height.
- Codegen: `resizeMode` is a plain nested `string` field (nested struct fields can't use `WithDefault`); the `'cover'` default is applied in JS normalization, with defensive native fallbacks.

Web caveat: CSS `border-radius` on `` rounds the element box, so `contain`/`center` + radius won't round the letterboxed content itself (platform limitation).

### Testing

Manual, via the storybook Image story (`Block › Image`) which gained `maxHeight`, `aspectRatio` and `resizeMode` controls:

- Legacy `height`-only rendering is unchanged
- All five `resizeMode` values across wide/tall/square sources; `center` verified not to upscale small images
- `maxHeight` with tall and short images - container settles to the fitted height after load (no junk space below, no clipping), text below reflows correctly
- `aspectRatio` fills width with height derived from the ratio; precedence over `maxHeight`/`height` verified
- Toggling `resizeMode`/`maxHeight` back and forth - no stale cached bitmaps or stale container heights
- Border radius rounds the visible image in all modes
- Verified on both flavors (`commonmark` single text view and `github` container)

### PR Checklist

- [x] E2E tests are passing


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/spans/ImageSpan.kt
```diff
@@ -39,14 +41,61 @@ class ImageSpan(
...
+        val fitted = targetWidth.toFloat() * intrinsicHeight / intrinsicWidth
+        return minOf(maxHeightPx.toFloat(), fitted).toInt().coerceAtLeast(1)
+      }
+      return maxHeightPx
+    }
+    return height
+  }
+
+  fun prepareForMeasurement(widthPx: Int) {
+    if (!dynamicBoxHeight || viewRef != null || widthPx <= 0) return
```
hryhoriiK97: `prepareForMeasurement` bails when `viewRef != null`. After first paint we store the display spannable, and width-change remasure reuses it — so `aspectRatio` / `maxHeight` can keep the old height.

Don't just delete the `viewRef` check (Yoga width ≠ available width with leading margins). Sync height on measure **and** when width changes before load (see also `updateWidthAndRecreate` — it only resolves when `sourceDrawable != null`).

hryhoriiK97 (commented):
Overall LGTM! I've left one comment

hryhoriiK97 (approved):

PR #542, opened 2026-07-16
feat: add horizontalOverflow style prop for edge-to-edge scrollable tables

### LINKED ISSUE #378: Table horizontal container does not fit the full horizontal contents of the table when scrolling to the right
**Describe the bug**
Tables horizontal container and scrolling no longer fits the full horizontal contents of the table when scrolling to the right. See video.

**To Reproduce**
1. Render a table with several columns that need horizontal scrolling to the right
2. Scroll to the right and see that the table is cut off

**Expected behavior**
Expect to be able to scroll to see the full table

 - Device: iPhone 17 sim
 - OS: iOS 26
 - Version: 0.6.0

**Additional context**
I don't recall this being a problem in 0.5.0

hryhoriiK97:
### What/Why?
When set, scrollable tables extend beyond the markdown container by the specified amount on each side, allowing wide tables to reach screen edges instead of appearing clipped. Users set the value to match their parent's horizontal padding.

- iOS: uses contentInset on the scroll view for proper alignment
- Android: uses scroll view padding with clipToPadding
- macOS: uses NSScrollView contentInsets
- Web: not supported — React Native Web sets `overflow: hidden` on all Views, preventing children from extending beyond parent bounds. Desktop browsers already handle horizontal table scrolling well via `overflowX: auto`.
- Negative values are clamped to 0
- No effect on tables that fit within the container

#### Screenshots

| iOS before | iOS after |
| - | - |
| <video src="https://github.com/user-attachments/assets/f3aa5863-c770-4eb8-a3aa-37098bfceb98" width=300 /> | <video src="https://github.com/user-attachments/assets/f74d1c53-9430-4ccc-80ee-a206917417d2" width=300 /> |

| Android before | Android after |
| - | - |
| <video src="https://github.com/user-attachments/assets/0fa27881-d487-4fcb-b6cb-ac373edd2a18" width=300 /> | <video src="https://github.com/user-attachments/assets/2b6daaa1-6bc6-4f82-89ac-45f8e46d3912" width=300 /> |

Closes: #378 

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -572,13 +572,13 @@ class EnrichedMarkdown
       val containerWidth = width
       if (containerWidth <= 0) return
 
-      val needsOverflow =
+      val needsOverhang =
         segmentViews.any { view ->
           view is TableContainerView &&
             ceil(view.tableStyle.horizontalOverflow.toDouble()).toInt() > 0
         }
-      if (clipChildren == needsOverflow) {
-        clipChildren = !needsOverflow
+      if (clipChildren == needsOverhang) {
+        clipChildren = !needsOverhang
       }
 
       var currentY = 0
@@ -592,18 +592,18 @@ class EnrichedMarkdown
 
         currentY += segment?.segmentMarginTop ?: 0
 
-        val tableOverflow =
+        val overhang =
           if (view is TableContainerView) {
             max(ceil(view.tableStyle.horizontalOverflow.toDouble()).toInt(), 0)
           } else {
             0
           }
 
-        if (tableOverflow > 0) {
-          val extendedWidth = containerWidth + tableOverflow * 2
+        if (overhang > 0) {
+          val extendedWidth = containerWidth + overhang * 2
           val extWidthSpec = MeasureSpec.makeMeasureSpec(extendedWidth, MeasureSpec.EXACTLY)
           view.measure(extWidthSpec, heightSpec)
-          view.layout(-tableOverflow, currentY, containerWidth + tableOverflow, currentY + view.measuredHeight)
+          view.layout(-overhang, currentY, containerWidth + overhang, currentY + view.measuredHeight)
         } else {
           view.measure(widthSpec, heightSpec)
           view.layout(0, currentY, containerWidth, currentY + view.measuredHeight)
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -32,6 +32,7 @@ import com.swmansion.enriched.markdown.views.TableContainerView
 import java.util.EnumSet
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
+import kotlin.math.ceil
 import kotlin.math.max
 
 class EnrichedMarkdown
@@ -571,6 +572,15 @@ class EnrichedMarkdown
       val containerWidth = width
       if (containerWidth <= 0) return
 
+      val needsOverflow =
+        segmentViews.any { view ->
+          view is TableContainerView &&
+            ceil(view.tableStyle.horizontalOverflow.toDouble()).toInt() > 0
+        }
+      if (clipChildren == needsOverflow) {
+        clipChildren = !needsOverflow
+      }
+
       var currentY = 0
       val lastIndex = segmentViews.lastIndex
       val widthSpec = MeasureSpec.makeMeasureSpec(containerWidth, MeasureSpec.EXACTLY)
@@ -584,13 +594,12 @@ class EnrichedMarkdown
 
         val tableOverflow =
           if (view is TableContainerView) {
-            max(view.tableStyle.horizontalOverflow.toInt(), 0)
+            max(ceil(view.tableStyle.horizontalOverflow.toDouble()).toInt(), 0)
           } else {
             0
           }
 
         if (tableOverflow > 0) {
-          if (clipChildren) clipChildren = false
           val extendedWidth = containerWidth + tableOverflow * 2
           val extWidthSpec = MeasureSpec.makeMeasureSpec(extendedWidth, MeasureSpec.EXACTLY)
           view.measure(extWidthSpec, heightSpec)
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
@@ -32,6 +32,7 @@ import com.swmansion.enriched.markdown.views.TableContainerView
 import java.util.EnumSet
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
+import kotlin.math.max
 
 class EnrichedMarkdown
   @JvmOverloads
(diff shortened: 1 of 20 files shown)


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -580,9 +581,24 @@ class EnrichedMarkdown
...
 
         currentY += segment?.segmentMarginTop ?: 0
-        view.measure(widthSpec, heightSpec)
 
-        view.layout(0, currentY, containerWidth, currentY + view.measuredHeight)
+        val tableOverflow =
+          if (view is TableContainerView) {
+            max(view.tableStyle.horizontalOverflow.toInt(), 0)
+          } else {
+            0
```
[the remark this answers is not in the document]
hryhoriiK97: [b1c17d1](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/b1c17d105101d25f63fffd408436973e8229d576)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
```diff
@@ -327,12 +327,31 @@ class TableContainerView(
...
     bottom: Int,
   ) {
-    scrollView.layout(0, 0, right - left, bottom - top)
     val viewWidth = right - left
-    scrollView.isHorizontalScrollBarEnabled = totalTableWidth > viewWidth
+    val overflow = max(tableStyle.horizontalOverflow.toInt(), 0)
+    val originalWidth = viewWidth - overflow * 2
+    val needsEdgeToEdge = overflow > 0 && totalTableWidth > originalWidth
+
+    if (needsEdgeToEdge) {
```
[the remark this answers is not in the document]
hryhoriiK97: [b1c17d1](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/b1c17d105101d25f63fffd408436973e8229d576)

packages/react-native-enriched-markdown/ios/views/TableContainerView.m
```diff
@@ -554,10 +554,33 @@ - (void)layoutSubviews
...
+    _scrollView.contentInset = UIEdgeInsetsZero;
+    _scrollView.contentSize = CGSizeMake(containerWidth, _totalTableHeight);
+    _scrollView.scrollEnabled = NO;
+    _gridContainer.frame = CGRectMake(overflow, 0, _totalTableWidth, _totalTableHeight);
+  } else {
+    _scrollView.contentInset = UIEdgeInsetsZero;
+    _scrollView.contentSize = CGSizeMake(MAX(_totalTableWidth, containerWidth), _totalTableHeight);
+    _scrollView.scrollEnabled = (_totalTableWidth > containerWidth);
+    _gridContainer.frame = CGRectMake(0, 0, _totalTableWidth, _totalTableHeight);
+  }
```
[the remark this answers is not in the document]
hryhoriiK97: [b1c17d1](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/b1c17d105101d25f63fffd408436973e8229d576)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
```diff
@@ -327,12 +327,30 @@ class TableContainerView(
     right: Int,
     bottom: Int,
   ) {
-    scrollView.layout(0, 0, right - left, bottom - top)
     val viewWidth = right - left
-    scrollView.isHorizontalScrollBarEnabled = totalTableWidth > viewWidth
+    val overflow = max(ceil(tableStyle.horizontalOverflow.toDouble()).toInt(), 0)
```
eszlamczyk: nitpick: I believe calling this `overflow` is missleading (for me "overflow" sounds more like a "table being to wide and then overflowing"). Maybe calling it something like "configuredOverflow" would be better

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
```diff
@@ -327,12 +327,30 @@ class TableContainerView(
...
+    val needsEdgeToEdge = overflow > 0 && totalTableWidth > viewWidth
```
eszlamczyk: This doesn't seem right.

viewWidth here is already the extended frame - `layoutSegments()` lays the table out at `containerWidth + horizontalOverflow * 2` (`EnrichedMarkdown.kt:606`) - so this condition only triggers when the table is wider than the container plus both bleed areas.

Tables whose width falls between the container width and the extended width land in the else if (`overflow > 0`) branch below, where the behavior is an inconsistent mix: `clipToPadding = true` clips them at the container edge, and the scrollbar is turned off - but `isHorizontalScrollBarEnabled = false` only hides the indicator, it doesn't disable scrolling (the scroll range is `totalTableWidth + 2 * overflow − viewWidth > 0`). So a slightly-overflowing table renders clipped with no hint that it scrolls. 

Concrete example: 
`container 360, horizontalOverflow 20`
any table between 361–400 wide hits this. 

The RTL scrollTo at line 350 also uses the full viewWidth in this branch, so these tables skip their initial scroll-to-end as well.

Comparing against the actual content width puts any overflowing table on the edge-to-edge path, and matches the threshold iOS uses (`originalWidth = containerWidth - overflow * 2` in `TableContainerView.m`):
```kotlin
      val contentWidth = viewWidth - overflow * 2
      val needsEdgeToEdge = overflow > 0 && totalTableWidth > contentWidth
```

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
```diff
@@ -327,12 +327,30 @@ class TableContainerView(
...
+      scrollView.clipToPadding = false
+      scrollView.isHorizontalScrollBarEnabled = true
+    } else if (overflow > 0) {
+      scrollView.setPadding(overflow, 0, overflow, 0)
+      scrollView.clipToPadding = true
+      scrollView.isHorizontalScrollBarEnabled = false
+    } else {
+      scrollView.setPadding(0, 0, 0, 0)
+      scrollView.isHorizontalScrollBarEnabled = totalTableWidth > viewWidth
+    }
```
eszlamczyk: Nitpick: I believe we could write this cleaner:
1. We always set `scrollView.setPadding(overflow, 0, overflow, 0)` (in else clause overflow is clamped to 0 so it is logicaly the same)
2. We could use `totalTableWidth > contentWidth` condition as a check for clip to padding for example `scrollView.clipToPadding = !tableOverflows`

packages/react-native-enriched-markdown/ios/views/TableContainerView.m
```diff
@@ -554,13 +554,43 @@ - (void)layoutSubviews
...
+    _scrollView.contentSize = CGSizeMake(containerWidth, _totalTableHeight);
+    _scrollView.scrollEnabled = NO;
+    _gridContainer.frame = CGRectMake(overflow, 0, _totalTableWidth, _totalTableHeight);
+  } else {
+    _scrollView.contentInset = UIEdgeInsetsZero;
+    _scrollView.contentOffset = CGPointZero;
+    _scrollView.contentSize = CGSizeMake(MAX(_totalTableWidth, containerWidth), _totalTableHeight);
+    _scrollView.scrollEnabled = (_totalTableWidth > containerWidth);
+    _gridContainer.frame = CGRectMake(0, 0, _totalTableWidth, _totalTableHeight);
+  }
```
eszlamczyk: Possibly there is the same simplification there as on anrdoid

eszlamczyk (commented):
Overall okay, just this section where the paddings/clippings/overflows are being calculated doesn't sit well with me, please address the comments first

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
hryhoriiK97: Good call, renamed to overhang - less ambiguous than overflow which sounds like a runtime state. Kept the public API as horizontalOverflow since it matches CSS/RN conventions and is easier to discover.

packages/react-native-enriched-markdown/ios/views/TableContainerView.m
hryhoriiK97: [bb6aadf](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/bb6aadf8f492cd29a355025488a227783ce3b347)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
hryhoriiK97: [bb6aadf](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/bb6aadf8f492cd29a355025488a227783ce3b347)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
hryhoriiK97: [bb6aadf](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/bb6aadf8f492cd29a355025488a227783ce3b347)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/TableContainerView.kt
hryhoriiK97: [bb6aadf](https://github.com/software-mansion/react-native-enriched-markdown/pull/542/commits/bb6aadf8f492cd29a355025488a227783ce3b347)

eszlamczyk (approved):

PR #544, opened 2026-07-16
feat: custom image request headers

### LINKED ISSUE #369: Support custom request headers for remote images, especially Referer
Some Markdown image URLs are protected by CDN / hotlink protection rules and require a `Referer` HTTP header to load successfully.

Currently `EnrichedMarkdownText` downloads remote images internally, so app code cannot pass request headers for those image requests. This makes images fail to render even though the same URL works when requested with the correct `Referer`.

Please add an API for passing headers to remote image requests, for example:

```tsx
<EnrichedMarkdownText
  markdown={markdown}
  imageRequestHeaders={{
    Referer: 'https://example.com',
  }}
/>

### PR DESCRIPTION
**Pull request by eszlamczyk**
### What/Why?

Closes #369

Adds an `imageRequestHeaders` prop to `EnrichedMarkdownText` that attaches custom HTTP headers to remote image requests. Some Markdown image URLs are protected by CDN / hotlink protection and require a `Referer` (or `Authorization`) header to load - previously images were downloaded internally with no way to pass headers, so those images silently failed to render.

```tsx
<EnrichedMarkdownText
  markdown={markdown}
  imageRequestHeaders={{ Referer: 'https://example.com' }}
/>
```

#### Implementation notes

- Headers are per-view, not global: the downloaders are process-wide singletons, so headers travel per-request (view -> `StyleConfig` ->`ImageSpan`/`ENRMImageAttachment` -> downloader) instead of being stored on the singleton - two views with different headers would cause race condition if done otherwise.
- Cache: headers participate in the memory cache keys, in-flight request dedup, and the attachment registry key (iOS), so the same URL fetched with different headers is **cached separately.** The HTTP disk caches (OkHttp / `NSURLCache`) remain URL-keyed.
- Android measurement pass: the C++ measure-props serialization (`conversions.h toDynamic`) only forwards a curated subset of props; `imageRequestHeaders` is now included. Without this, the Fabric measurement render fired a duplicate header-less request per image.
- **Standalone Android package**: `EnrichedMarkdownText.setImageRequestHeaders(Map<String, String>)` on the view and an `imageRequestHeaders` parameter on the Compose wrapper. Threaded through `Renderer.configure` → `RendererConfig` → `ImageSpan` (no `StyleConfig` mutation, since standalone configs are user-owned and shareable across views).
- **Web not supported** - browsers don't allow custom headers on `` requests; documented on the prop.

### Testing

Manual verification against a local mock server:

<details>
<summary> Run this node server </summary>

```js
/**
 * Mock CDN for testing `imageRequestHeaders`.
 *
 * Always responds 200 OK with a generated PNG:
 *   - GET /referer-image.png  → green if `Referer: https://example.com`, red otherwise
 *   - GET /auth-image.png     → green if `Authorization: Bearer secret-token`, red otherwise
 *
 * Responses are sent with `Cache-Control: no-store` so the URL-keyed disk
 * caches (OkHttp / NSURLCache) don't mask header changes between test runs.
 *
 * Run:  node scripts/mock-cdn-server.js
 * URLs: iOS simulator      http://localhost:3001/referer-image.png
 *       Android emulator   http://10.0.2.2:3001/referer-image.png
 *       Physical device    http://<your-lan-ip>:3001/referer-image.png
 */
const http = require('http');
const zlib = require('zlib');

const PORT = 3001;

const CRC_TABLE = Array.from({ length: 256 }, (_, n) => {
  let c = n;
  for (let k = 0; k < 8; k++) c = c & 1 ? 0xedb88320 ^ (c >>> 1) : c >>> 1;
  return c >>> 0;
});

function crc32(buf) {
  let c = 0xffffffff;
  for (const byte of buf) c = CRC_TABLE[(c ^ byte) & 0xff] ^ (c >>> 8);
  return (c ^ 0xffffffff) >>> 0;
}

function chunk(type, data) {
  const length = Buffer.alloc(4);
  length.writeUInt32BE(data.length);
  const body = Buffer.concat([Buffer.from(type, 'ascii'), data]);
  const crc = Buffer.alloc(4);
  crc.writeUInt32BE(crc32(body));
  return Buffer.concat([length, body, crc]);
}

function solidPng(width, height, [r, g, b]) {
  const ihdr = Buffer.alloc(13);
  ihdr.writeUInt32BE(width, 0);
  ihdr.writeUInt32BE(height, 4);
  ihdr[8] = 8; // bit depth
  ihdr[9] = 2; // color type: RGB
  const row = Buffer.concat([Buffer.from([0]), Buffer.alloc(width * 3)]);
  for (let x = 0; x < width; x++) row.set([r, g, b], 1 + x * 3);
  const raw = Buffer.concat(Array.from({ length: height }, () => row));
  return Buffer.concat([
    Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]),
    chunk('IHDR', ihdr),
    chunk('IDAT', zlib.deflateSync(raw)),
    chunk('IEND', Buffer.alloc(0)),
  ]);
}

const GREEN = solidPng(400, 200, [46, 160, 67]);
const RED = solidPng(400, 200, [218, 54, 51]);

const ROUTES = {
  '/referer-image.png': (req) => req.headers.referer === 'https://example.com',
  '/auth-image.png': (req) => req.headers.authorization === 'Bearer secret-token',
};

http
  .createServer((req, res) => {
    const check = ROUTES[req.url];
    if (!check) {
      res.writeHead(404).end('Not found');
      return;
    }
    const ok = check(req);
    console.log(
      `${new Date().toISOString()} GET ${req.url} → ${ok ? 'GREEN (header ok)' : 'RED (header missing/wrong)'}`,
      `| Referer: ${req.headers.referer ?? '-'} | Authorization: ${req.headers.authorization ?? '-'}`
    );
    res
      .writeHead(200, {
        'Content-Type': 'image/png',
        'Cache-Control': 'no-store',
      })
      .end(ok ? GREEN : RED);
  })
  .listen(PORT, () => console.log(`Mock CDN running on http://localhost:${PORT}`));
```
</details>

The server always returns 200 with a **green** image when the expected header is present and a **red** one otherwise, and logs the received
headers per request — so both the rendered result and the actual wire traffic are verifiable. Routes cover `Referer` and `Authorization`
shapes. Responses are `Cache-Control: no-store` so disk caches don't mask header changes between runs.

Verified:

- **iOS (RN)**: green image with the prop set.
- **Android (RN, both flavors)**: green image; server log confirms a single request carrying `Referer: https://example.com` (measurement +
display renders coalesce via in-flight dedup after the `conversions.h` fix).
- **Android (standalone, Compose)**: green image, single request with the header.

### Screenshots

| | No Header | Wrong Header | Correct Header |
| -- | -- | -- | -- |
| Android | image | image | image |
| Android (native) | image | image | image |
| IOS | image | image | image |

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/styles/StyleConfig.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/styles/StyleConfig.kt
@@ -166,6 +166,8 @@ class StyleConfig(
     ImageStyle.fromReadableMap(map, styleParser)
   }
 
+  var imageRequestHeaders: Map<String, String> = emptyMap()
+
   val inlineImageStyle: InlineImageStyle by lazy {
     val map =
       requireNotNull(style.getMap("inlineImage")) {
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/styles/StyleConfig.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/styles/StyleConfig.kt
@@ -339,7 +339,8 @@ class StyleConfig(
       spoilerStyle == other.spoilerStyle &&
       superscriptStyle == other.superscriptStyle &&
       subscriptStyle == other.subscriptStyle &&
-      highlightStyle == other.highlightStyle
+      highlightStyle == other.highlightStyle &&
+      imageRequestHeaders == other.imageRequestHeaders
   }
 
   override fun hashCode(): Int {
@@ -366,6 +367,7 @@ class StyleConfig(
     result = 31 * result + superscriptStyle.hashCode()
     result = 31 * result + subscriptStyle.hashCode()
     result = 31 * result + highlightStyle.hashCode()
+    result = 31 * result + imageRequestHeaders.hashCode()
     return result
   }
 }
--- a/packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
+++ b/packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
@@ -1,5 +1,6 @@
 #import "ENRMImageDownloader.h"
 #import "ENRMImageAttachment.h"
+#import <CommonCrypto/CommonDigest.h>
 #include <TargetConditionals.h>
 
 static const NSUInteger kDiskCacheMemoryCapacity = 10 * 1024 * 1024;
@@ -19,11 +20,18 @@ NSString *ENRMImageCacheKey(NSString *url, NSDictionary<NSString *, NSString *> 
     return url;
   }
   NSArray<NSString *> *names = [headers.allKeys sortedArrayUsingSelector:@selector(compare:)];
-  NSMutableString *key = [NSMutableString stringWithString:url];
+  NSMutableString *joined = [NSMutableString string];
   for (NSString *name in names) {
-    [key appendFormat:@"|%@:%@", name, headers[name]];
+    [joined appendFormat:@"%@:%@\n", name, headers[name]];
   }
-  return [key copy];
+  NSData *data = [joined dataUsingEncoding:NSUTF8StringEncoding];
+  unsigned char digest[CC_SHA256_DIGEST_LENGTH];
+  CC_SHA256(data.bytes, (CC_LONG)data.length, digest);
+  NSMutableString *hash = [NSMutableString stringWithCapacity:CC_SHA256_DIGEST_LENGTH * 2];
+  for (int i = 0; i < CC_SHA256_DIGEST_LENGTH; i++) {
+    [hash appendFormat:@"%02x", digest[i]];
+  }
+  return [NSString stringWithFormat:@"%@|%@", url, hash];
 }
 
 @implementation ENRMImageDownloader {
--- a/packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
+++ b/packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
@@ -13,6 +13,19 @@ static inline NSUInteger ENRMImageByteCost(RCTUIImage *image)
   return CGImageGetBytesPerRow(cgImage) * CGImageGetHeight(cgImage);
 }
 
+NSString *ENRMImageCacheKey(NSString *url, NSDictionary<NSString *, NSString *> *headers)
+{
+  if (headers.count == 0) {
+    return url;
+  }
+  NSArray<NSString *> *names = [headers.allKeys sortedArrayUsingSelector:@selector(compare:)];
+  NSMutableString *key = [NSMutableString stringWithString:url];
+  for (NSString *name in names) {
+    [key appendFormat:@"|%@:%@", name, headers[name]];
+  }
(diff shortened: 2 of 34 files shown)


packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
```diff
@@ -13,6 +13,19 @@ static inline NSUInteger ENRMImageByteCost(RCTUIImage *image)
...
+  if (headers.count == 0) {
+    return url;
+  }
+  NSArray<NSString *> *names = [headers.allKeys sortedArrayUsingSelector:@selector(compare:)];
+  NSMutableString *key = [NSMutableString stringWithString:url];
+  for (NSString *name in names) {
+    [key appendFormat:@"|%@:%@", name, headers[name]];
+  }
+  return [key copy];
+}
```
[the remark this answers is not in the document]
eszlamczyk: First part is irrelavant as the Stuff already lies in the prop memory, second part addressed

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/styles/StyleConfig.kt
```diff
@@ -166,6 +166,8 @@ class StyleConfig(
     ImageStyle.fromReadableMap(map, styleParser)
   }
 
+  var imageRequestHeaders: Map<String, String> = emptyMap()
```
hryhoriiK97, line 169: Noticed that `imageRequestHeaders` is on `StyleConfig` but not included in `equals()` / `hashCode()`. Right now everything works because both sides of any comparison always have the same headers set, but technically two configs with different headers would look "equal" — could trip someone up down the road if they rely on that check. Quick fix would be to add it to both:

```kotlin
// equals():
imageRequestHeaders == other.imageRequestHeaders

// hashCode():
result = 31 * result + imageRequestHeaders.hashCode()
```

packages/react-native-enriched-markdown/ios/utils/ENRMImageDownloader.m
```diff
@@ -13,6 +14,26 @@ static inline NSUInteger ENRMImageByteCost(RCTUIImage *image)
...
 
+NSString *ENRMImageCacheKey(NSString *url, NSDictionary<NSString *, NSString *> *headers)
+{
+  if (headers.count == 0) {
+    return url;
+  }
+  NSArray<NSString *> *names = [headers.allKeys sortedArrayUsingSelector:@selector(compare:)];
+  NSMutableString *joined = [NSMutableString string];
+  for (NSString *name in names) {
+    [joined appendFormat:@"%@:%@\n", name, headers[name]];
```
hryhoriiK97: Small thing - iOS and Android build the hash input string slightly differently. iOS adds a trailing `\n` after each pair, Android doesn't:

```objc
// iOS → "A:1\nB:2\n"
[joined appendFormat:@"%@:%@\n", name, headers[name]];
```
```kotlin
// Android → "A:1\nB:2"
.joinToString(separator = "\n") { "${it.key}:${it.value}" }
```

Doesn't break anything since caches are per-platform, just one of those things that's nice to keep in sync.

hryhoriiK97 (approved):

PR #545, opened 2026-07-17
feat(input): rebased version of "list state, toolbar, docs, and e2e (lists 5/5)"

hsource:
### What/Why?

This is a stub PR consisting of the rebased changes from #514 by @wildseansy. I created this because we can't wait to have this PR be merged in!

The commits consist of purely the following PRs, all by @wildseansy. I claim no credit for the code.

1. #510 
2. #511 
3. #512 
4. #513 
5. #514 

When rebasing, I resolved these conflicts manually:

- 3/5: Added renamed pruneOrphanedBlockAnchors to new adjustStoresForEditAtLocation function
- 4/5:
  - Update updatePlaceholderVisibility so it checks if any block is active, not just headings (the heading check was added recently to main)
  - Update toggleBlockType to include both the new clearListParagraphStyleFromTypingAttributes and updatePlaceholderVisibility calls (a call to updatePlaceholderVisibility was added recently)

### Testing

**Automated testing:**

@wildseansy has already added 

**Manual testing:**

Per https://github.com/software-mansion/react-native-enriched-markdown/blob/main/CONTRIBUTING.md

1. Run `yarn` followed by `yarn prepare`
2. Run `yarn react-native-example android` and `yarn react-native-example ios` to run on devices

**iOS demo gif:**

List demo

**Android demo video:**

https://github.com/user-attachments/assets/2f70c845-f295-4b11-81e1-0b2a5ac92467

### PR Checklist

- [X] Code compiles and runs on iOS
- [X] Code compiles and runs on Android
- [X] Updated documentation/README if applicable
- [X] Ran example app to verify changes
- [X] E2E tests are passing
- [X] Required E2E tests have been added (if applicable)


packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
```diff
@@ -193,6 +197,49 @@ - (void)normalizeToLineBoundsInText:(NSString *)text
   }
 
   ENRMRemoveIndexesInReverse(_ranges, indexesToRemove);
+  [self recomputeListMetadata];
+}
+
+/// Clamps list depths to valid ancestry (an item nests at most one level under
+/// the previous adjacent list item — CommonMark cannot represent orphan nesting)
+/// and renumbers ordered items among their adjacent same-depth, same-type run.
+- (void)recomputeListMetadata
```
hryhoriiK97, line 229: `recomputeListMetadata` ancestry-clamps but never coerces into `[0, MAX_LIST_DEPTH]`. Android's parser/indent already clamp, but iOS `ENRMInputParser` sets list depth from `listContainerStack.size() - 1` with no `kENRMMaxListDepth` cap — deep nests can exceed the counter array (`MAX + 2`) and crash in recompute. Please clamp in the metadata pass and in the iOS parser so both platforms share the same max.

packages/react-native-enriched-markdown/ios/input/ENRMInputLayoutManager.mm
```diff
@@ -1,5 +1,236 @@
...
+                              font:(UIFont *)font
+                             color:(UIColor *)color
+{
+  NSString *label = [NSString stringWithFormat:@"%ld.", (long)MAX(ordinal, (NSInteger)1)];
+  NSDictionary *attrs = @{NSFontAttributeName : font, NSForegroundColorAttributeName : color};
+  CGSize size = [label sizeWithAttributes:attrs];
+  [label drawAtPoint:CGPointMake(markerRight - size.width, baselineY - font.ascender) withAttributes:attrs];
+}
+
+- (void)drawGlyphsForGlyphRange:(NSRange)glyphsToShow atPoint:(CGPoint)origin
```
hryhoriiK97, line 133: `ENRMInputLayoutManager` always offsets markers left from `usedRect.origin.x` (and the empty-editor path from `inset.left`), so they stay on the wrong side when `writingDirection` is RTL. The readonly renderer already handles this (`ListMarkerDrawer` checks `ENRMParagraphIsRTL`, mirrors marker X, flips ordered to `".N"`). Please add the same treatment here. Also, Android input honors `dir` for placement but still draws `"N."` instead of `".N"` — should match readonly `OrderedListSpan`.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -932,12 +956,321 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+    }
+    return YES;
+  }
+  return NO;
+}
+
+/// Reconciles blocks after Return: a non-empty item continues as a new item at
+/// the same depth, an empty item exits the list (both lines revert to plain
+/// paragraphs). Whether a type continues is the handler's continuesOnNewline.
+- (void)reconcileBlockContinuationAfterNewlineAt:(NSUInteger)newlineLocation previousItemWasEmpty:(BOOL)previousWasEmpty
```
hryhoriiK97, line 1210: iOS Enter continuation leaves ordered markers stuck at `1.` — `reconcileBlockContinuationAfterNewlineAt:` calls `setBlockType` on both the original and new paragraph (creating fresh `ENRMBlockRange` with default `ordinal = 1`) **after** `adjustStoresForEditAtLocation:` has already run `normalizeToLineBoundsInText:`. No second normalize follows, so ordinals are never recomputed.

Android doesn't have this problem: `handleNewlineBlockContinuation` only sets the new line's block (the original keeps its correct ordinal), and `normalizeToLineBounds` runs after continuation in the caller.

Simplest fix: add `[_blockStore normalizeToLineBoundsInText:text]` after `reconcileBlockContinuationAfterNewlineAt:` in `handleTextChanged`.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -932,12 +956,321 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+- (void)outdentList
+{
+  [self changeListDepthBy:-1];
+}
+
+/// Adjusts the nesting depth of the cursor's list item by `delta`, clamped to
+/// [0, kENRMMaxListDepth] and preserving the item's list type. QoL on the
+/// boundaries: indent on a non-list paragraph starts a depth-0 bullet (ignored
+/// on headings), and outdent past depth 0 removes the marker.
+- (void)changeListDepthBy:(NSInteger)delta
```
hryhoriiK97, line 1008: `changeListDepthBy:` should indent/outdent each selected paragraph (like Android's `forEachSelectedLine` and like iOS `toggleBlockType`'s paragraph enumeration). Today it calls `setBlockType` once on the full selection range; normalize then clips to the first line and can drop list state on the other selected items.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputEventEmitter.kt
```diff
@@ -27,6 +28,8 @@ class InputEventEmitter(
 ) {
   private var prevState: Map<StyleType, Boolean> = emptyMap()
   private var prevHeadingLevel: Int = 0
+  private var prevUnorderedList: Pair<Boolean, Int> = false to 0
+  private var prevOrderedList: Pair<Boolean, Int> = false to 0
   private var prevCaretRect: CaretRect? = null
 
   fun emitChangeText() {
```
hryhoriiK97, line 35: Strip ZWSP from the `onChangeText` payload (and any plain-text getter) — `MarkdownSerializer` already strips it for the markdown output, but the plain-text path doesn't. Empty-list anchors are an internal editing detail and shouldn't leak to JS.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -932,12 +956,321 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+  if (alreadyActive) {
+    // The just-cleared line is a plain paragraph now; drop any list indent the
+    // typing attributes carried so the next typed character isn't indented.
+    [self clearListParagraphStyleFromTypingAttributes];
+  }
+  [self updateEmptyBulletMarker];
   [self emitFormattingChanged];
 }
 
+- (void)toggleUnorderedList
```
hryhoriiK97, line 984: When toggling unordered ↔ ordered on an existing list item, keep the current `depth` instead of forcing `level: 0`.

docs/INPUT.md
```diff
@@ -85,6 +85,28 @@ Behavior notes:
 
 Heading font sizes, weights, and colors are configurable per level — see [Customizing Styles](#customizing-enrichedmarkdowntextinput--styles).
 
+## Bullet Lists
```
hryhoriiK97: `INPUT.md` / API docs should state that input list items are single-line: loose-list continuation paragraphs are not kept as list-styled blocks on import/parse.

.maestro/enrichedMarkdownInput/flows/basic/block_elements/unordered_list_test.yaml
```diff
@@ -0,0 +1,55 @@
...
+    id: toolbar-indent
+- waitForAnimationToEnd
+
+- tapOn:
+    id: blur-button
+
+- runFlow:
+    file: '../../../subflows/capture_or_assert_screenshot.yaml'
+    env:
+      SCREENSHOT_NAME: 'unordered_list_input'
```
hryhoriiK97, line 55: Please add an Android screenshot baseline for `unordered_list_input`, and cover ordered lists (and ideally outdent / unordered↔ordered toggle) in Maestro — right now only one unordered happy-path exists with an iOS-only baseline.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -1743,21 +2135,44 @@ - (void)handleTextChanged
...
     }
   }
 #endif
 
-  [self applyFormattingScopedToEditAtLocation:editLocation insertedLength:insertedLength];
+  // An edit that adds or removes lines can shift the depth clamp and ordered
+  // numbering of items far below the edited line; re-stamp the whole document
+  // in that case. Same-line typing keeps the scoped per-keystroke path.
+  NSString *postEditText = ENRMGetPlainText(_textView);
+  NSUInteger lineCount = 1;
```
hryhoriiK97: The full-text line-count scan in `handleTextChanged` is O(document length) on every keystroke. Android's `editTouchedNewline()` does the same thing by only looking at the inserted/deleted range — O(edit size). Either do the same on iOS, or cache the count and update with a delta (`+= insertedNewlines - deletedNewlines`).

docs/INPUT.md
```diff
@@ -85,6 +85,28 @@ Behavior notes:
...
+Behavior notes:
+
+- **Return continues the list** — pressing Return inside an item starts a new bullet at the same depth on the next line. Pressing Return on an empty item exits the list (the line becomes a regular paragraph).
+- **Backspace at the start of an item** outdents it; at depth 0 it removes the bullet.
+- An emptied list line stays a list item (its bullet keeps showing) until you toggle it off or Backspace out of it, so text you type next continues in the list.
+- The bullet marker exists only in the serialized Markdown and the rendered glyph — it is never part of the editor's text, so it won't collide with `-`-prefixed text such as mentions.
+
+Whether the cursor's paragraph is a list item, and its nesting depth, are reported through the [`onChangeState`](API_REFERENCE.md#onchangestate) payload as `unorderedList` / `orderedList` (`{ isActive, depth }`), so you can highlight the list buttons in your toolbar.
+
+Use the [`listItemSpacing`](API_REFERENCE.md#listitemspacing) prop to add vertical spacing between items so they read as separate rows.
```
hryhoriiK97: `listItemSpacing` docs should say list items (not only bullets) — ordered handlers use the same spacing. Optionally rename INPUT `## Bullet Lists` to cover both types.

packages/react-native-enriched-markdown/ios/input/styles/ENRMUnorderedListBlockHandler.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#import "ENRMBlockHandler.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Block handler for unordered (bullet) list items. Reserves the head-indent
+/// column for the marker via the paragraph style (the layout manager draws the
+/// glyph into it) and serializes to a `- ` line prefix indented two spaces per
```
hryhoriiK97: Nit: header says "indented two spaces per nesting level" — code uses three (`depth * 3`). Three is correct; please fix the comment.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -72,14 +72,26 @@ @implementation EnrichedMarkdownTextInput {
   ENRMPlaceholderLabel *_placeholderLabel;
 
   NSUInteger _lastTextLength;
+  NSUInteger _lastLineCount;
```
hryhoriiK97: Nit: `_lastLineCount` is never initialized after `importMarkdown:` or `pasteMarkdown:` (compare `_lastTextLength` which is). The first keystroke after loading a multi-line document triggers an unnecessary full-document reformat because `lineCount != 0`. Please initialize `_lastLineCount` alongside `_lastTextLength` in those paths.

hryhoriiK97 (commented):
@hsource thanks again for the PR! I've left a few comments – can you address them, please?

packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1) (targets this PR's branch). `ENRMInputParser` now caps the item depth at `kENRMMaxListDepth` like Android's parser ([ENRMInputParser.mm#L196-L201](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/ENRMInputParser.mm#L196-L201)), and `recomputeListMetadata` coerces levels into `[0, kENRMMaxListDepth]` before the ancestry clamp, so the counter arrays can never be over-indexed even if a bad level slips in from elsewhere ([ENRMBlockStore.mm#L225-L234](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm#L225-L234)).

packages/react-native-enriched-markdown/ios/input/ENRMInputLayoutManager.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1). `ENRMInputLayoutManager` now resolves each list line's direction from its paragraph style via `ENRMParagraphIsRTL` (the same helper the readonly `ListMarkerDrawer` uses), mirrors the marker to the trailing edge, and flips ordered labels to `".N"` ([ENRMInputLayoutManager.mm#L189-L222](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/ENRMInputLayoutManager.mm#L189-L222)). Empty list lines have no character to read the style from, so the orchestrator resolves their direction from the writing-direction mode and hands it over via a new `emptyBulletRTL` property ([EnrichedMarkdownTextInput.mm#L1322-L1338](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L1322-L1338)). On Android, `InputOrderedListMarkerSpan` now draws `".N"` left-aligned when `dir < 0`, matching the readonly `OrderedListSpan` ([InputOrderedListMarkerSpan.kt#L47-L53](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/spans/InputOrderedListMarkerSpan.kt#L47-L53)).

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1) with the suggested approach: `handleTextChanged` re-runs `normalizeToLineBoundsInText:` immediately after `reconcileBlockContinuationAfterNewlineAt:`, so the re-seeded ranges (default `ordinal = 1`) get renumbered by the metadata pass ([EnrichedMarkdownTextInput.mm#L2209-L2215](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L2209-L2215)).

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): `changeListDepthBy:` now enumerates the selected paragraphs and sets one range per line — each clamped from its own depth — mirroring Android's `forEachSelectedLine` and `toggleBlockType`'s paragraph enumeration ([EnrichedMarkdownTextInput.mm#L1012-L1057](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L1012-L1057)).

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputEventEmitter.kt
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): `emitChangeText` now strips the ZWSP anchor from the `onChangeText` payload ([InputEventEmitter.kt#L37-L43](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputEventEmitter.kt#L37-L43)). That's the only path plain text crosses to JS on Android; the markdown payloads were already stripped by `MarkdownSerializer`.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1) on both platforms: when the new type and the line's existing block are both list items, the existing depth is reused (per line, so mixed-depth selections keep their shape). iOS `toggleBlockType:` ([EnrichedMarkdownTextInput.mm#L945-L975](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L945-L975)), Android `toggleListType` ([EnrichedMarkdownTextInputView.kt#L690-L710](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt#L690-L710)).

docs/INPUT.md
wildseansy: Added in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): INPUT.md's behavior notes now state that list items are single-line and that loose-list continuation paragraphs are not kept as list-styled blocks on import/paste ([INPUT.md#L105](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/docs/INPUT.md#L105)); the same note is on `toggleUnorderedList()` in API_REFERENCE.md ([API_REFERENCE.md#L937](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/docs/API_REFERENCE.md#L937)).

.maestro/enrichedMarkdownInput/flows/basic/block_elements/unordered_list_test.yaml
wildseansy: Partially addressed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): added two smoke-tagged input flows — [ordered_list_test.yaml](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/.maestro/enrichedMarkdownInput/flows/basic/block_elements/ordered_list_test.yaml) (ordered items, Return continuation, indent restarts numbering, outdent renumbers back to 1./2./3.) and [list_toggle_test.yaml](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/.maestro/enrichedMarkdownInput/flows/basic/block_elements/list_toggle_test.yaml) (unordered → indent → toggle to ordered, depth kept). I couldn't run the example apps in my environment, so three baselines still need an `UPDATE_SCREENSHOTS=true` capture run: the Android `unordered_list_input` you asked for, plus both platforms for the two new flows — called out in the fix PR description.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1) using the `editTouchedNewline` approach: the inserted run is scanned in the post-edit text, and the replaced run is checked in `shouldChangeTextInRange` before the deleted characters disappear — both O(edit size) ([EnrichedMarkdownTextInput.mm#L2228-L2254](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L2228-L2254), capture at [#L2356](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L2356) and [#L2468 (macOS)](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm#L2468)). `_lastLineCount` is removed entirely.

docs/INPUT.md
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): `listItemSpacing` docs now read "each list item (bullet and numbered alike)" in both API_REFERENCE.md and INPUT.md, and the INPUT.md section is renamed to "Lists".

packages/react-native-enriched-markdown/ios/input/styles/ENRMUnorderedListBlockHandler.h
wildseansy: Fixed in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1) — the header comment now says three spaces ([ENRMUnorderedListBlockHandler.h#L9](https://github.com/wildseansy/react-native-enriched-markdown/blob/b3cd94ef210697c3fa46383601a8a0f8c95db7e1/packages/react-native-enriched-markdown/ios/input/styles/ENRMUnorderedListBlockHandler.h#L9)).

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Resolved as a side effect of the line-count perf fix in [wanderlog#1](https://github.com/wanderlog/react-native-enriched-markdown/pull/1): `_lastLineCount` no longer exists (full-vs-scoped reformat is decided per edit from the inserted/replaced runs), so there's no cached count left to initialize or go stale after `importMarkdown:`/`pasteMarkdown:`.

wildseansy:
@hsource - I addressed the comments in a PR to your branch

hsource:
I merged it in! @hryhoriiK97 - take a look
@wildseansy - I also added you as a maintainer so you can freely push to the branch

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextInputNativeComponent.ts
```diff
@@ -272,6 +276,14 @@ export interface NativeProps extends ViewProps {
    */
   writingDirection?: CodegenTypes.WithDefault<string, 'first-strong'>;
 
+  /**
+   * Vertical spacing (points) added above each bullet list item so items read as
+   * separate rows. iOS applies it via paragraphSpacingBefore; Android via a
+   * LineHeightSpan.
+   * @default 0
+   */
+  listItemSpacing?: CodegenTypes.WithDefault<CodegenTypes.Int32, 0>;
```
hryhoriiK97: This should be part of `markdownStyle`, not a separate prop. All other visual properties (colors, fonts, heading styles, spoiler styles) go through `markdownStyle` — having one list spacing property break that pattern will confuse consumers. Something like `markdownStyle.list.itemSpacing` would be consistent. That would also eliminate the `applyComposedStyle()` workaround on Android.

hryhoriiK97: Should be `Float`, not `Int32`. All other spacing values in the library (`marginTop`, `marginBottom`, `gapWidth`, `padding`, etc.) are floats. This prevents sub-point values like `4.5` and is inconsistent with everything else.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/model/InputFormatterStyle.kt
```diff
@@ -11,6 +11,10 @@ data class InputFormatterStyle(
   val spoilerBackgroundColor: Int,
   /** Per-level heading styling, indexed 0..5 for H1..H6. Always length 6. */
   val headings: List<InputHeadingStyle>,
+  /** Display density (px per dp) so block handlers can build density-correct spans without a Context. */
+  val displayDensity: Float = 1f,
+  /** Extra vertical spacing (px) above each bullet item, from the `listItemSpacing` prop. */
+  val listItemSpacingPx: Int = 0,
```
hryhoriiK97: Nit: these two don't really belong here — `InputFormatterStyle` is meant to be a pure style data class. `displayDensity` is a platform constant that could be passed to the `BlockHandler` at construction time or resolved from `Context` at the span-creation site. `listItemSpacingPx` goes away if `listItemSpacing` moves into `markdownStyle`.

docs/API_REFERENCE.md
```diff
@@ -629,6 +637,8 @@ interface StyleState {
   link: { isActive: boolean };
   // Heading level of the cursor's paragraph: 0 = none, 1-6 = H1-H6.
   heading: { isActive: boolean; level: number };
+  unorderedList: { isActive: boolean; depth: number };
+  orderedList: { isActive: boolean; depth: number };
```
hryhoriiK97, line 638: Worth documenting that `depth` is always `0` when `isActive` is `false`. The existing `heading.level` works differently (level 0 = no heading), so without a note here people will wonder about the semantics.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -301,6 +377,12 @@ class EnrichedMarkdownTextInputView(
       }
     }
 
+    if (!isTextChanging && !isProcessingTextChange) {
+      // The caret moving on/off an empty bullet line toggles the ZWSP anchor and the
+      // placeholder visibility; skip during a text-change pass (handled there).
+      syncEmptyListAnchor()
```
hryhoriiK97, line 389: `syncEmptyListAnchor()` scans the entire document backwards on every selection change. For a short document that's fine, but on a large document with frequent caret moves this is O(n) per event. Could be worth short-circuiting early — e.g. skip the full scan when the document has no list blocks at all, or track ZWSP positions in a set instead of scanning.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
```diff
@@ -37,32 +43,38 @@ object MarkdownSerializer {
...
+          // underline for the previous line; emit an empty list line bare. An
+          // empty "# " heading is valid ATX and keeps its prefix.
+          if (isListItem && plainLines[lineIndex].replace(ZWSP, "").isEmpty()) continue
           markdownLines[lineIndex] = prefix + markdownLines[lineIndex]
         }
       }
     }
 
-    return markdownLines.joinToString("\n")
+    return markdownLines.joinToString("\n").replace(ZWSP, "")
```
hryhoriiK97: The ZWSP anchor is stripped in 4 separate places: `emitChangeText`, two paths in `serialize`, and the empty-item skip on line 71. If a future code path emits text without stripping, the ZWSP leaks to JS. A single `stripZWSP()` call at the serialization boundary (or a shared helper all output paths funnel through) would be less fragile than scattering the strips.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -925,19 +957,386 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+/// The list block owning the caret's paragraph (matched by line start, so empty
+/// anchors and line-end carets register), or nil.
+- (nullable ENRMBlockRange *)listBlockForCursorParagraph
+{
+  return [self listBlockForParagraphAtPosition:_textView.selectedRange.location];
+}
+
+- (nullable ENRMBlockRange *)listBlockForParagraphAtPosition:(NSUInteger)position
+{
+  NSString *text = _textView.textStorage.string;
```
hryhoriiK97, line 1075: This method linearly scans `_blockStore.allRanges` on every call — and it gets called from `emitOnChangeState`, `updateEmptyBulletMarker`, `syncTypingAttributesWithCursorBlock`, `handleTextChanged`, `capturePreEditBlockForRange`, `changeListDepthBy`, `handleListKeyForReplacementRange`, etc. Same pattern on Android with `blockStore.allRanges.firstOrNull { ... }` in 5+ places. A `lineStart → BlockRange` lookup map in `BlockStore` would make all of these O(1) and be a single-point change.

packages/react-native-enriched-markdown/ios/input/ENRMInputLayoutManager.mm
```diff
@@ -1,5 +1,309 @@
...
+/// indent from the trailing (right) edge, so the marker mirrors past it.
+/// `leadingOffset` is the same padding+indent that positions the LTR column.
+static CGFloat ENRMTrailingMarkerX(CGPoint origin, NSTextContainer *container, CGFloat leadingOffset)
+{
+  return origin.x + container.size.width - leadingOffset;
+}
+
+- (void)drawGlyphsForGlyphRange:(NSRange)glyphsToShow atPoint:(CGPoint)origin
+{
+  [super drawGlyphsForGlyphRange:glyphsToShow atPoint:origin];
```
hryhoriiK97, line 135: This entire method is a single 140+ line `enumerateLineFragmentsForGlyphRange:usingBlock:` callback doing paragraph lookup, attribute reads, RTL detection, baseline math, and drawing. It also allocates `NSMutableSet<NSNumber *>` and boxes `NSNumber` objects on every draw pass. Worth splitting into helpers (`drawBulletMarker:…`, `drawOrderedMarker:…`, `resolveMarkerOrigin:…`) for maintainability, and the per-frame allocations are worth watching for scroll perf in long documents.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -925,19 +957,386 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+  }
+  return match;
+}
+
+/// Records the block kind/level of the line being edited before the change, so a
+/// Return that continues a list can recover the previous item's depth and an
+/// in-line replacement (autocorrect/paste) that wipes the line can heal it.
+- (void)capturePreEditBlockForRange:(NSRange)range
+{
+  _preEditBlockType = ENRMInputBlockTypeParagraph;
```
hryhoriiK97: `_preEditBlockType` / `_preEditBlockLevel` are set here as instance variables and consumed later by `reconcileBlockContinuationAfterNewlineAt:`. If two rapid edits overlap (fast typing + autocorrect, or programmatic `setValue` during an edit), the pre-edit state may belong to the wrong edit. Android handles this inline in `handleNewlineBlockContinuation` using the post-edit text, which is safer — no window for staleness.

packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
```diff
@@ -45,6 +45,7 @@ - (void)setRanges:(NSArray<ENRMBlockRange *> *)ranges
       return NSOrderedDescending;
     return NSOrderedSame;
   }] mutableCopy];
+  [self recomputeListMetadata];
 }
```
hryhoriiK97, line 68: The "one block per paragraph" invariant relies entirely on `removeBlocksOverlapping` being called inside `setBlock`. There is no assertion or validation — if a code path mutates ranges without going through the standard toggle path, a line could end up with both a heading and a list item. An `NSAssert` / `check()` in `normalizeToLineBounds` that no two ranges share the same paragraph start would catch this early.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
```diff
@@ -925,19 +957,386 @@ - (void)toggleBlockType:(ENRMInputBlockType)type level:(NSInteger)level
...
+  }
+  return NO;
+}
+
+/// Reconciles blocks after Return: a non-empty item continues as a new item at
+/// the same depth, an empty item exits the list (both lines revert to plain
+/// paragraphs). Whether a type continues is the handler's continuesOnNewline.
+- (void)reconcileBlockContinuationAfterNewlineAt:(NSUInteger)newlineLocation previousItemWasEmpty:(BOOL)previousWasEmpty
+{
+  id<ENRMBlockHandler> handler = [_formatter handlerForBlockType:_preEditBlockType];
```
hryhoriiK97, line 1212: When exiting a list (Return on empty item), this method directly strips `NSParagraphStyleAttributeName` via `beginEditing`/`removeAttribute`/`endEditing`. The formatter normally owns paragraph styles — it strips and applies them during `applyBlockRanges:`. This manual strip is a parallel path the formatter does not know about. If the formatter later expects its last-stamped ranges to still be present, it could leave stale styles or re-apply something that was just removed.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -256,9 +309,30 @@ class EnrichedMarkdownTextInputView(
 
     isProcessingTextChange = true
     try {
-      adjustStoresForEdit(editStart, deletedLength, insertedLength)
+      formattingStore.adjustForEdit(editStart, deletedLength, insertedLength)
+      blockStore.adjustForEdit(editStart, deletedLength, insertedLength)
+      pruneOrphanedAnchors()
+      handleNewlineBlockContinuation(editStart, deletedLength, insertedLength)
+      text?.let { blockStore.normalizeToLineBounds(it) }
       applyPendingStyles(editStart, insertedLength)
```
hryhoriiK97, line 323: This runs inside `onAfterTextChanged` and calls `runAsATransaction { editable.delete(...) }` — deleting characters, adjusting the block store, and moving the selection, all inside the callback responding to a text change. The `isProcessingTextChange` guard prevents infinite recursion, but modifying the `Editable` inside `afterTextChanged` is a known Android pitfall — it can interact badly with IME composition (Chinese/Japanese input), undo/redo, and accessibility services.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -512,6 +664,187 @@ class EnrichedMarkdownTextInputView(
...
+    val editable = text ?: return
+    val turningOff = listBlockAtCursor()?.type == type
+    if (turningOff) {
+      forEachSelectedLine { ls, le -> blockStore.removeBlock(ls, le, editable) }
+    } else {
+      forEachSelectedLine { ls, le ->
+        // Switching unordered <-> ordered keeps each line's nesting depth.
+        val existing = blockStore.allRanges.firstOrNull { it.start == ls && it.type in BlockType.LIST_ITEMS }
+        blockStore.setBlock(type, existing?.level ?: 0, ls, le, editable)
+      }
```
hryhoriiK97, line 706: This `blockStore.allRanges.firstOrNull { it.start == ls && it.type in BlockType.LIST_ITEMS }` pattern appears verbatim in at least 5 places: `toggleListType`, `changeListDepthBy`, `handleListKey`, `syncEmptyListAnchor` (twice). Each is a linear scan. iOS has `listBlockForParagraphAtPosition:` as a single method. Android should have the same — reduces duplication and makes a future O(1) lookup a one-line change.

hryhoriiK97 (commented):
@hsource @wildseansy thank you! I've left a few additional comments.

packages/react-native-enriched-markdown/src/EnrichedMarkdownTextInputNativeComponent.ts
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — `listItemSpacing` moved into `markdownStyle.list.itemSpacing`, consistent with the other visual props. On Android this removed the `applyComposedStyle()` workaround entirely.

wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — it's now `CodegenTypes.Float` (Android carries it as a dp `Float`, iOS as `CGFloat`).

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/model/InputFormatterStyle.kt
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — `InputFormatterStyle` is now pure style data: `displayDensity` moved to an `InputFormatter` constructor arg passed to the list handlers, and `listItemSpacingPx` is gone (spacing is carried as a dp `Float` and converted to px at span creation).

docs/API_REFERENCE.md
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — `API_REFERENCE.md` now documents that list `depth` is always `0` when `isActive` is false, contrasting with `heading.level` where `0` is itself meaningful.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — added an anchor count that is made exact on every strip scan (it recounts the anchors it keeps), so the O(document) backward scan is skipped whenever no empty-list anchors are live. Incoming markdown is also scrubbed of the reserved U+200B at parse time so foreign copies can't seed a stray anchor.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/MarkdownSerializer.kt
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — all output now funnels through a single `MarkdownSerializer.stripZwsp` choke point (the block-aware serialize strips once at its return; `InputEventEmitter` and the context-menu `selectedText` payload use the same helper). Incoming markdown is scrubbed at parse time as well.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — added `BlockStore.blockStartingAt` on both platforms (O(log n) binary search over the start-sorted ranges) and routed the paragraph-start scans through it, behind shared helpers (`blockForParagraphAtPosition:` on iOS, `listBlockStartingAt` on Android). The per-call linear `allRanges` scans in the sites you listed are gone.

packages/react-native-enriched-markdown/ios/input/ENRMInputLayoutManager.mm
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — extracted a shared `drawListMarkerOrdered:...` helper used by both the in-text fragment pass and the trailing empty-line pass (the RTL/ordered placement math now lives in one place), and hoisted the per-draw `NSMutableSet` to a reused ivar cleared at the top of each pass, so the scroll draw path allocates nothing.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Addressed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2. Full inlining à la Android isn't safe here: the exit case (Return on an empty item) needs the pre-edit block type, but the empty item's zero-length anchor is consumed by the inserted newline, so it can't be recovered from the post-edit store. So instead: the pre-edit state is cleared immediately after the newline-continuation pass consumes it (a later edit that reaches reconciliation without a fresh capture can no longer act on stale state), and the macOS `textInputShouldChangeText:inRange:` delegate now captures it too — it previously carried stale values into reconciliation. The continue case recovers type/level through the shared post-edit lookup.

packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
wildseansy: Investigated this but didn't add the assert. `normalizeToLineBounds`' duplicate-drop is a legitimate self-heal path that ordinary edits hit — e.g. backspace-merging two heading lines briefly leaves a heading anchor and the merged block sharing a start, which normalize correctly drops. An assert there false-positives (debug crash on a normal 3-action edit) and can't distinguish that benign merge from genuine corruption without more state than is available at that point. The one-block-per-paragraph invariant is instead maintained structurally by `removeBlocksOverlapping`/`setBlock` and this dedup itself; left a comment noting the drop is deliberately silent.

packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
wildseansy: Kept the manual strip here (with a clarifying comment) rather than routing through the formatter. Empty list lines never carry a stamped `ENRMBlockTypeAttributeName` (zero-length ranges aren't stamped — the marker is drawn by the layout manager), so the formatter's marker-keyed reset can't reach their paragraph style; the list style rides in via the newline's typing attributes. The manual strip is the only thing that clears the leftover indent/spacing on exit and on toggling-off an empty item — removing it leaves the exited line permanently indented (verified this regresses). So this is deliberate; documented why in https://github.com/wanderlog/react-native-enriched-markdown/pull/2.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
wildseansy: Documented in https://github.com/wanderlog/react-native-enriched-markdown/pull/2. The mutation is deliberately synchronous and guarded: every buffer change is wrapped in `runAsATransaction` (sets `isDuringTransaction`, so `MarkdownTextWatcher` early-returns and the edit doesn't re-enter the pass), the whole `onAfterTextChanged` body holds `isProcessingTextChange`, and `isManagingAnchor` blocks re-entry via the selection path. It has to stay synchronous — deferring the insert/delete would run it outside every guard and break the single-stamp ordering, the caret placement past the anchor, and `lastProcessedText`, reintroducing the double-bullet and stale-caret bugs. Added a comment on `syncEmptyListAnchor` spelling this out.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
wildseansy: Fixed in https://github.com/wanderlog/react-native-enriched-markdown/pull/2 — added `listBlockStartingAt(ls)`, and `blockOnParagraphAt` now routes through `BlockStore.blockStartingAt` (O(log n)), replacing the repeated linear `allRanges.firstOrNull { it.start == ls }` scans in `toggleListType`, `changeListDepthBy`, `syncEmptyListAnchor` (both), and `handleNewlineBlockContinuation`.

wildseansy:
@hryhoriiK97 addressed the second round in [wanderlog#2](https://github.com/wanderlog/react-native-enriched-markdown/pull/2) (targets this branch, same as before). Replied inline on each thread. Quick summary: the `markdownStyle.list.itemSpacing` move, the `InputFormatterStyle` cleanup, the docs, the O(n) anchor scan, the ZWSP consolidation, the shared `blockStartingAt` lookup (both platforms), the layout-method split, and the pre-edit/macOS hardening are all fixed in code. Two I pushed back on with reasoning rather than changing: the `normalizeToLineBounds` assertion (its duplicate-drop is a legitimate self-heal that ordinary edits hit, so it false-positives) and the list-exit paragraph-style strip (empty list lines carry no stamped block marker, so the formatter's reset can't reach them — the manual strip is load-bearing). Verified on the iOS simulator; couldn't run the Android/macOS builds or the Maestro suite in my env.

wildseansy:
@hsource - I wasn't able to submit changes directly to `wanderlog:feat/list-5-state` branch.  Not sure if permissions are configured correctly.  Instead submitted PR to that branch

hsource:
Weird! Sorry about that and thanks for the fixes. @wildseansy  It looks like the invitation is pending, so you may need to accept it to directly edit: https://github.com/wanderlog/react-native-enriched-markdown/invitations

@hryhoriiK97 I've merged in Sean's latest changes!

wildseansy:
@hsource - Ah that was definitely the issue.  I didn't accept. Thanks for the tip

packages/react-native-enriched-markdown/ios/input/ENRMFormattingStore.mm
```diff
@@ -218,6 +218,31 @@ - (void)adjustForEditAtLocation:(NSUInteger)editLocation
...
   }
+
+  [self coalesceAdjacentSameTypeRanges];
+}
+
+/// Merge same-type (and same-url) ranges left adjacent or overlapping by an
+/// edit — e.g. deleting the space in "**foo** **bar**" leaves two touching
+/// bold ranges that would serialize as "**foo****bar**". `addRange:` keeps
+/// this invariant on insert; the edit path must too.
+- (void)coalesceAdjacentSameTypeRanges
```
wildseansy, line 229: This was a bug I found testing.  Thought it was low enough effort/risk to just throw in. Happy to extract to another PR if needed

hryhoriiK97 (approved):

PR #582, opened 2026-07-22
fix(iOS): main-thread deadlock in markdown measurement under locked ShadowTree commits

### LINKED ISSUE #550: iOS device-only deadlock (watchdog 0x8BADF00D): measureContent dispatch_sync to main vs Reanimated ShadowTree commit

- [x] I tested using the latest version of the library, as the bug might be already fixed. *(The affected code path in `ios/internals/ShadowMeasurementUtils.h` is unchanged on current `main` / 0.8 nightlies.)*
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers. *(Searched for deadlock / freeze / `0x8BADF00D` / `measureContent` — no existing issue covers this.)*

### Bug summary

On **physical iPhones**, screens that mount `EnrichedMarkdownText` while a Reanimated animation is running can hard-freeze the app; iOS's watchdog then kills it ~1 minute later with `0x8BADF00D` ("app is stuck / deadlock"). We captured **19 watchdog kills in a single day** on our chat screen (long markdown history + always-on shimmer/keyboard animations). The **iOS Simulator never reproduces it**, and since first filing we've root-caused the exact interleaving — the deadlock additionally requires RN's experimental `preventShadowTreeCommitExhaustion` flag (details in Root cause; even without it, the same `dispatch_sync` design causes heavy main-thread jank during streaming).

Every `.ips` thread dump shows the same two-thread cycle:

**JS/layout thread** — holds the ShadowTree revision mutex (via RN's commit-exhaustion locked fallback, see below), inside Yoga layout:

```
__ulock_wait → _dispatch_sync_f_slow → __DISPATCH_WAIT_FOR_QUEUE__
facebook::react::ENRMFontScaleForMeasurement(bool)          ← dispatch_sync to main
facebook::react::ENRMMeasureMarkdownContent<…>
facebook::react::EnrichedMarkdownShadowNode::measureContent
facebook::yoga::calculateLayout … RootShadowNode::layoutIfNeeded
facebook::react::ShadowTree::tryCommit
facebook::react::ShadowTree::commit             ← exhaustion wrapper, mutex held
facebook::react::UIManager::completeSurface     ← ordinary React commit
```

**Main thread** — Reanimated committing its animation batch:

```
__psynch_mutexwait → std::recursive_mutex::lock
facebook::react::ShadowTree::sharedRevisionLock
facebook::react::ShadowTree::tryCommit
reanimated::ReanimatedModuleProxy::commitUpdates → performOperations
-[REANodesManager maybeFlushUIUpdatesQueue]
```

Classic lock-order inversion:

- layout thread: **revision mutex → main queue** (`dispatch_sync`)
- main thread: **main queue (its execution context) → revision mutex** (`sharedRevisionLock`)

#### Root cause

`ios/internals/ShadowMeasurementUtils.h` documents its own safety assumption:

```objc
// measureContent runs on Yoga's background layout thread and uses
// dispatch_sync to main for UIKit reads. Safe because RN never
// synchronously joins the layout queue from main. A synchronous
// layout flush from main would deadlock.
```

The assumption doesn't hold, and we've now root-caused the exact path (symbolicated `.ips` reports + an instrumented RN-from-source build). In stock RN, `ShadowTree::tryCommit` runs layout *without* holding the revision mutex — but RN does not guarantee that, and today's experimental flag set breaks it: with `preventShadowTreeCommitExhaustion` enabled (`releaseLevel = Experimental`), a commit that fails **3 consecutive optimistic attempts** is re-run in a fallback that holds `revisionMutexRecursive_` across the ENTIRE `tryCommit` — **layout included** (`ShadowTree.cpp`, `MAX_COMMIT_ATTEMPTS_BEFORE_LOCKING`).

That's precisely what our production traces show: chat streams markdown (slow measured commits from the JS thread) while Reanimated commits from the main thread every frame (we run `DISABLE_COMMIT_PAUSING_MECHANISM`, so it never yields to in-flight React commits). Each main-thread commit bumps the revision and fails the JS thread's slow in-flight attempt; after three losses the JS thread enters the locked fallback, the next Reanimated commit blocks main on `sharedRevisionLock`, and the next `dispatch_sync(main)` from `measureContent` completes the deadlock. The main thread never *joins the layout queue* — it just blocks on a mutex that layout is running under.

So the exposure today is: RN experimental release level + a main-thread ShadowTree committer (Reanimated with commit pausing disabled makes it near-certain; ours logged 19 kills/day). But the contract violation is the library's: `measureContent` must not block on the main thread, because RN is free to hold locks around layout — the experimental flag does exactly that, and experimental flags routinely graduate to stable defaults. Separately, even where the deadlock can't form, the `dispatch_sync` design serializes **all** markdown measurement onto the main thread, which visibly janks animations and scroll during streaming.

`ENRMMeasureMarkdownContent` hops synchronously to main in **three** places, so no prop combination avoids it:

1. `ENRMFontScaleForMeasurement` — runs on **every** non-streaming measure, *before* the measurement cache is consulted, so even fully cached content deadlocks. `allowFontScaling={false}` skips this one but not #3.
2. The streaming current-size read (`view.bounds` under `streamingAnimation`).
3. The measure block itself (rendered-view or mock-view measurement) on every cache miss.

#### Why it's device-only in practice

Two compounding timing effects, both proportional to measurement cost: (1) a commit only *fails* when a main-thread commit lands during it — on device a streaming/history commit spends **30–80 ms per measured node** (we instrumented it), so losing 3 races in a row is routine, while on an M-series simulator measures take microseconds and the fallback never engages; (2) once in the locked fallback, the deadlock window is the `dispatch_sync` traffic during the locked layout. Simulator-based QA will not catch this.

#### Suggested direction

`measureContent` must never `dispatch_sync` to the main thread while holding the commit lock:

- Font scale: cache `RCTFontSizeMultiplier()` in an atomic, updated on main via `UIContentSizeCategoryDidChangeNotification`, and read the atomic from the layout thread.
- Measurement: use thread-confined TextKit (`NSLayoutManager`/`NSTextStorage` created on the layout thread) instead of `dispatch_sync`ing to a main-thread mock `UIView`, mirroring how RN core's `ParagraphShadowNode` measures text off-main.

### Library version

0.7.4 (affected code identical on `main`)

### Environment info

```shell
Library: react-native-enriched-markdown 0.7.4
React Native: 0.86.0 (New Architecture / Fabric, releaseLevel: experimental
  → preventShadowTreeCommitExhaustion enabled)
React: 19.2.3
Expo SDK: 57
react-native-reanimated: 4.5.0 (staticFeatureFlags: DISABLE_COMMIT_PAUSING_MECHANISM)
Platform: iOS — physical devices only; iOS Simulator does not reproduce (see timing note)
Build tooling: Xcode 26.6 (17F113), macOS 26.5.2, Node 24.12.0, Bun 1.3.14
```

### Steps to reproduce

1. Use a **physical iPhone** (the simulator will not reproduce; measures are too fast there for commits to fail).
2. Run RN at `releaseLevel: experimental` (→ `preventShadowTreeCommitExhaustion`) and Reanimated with `DISABLE_COMMIT_PAUSING_MECHANISM` — our production configuration, and the preconditions for layout ever holding the revision mutex.
3. Have a continuously running Reanimated animation on screen (e.g. a looped shimmer), i.e. per-frame main-thread ShadowTree commits.
4. Stream markdown into mounted `EnrichedMarkdownText` components while mounting more in bursts (chat streaming + history/pagination). Each slow measured commit that loses 3 revision races enters the locked fallback; instrumentation shows these engaging within minutes (see the repro repo's INSTRUMENTATION.md).
5. Each locked commit that re-measures dirty text can deadlock; when it does, the UI hard-freezes and ~60 s later iOS's watchdog terminates the app with `0x8BADF00D`. The `.ips` report shows the two stacks quoted above (ours did, 19× in one day).

https://github.com/slavaluka/enriched-markdown-deadlock-repro

Bare RN 0.86 app pinning `react-native-enriched-markdown@0.7.4` + `reanimated@4.5.0`, mirroring our production configuration (`releaseLevel: Experimental`, `DISABLE_COMMIT_PAUSING_MECHANISM`) and workload (history mega-mount, 5 streaming cards @30 ms, pagination bursts, a per-frame layout-prop animation). Its INSTRUMENTATION.md documents three small probes (RN core via `RCT_USE_PREBUILT_RNCORE=0`, Reanimated, this library) that show the whole chain live on device: a constant stream of `commit FAILED` lines and `LOCKED FALLBACK ENGAGED` within minutes — each engagement that re-measures dirty text is a deadlock roll. The freeze itself is probabilistic per engagement (heavier production trees lose the roll daily; the minimal app needs longer runs), but every structural precondition is observable on demand, and the production `.ips` reports pin the fatal interleaving exactly.

---

🤖 Filed with [Claude Code](https://claude.com/claude-code) from our internal investigation.

eszlamczyk:
### What/Why?

Fixes #550 - an iOS, physical-device-only deadlock that froze the app and got it killed by the watchdog (`0x8BADF00D`); the reporter's production app logged ~19 such kills per day.

##### The mechanism
(confirmed from symbolicated production `.ips` reports and an instrumented RN-from-source build):
1. `measureContent` used `dispatch_sync` to the main thread - on every measure for the Dynamic Type font scale, on streaming ticks for `view.bounds`, and on cache misses to render/measure through a throwaway mock view.
2. RN's experimental `preventShadowTreeCommitExhaustion` flag (`releaseLevel: experimental`) re-runs an entire commit - **layout included - while holding `revisionMutexRecursive_`** after 3 failed optimistic commits.
3. Reanimated (with `DISABLE_COMMIT_PAUSING_MECHANISM`) commits from the main thread every frame: it both causes those consecutive commit failures during slow measured layouts and is the thread that then blocks on the mutex.
4. Once the JS thread measures under the locked commit and main is parked on the mutex, the next `dispatch_sync(main)` completes the cycle; JS waits for main, main waits for the mutex JS holds. Frozen forever.

The old code documented its own assumption (`safe because RN never synchronously joins the layout queue from main`) - RN's flag broke that assumption without breaking any contract. The only robust fix is structural: **nothing in the
measurement path may wait on the main thread, ever** (Yoga's `measureContent` makes no promises about calling context or held locks).

##### The fix 
Measurement is now main-thread-free by construction, mirroring RN core's `RCTTextLayoutManager` and this library's own Android `MeasurementStore`:

- **Font scale**: read from `LayoutContext::fontSizeMultiplier` (the parameter RN already delivers into `measureContent`, same source `ParagraphShadowNode` uses; RN refreshes it and re-runs layout on Dynamic Type changes). Pixel rounding uses `LayoutContext::pointScaleFactor` - `RCTScreenScale()` dispatch_syncs to main on first use.
- **Streaming size**: the view publishes its last committed size into a lock-free packed `std::atomic<uint64_t>` mailbox (`ENRMAtomicSize`) from `updateLayoutMetrics:`; the measure path reads it wait-free instead of syncing `view.bounds`.
- **View-free measurement** (`ENRMViewFreeMeasurement.h`): parse -> the SAME renderer the visible view uses (`ENRMRenderASTNodes` / `ENRMRenderSegmentsFromAST`, so measured layout ≡ rendered layout) -> a fresh per-call `NSTextStorage`/`NSLayoutManager`/`NSTextContainer` stack on the calling thread (TextKit is thread-confined, not main-only - the RCTTextLayoutManager / AsyncDisplayKit pattern). The segmented `EnrichedMarkdown` view gets static, view-free table and math measurers (`+[TableContainerView measureHeightForTableNode:…]`, `+[ENRMMathContainerView measureHeightForLatex:…]`) extracted from the same code the views run, so heights cannot drift. Mock-view measurement is deleted.
- **`@autoreleasepool` per measure**: the old main-thread path silently relied on the main runloop draining autoreleased objects; the JS/layout thread has no such drain, and without explicit pools sustained re-measurement accumulates until jetsam (caught during device verification).
- RaTeX concurrency (math now parses on the calling thread, possibly concurrently with the render thread) was audited against RaTeX sources - stateless engine, NSLock-guarded idempotent font registration, `thread_local` FFI error slot, immutable `OnceLock`/`LazyLock` statics, per-call renderer caches.

##### Independent bonus
The `dispatch_sync` design also serialized *all* markdown measurement onto the main thread (30–80 ms per node on device), visibly stuttering animations and scroll during streaming - the second complaint in #550. Measurement now runs off main entirely.

##### Known limitation
`EnrichedMarkdownTextInput`'s `measureContent` still measures live editable `UITextView` state through `dispatch_sync` and needs its own strategy - tracked separately. The threading contract in `ShadowMeasurementUtils.h` documents it as the one remaining violation.

### Testing

- Positive control (old library, dedicated repro app) aggressive harness -`MAX_COMMIT_ATTEMPTS_BEFORE_LOCKING = 1` in instrumented RN, width jitter forcing a full re-measure of every card every 900 ms, 8 streaming heads at 16 ms. Deadlocked within seconds, reproducibly. Paused-debugger stacks and the watchdog `.ips` (`0x8BADF00D`, `EXC_CRASH SIGKILL`) match the 20 production reports frame for frame:
    - JS thread: `__ulock_wait -> _dispatch_sync_f_slow -> ENRMFontScaleForMeasurement -> ENRMMeasureMarkdownContent -> measureContent -> yoga -> ShadowTree::tryCommit -> ShadowTree::commit` (locked fallback)
    - Main thread: `__psynch_mutexwait → recursive_mutex::lock →
    ShadowTree::tryCommit → reanimated::…::commitUpdates`
- Fix verification (this branch, identical harness, same iPhone 13 mini) - 20+ minutes, on the order of tens of thousands of locked-fallback engagements, every `LOCKED FALLBACK ENGAGED` paired with `DONE`, zero sync probes (nothing left to probe), no freeze, no watchdog, no jetsam. **Memory flat (~415 MB footprint on simulator over multi-minute profiling; CoreAnimation/malloc categories attributed with `footprint`/`heap`)**.
- Example app on simulator — Text screen (headings, images, nested lists, RTL/mixed-direction paragraphs, blockquote) renders end-to-end with no clipping/gaps through the final element; Stream screen (streaming tables + block LaTeX) keeps stable heights mid-stream and at completion. Did not find any regression (via maestro nor via manuall tests 
- Library pod and example app build clean for simulator and device.

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.h
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.h
@@ -13,6 +13,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)renderMarkdownSynchronously:(NSString *)markdownString;
 - (BOOL)hasRenderedMarkdown:(NSString *)markdown;
 - (BOOL)hasRenderedWithStyleFingerprint:(size_t)fingerprint;
+
+/// Size last committed by Fabric via `updateLayoutMetrics:`, readable from
+/// any thread without blocking (issue #550: the shadow-node measure path must
+/// never wait on the main thread, and reading `bounds` off-main is unsafe).
+/// Backed by a single lock-free atomic, so the pair can't tear. Returns
+/// CGSizeZero until the first layout is committed.
+- (CGSize)lastCommittedLayoutSize;
 @end
 
 NS_ASSUME_NONNULL_END
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
@@ -2,6 +2,7 @@
 #import "ContextMenuUtils.h"
 #import "ENRMAccessibilityLabels.h"
 #import "ENRMAsyncRenderCoordinator.h"
+#import "ENRMAtomicSize.h"
 #import "ENRMImageAttachment.h"
 #import "ENRMMarkdownParser.h"
 #import "ENRMTailFadeInAnimator.h"
@@ -67,7 +68,6 @@ typedef NS_OPTIONS(NSUInteger, ENRMDirtyFlags) {
 
 static char kENRMSegmentFadeAnimatorKey;
 
-
 @interface EnrichedMarkdown () <RCTEnrichedMarkdownViewProtocol, UITextViewDelegate, ENRMImageLayoutObserver>
 + (ENRMMd4cFlags *)flagsFromProps:(const EnrichedMarkdownMd4cFlagsStruct &)props;
 - (void)emitLinkPress:(NSString *)url;
@@ -119,6 +119,8 @@ static char kENRMSegmentFadeAnimatorKey;
 
   ENRMWritingDirectionMode _writingDirectionMode;
   NSWritingDirection _resolvedLayoutDirection;
+
+  ENRMAtomicSize _lastCommittedSize;
 }
 
 + (ComponentDescriptorProvider)componentDescriptorProvider
@@ -367,6 +369,11 @@ static char kENRMSegmentFadeAnimatorKey;
   return CGSizeMake(measuredWidth, measuredHeight);
 }
 
+- (CGSize)lastCommittedLayoutSize
+{
+  return _lastCommittedSize.load();
+}
+
 - (BOOL)hasRenderedMarkdown:(NSString *)markdown
 {
   return _renderedMarkdown != nil && [_renderedMarkdown isEqualToString:markdown];
@@ -409,6 +416,8 @@ static char kENRMSegmentFadeAnimatorKey;
            oldLayoutMetrics:(const LayoutMetrics &)oldLayoutMetrics
 {
   [super updateLayoutMetrics:layoutMetrics oldLayoutMetrics:oldLayoutMetrics];
+
+  _lastCommittedSize.store(CGSizeMake(layoutMetrics.frame.size.width, layoutMetrics.frame.size.height));
 
   NSWritingDirection resolved = _resolvedLayoutDirection;
   if (layoutMetrics.layoutDirection == LayoutDirection::RightToLeft) {
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.h
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.h
@@ -13,6 +13,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)renderMarkdownSynchronously:(NSString *)markdownString;
 - (BOOL)hasRenderedMarkdown:(NSString *)markdown;
 - (BOOL)hasRenderedWithStyleFingerprint:(size_t)fingerprint;
+
+/// Size last committed by Fabric via `updateLayoutMetrics:`, readable from
+/// any thread without blocking (issue #550: the shadow-node measure path must
+/// never wait on the main thread, and reading `bounds` off-main is unsafe).
+/// Backed by a single lock-free atomic, so the pair can't tear. Returns
+/// CGSizeZero until the first layout is committed.
+- (CGSize)lastCommittedLayoutSize;
 @end
 
 NS_ASSUME_NONNULL_END
--- a/packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
+++ b/packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
@@ -3,6 +3,7 @@
 #import "ContextMenuUtils.h"
 #import "ENRMAccessibilityLabels.h"
 #import "ENRMAsyncRenderCoordinator.h"
+#import "ENRMAtomicSize.h"
 #import "ENRMContextMenuTextView+macOS.h"
 #import "ENRMImageAttachment.h"
 #import "ENRMMarkdownParser.h"
@@ -112,6 +113,8 @@ typedef NS_OPTIONS(NSUInteger, ENRMDirtyFlags) {
   NSWritingDirection _resolvedLayoutDirection;
 
   ENRMDirtyFlags _dirtyFlags;
+
+  ENRMAtomicSize _lastCommittedSize;
 }
 
 + (ComponentDescriptorProvider)componentDescriptorProvider
@@ -142,6 +145,11 @@ typedef NS_OPTIONS(NSUInteger, ENRMDirtyFlags) {
   return size;
 }
 
+- (CGSize)lastCommittedLayoutSize
+{
+  return _lastCommittedSize.load();
+}
+
 - (BOOL)hasRenderedMarkdown:(NSString *)markdown
 {
   return _renderedMarkdown != nil && [_renderedMarkdown isEqualToString:markdown];
@@ -169,6 +177,8 @@ typedef NS_OPTIONS(NSUInteger, ENRMDirtyFlags) {
            oldLayoutMetrics:(const LayoutMetrics &)oldLayoutMetrics
 {
   [super updateLayoutMetrics:layoutMetrics oldLayoutMetrics:oldLayoutMetrics];
+
+  _lastCommittedSize.store(CGSizeMake(layoutMetrics.frame.size.width, layoutMetrics.frame.size.height));
 
   NSWritingDirection resolved = _resolvedLayoutDirection;
   if (layoutMetrics.layoutDirection == LayoutDirection::RightToLeft) {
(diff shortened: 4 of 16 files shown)


eszlamczyk:
Confirmed via maestro tests - no regression

hryhoriiK97 (commented):
Overall, it looks good! I found one regression with component measurement - check out TextScreen where we have long markdown. It looks like there's an additional margin at the end, even though allowTrailingMargin=false (which is applied by default) should eliminate it. Attaching screenshots for context.

| Before | After |
| - | - |
|  |  |

eszlamczyk:
Found the issue - (TLDR: RTL did wrong calculations) a raw NSLayoutManager defaults to `usesFontLeading = YES`, which adds the used fonts' leading on top of the paragraph style's lineHeight clamp, while the UITextView the view actually renders with ignores it. From what I've checked RN defaults this to NO so I've did the same

Effect:
image

hryhoriiK97 (approved):

PR #583, opened 2026-07-23
fix(android): crash indexoutofboundsexecption set span

### LINKED ISSUE #580: Android: crash IndexOutOfBoundsException: setSpan (-1 ... -1) on long-press when selectable is enabled on EnrichedMarkdownText
**Describe the bug**

Long-pressing an EnrichedMarkdownText with selection enabled (selectable defaults to true) crashes the app on Android when the platform's word-selection starts. The selection controller computes (-1 ... -1) offsets on the component's spannable and Selection.setSelection throws IndexOutOfBoundsException. Reproduced on a Samsung device — the trace goes through Samsung's semSetSelection / updateCharacterBasedSelectionAfterSelectWord path — but the invalid offsets come from the selection controller reading the component's spannable, so the word-boundary lookup on the custom spans looks like the root cause rather than the OEM.

**Steps to reproduce the behavior:**

Render <EnrichedMarkdownText markdown="some **markdown** text" /> (no selectable prop, or selectable={true})
On an Android device (reproduced on Samsung / One UI), long-press a word in the rendered text
See error (app crashes)

**Expected behavior**

Long-press starts text selection on the pressed word (or is a no-op) without crashing.

Device: Samsung (One UI)
OS: Android 13
Version: react-native-enriched-markdown 0.7.4, Expo 55, react-native 0.83.6 (Fabric / new architecture)

**Additional context**

Setting selectable={false} avoids the crash entirely — that's the workaround we ship. Stack trace:

java.lang.IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0
  android.text.SpannableStringInternal.checkRange(SpannableStringInternal.java:499)
  android.text.SpannableStringInternal.setSpan(SpannableStringInternal.java:199)
  android.text.SpannableString.setSpan(SpannableString.java:60)
  android.text.Selection.setSelection(Selection.java:94)
  android.text.Selection.setSelection(Selection.java:78)
  android.widget.TextView.semSetSelection(TextView.java:15735)
  android.widget.Editor$SelectionModifierCursorController.updateCharacterBasedSelectionAfterSelectWord(Editor.java:8868)
  android.widget.Editor$SelectionModifierCursorController.updateSelection(Editor.java:8634)
  android.widget.Editor$SelectionModifierCursorController.onTouchEvent(Editor.java:8579)
  android.widget.Editor.onTouchEvent(Editor.java:1962)
  android.widget.TextView.onTouchEvent(TextView.java:12294)
  com.swmansion.enriched.markdown.EnrichedMarkdownText.onTouchEvent(EnrichedMarkdownText.kt:359)
  android.view.View.dispatchTouchEvent(View.java:15544)
  ...

eszlamczyk:
### What/Why?

Fixes #580

Long-pressing an `EnrichedMarkdownText` (commonmark flavor, `selectable` default) crashes on
Samsung / One UI devices with:

```
java.lang.IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0
  at android.text.Selection.setSelection(...)
  at android.widget.TextView.semSetSelection(...)
```

##### Root cause. 
The component enables `setTextIsSelectable(true)` but installed a `LinkMovementMethod`-based movement method, and the two fight over the buffer's `Selection` spans. `LinkLongPressMovementMethod` removed the Selection on every `ACTION_DOWN` (and on UP/CANCEL, plus `LinkMovementMethod`'s own removal for any non-link touch), while the platform `Editor` - which processes each touch event *before* the movement method - was concurrently running its long-press word-selection machinery. Instrumentation on a stock emulator showed
`getSelectionStart()/getSelectionEnd()` returning `(-1, -1)` for the entire first half of every long-press gesture. AOSP tolerates that state; Samsung's One UI Editor caches the offsets mid-gesture and replays them after select-word via `semSetSelection` → `Selection.setSelection(spannable, -1, -1)` → crash. Replaying the cached offsets the way One UI
does reproduces the byte-identical exception on a stock emulator.

##### Fix.
`LinkLongPressMovementMethod` now extends `ArrowKeyMovementMethod` (the method `setTextIsSelectable` installs natively) and never mutates Selection spans. Link taps are dispatched from touch offsets: the pressed `LinkSpan` is recorded on `ACTION_DOWN` and `onClick` fires on `ACTION_UP` when the gesture stayed on the same span. Long-press scheduling, touch-slop cancellation, `isLinkTouchActive` (RNGH interop), spoiler taps, and the empty-space fall-through are unchanged. The same fix is ported to the copy in `packages/android-enriched-markdown/ui`.

###### Note
keyboard/d-pad traversal between links (a `LinkMovementMethod` feature) is not carried over; screen-reader link access is unaffected (handled by `accessibilityHelper`).

### Testing

> Repro setup (both paths): the reported crash is in the commonmark-flavor native view, and the example screens use `flavor="github"` (a different native view), so change the Playground preview to `flavor="commonmark"` (or remove the `flavor` prop) first. Then type `some **markdown** text` into the input so the preview renders it.

#### On a Samsung device (real crash, One UI) - NOT TESTED

1. Build the example app APK and install it (locally `yarn react-native-example android`, or upload the APK to a Remote Test Lab session - pick an Android 13 / One UI device).
2. In the Playground preview, **long-press a word and drag your finger slightly while holding**.
3. **Before this PR:** the app crashes with `IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0` (visible in logcat). **After this PR:** the word is selected normally with handles and the selection menu.

#### Without a Samsung device (any emulator - verifies the root cause and the fix)

The crash is Samsung-only, but the state that causes it is created by our code and is observable
anywhere. Add temporary logging around `super.onTouchEvent` in `EnrichedMarkdownText.onTouchEvent`:

```kotlin
Log.d("SelDebug", "${MotionEvent.actionToString(event.action)} BEFORE: $selectionStart..$selectionEnd")
val result = super.onTouchEvent(event)
Log.d("SelDebug", "${MotionEvent.actionToString(event.action)} AFTER:  $selectionStart..$selectionEnd")
```

1. Run the app, long-press a word in the preview while watching `adb logcat -s SelDebug`.
2. **Before this PR:** `ACTION_DOWN` flips the offsets to `-1..-1` and they stay there for every
   `ACTION_MOVE` until the long-press timeout selects the word — this is the invalid state
   Samsung's Editor caches and later feeds into `semSetSelection(-1, -1)`.
   **After this PR:** the offsets never become -1 at any point in the gesture; they keep their
   previous value until the long-press sets the word range.
3. (Optional, exact-crash proof) Replay the cached offsets the way One UI does — cache
   `selectionStart/End` after `super` on `ACTION_DOWN`, and on a later `ACTION_MOVE` where the
   buffer has a valid selection, call
   `Selection.setSelection(text as Spannable, cachedStart, cachedEnd)`. Before this PR a single
   long-press throws the byte-identical `setSpan (-1 ... -1)` exception on a stock emulator;
   after this PR the replay never fires because the cached offsets are never -1.

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
+++ b/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
@@ -2,21 +2,39 @@ package com.swmansion.enriched.markdown.utils.text.view
 
 import android.os.Handler
 import android.os.Looper
-import android.text.Selection
 import android.text.Spannable
-import android.text.method.LinkMovementMethod
+import android.text.method.ArrowKeyMovementMethod
 import android.view.MotionEvent
 import android.view.ViewConfiguration
 import android.widget.TextView
 import com.swmansion.enriched.markdown.spans.LinkSpan
 import kotlin.math.abs
 
-class LinkLongPressMovementMethod : LinkMovementMethod() {
+/**
+ * Movement method that adds link tap / long-press handling on top of
+ * [ArrowKeyMovementMethod], the method `setTextIsSelectable` installs.
+ *
+ * This class must never mutate the buffer's Selection spans. It previously
+ * extended LinkMovementMethod, which removes the Selection on every touch that
+ * does not land on a link. That deleted the selection the platform Editor was
+ * concurrently building during a long-press gesture, so the Editor observed
+ * getSelectionStart()/getSelectionEnd() == -1 mid-gesture. AOSP tolerates that
+ * state, but Samsung's One UI Editor caches the offsets during the gesture and
+ * replays them after select-word via semSetSelection ->
+ * Selection.setSelection(spannable, -1, -1), crashing with
+ * "IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0".
+ * Link clicks are therefore dispatched from touch offsets here instead of
+ * relying on LinkMovementMethod's selection-based implementation, and
+ * [ArrowKeyMovementMethod] keeps text selection behavior identical to a plain
+ * selectable TextView.
+ */
+class LinkLongPressMovementMethod : ArrowKeyMovementMethod() {
   private val handler = Handler(Looper.getMainLooper())
   private var longPressRunnable: Runnable? = null
 
   private var startX = 0f
   private var startY = 0f
+  private var pressedLink: LinkSpan? = null
 
   var isLinkTouchActive: Boolean = false
     private set
@@ -31,9 +49,9 @@ class LinkLongPressMovementMethod : LinkMovementMethod() {
         startX = event.x
         startY = event.y
 
-        val span = findLinkSpan(widget, buffer, event)
-        isLinkTouchActive = span != null
-        span?.let { scheduleLongPress(widget, it) }
+        pressedLink = findLinkSpan(widget, buffer, event)
+        isLinkTouchActive = pressedLink != null
+        pressedLink?.let { scheduleLongPress(widget, it) }
       }
 
       MotionEvent.ACTION_MOVE -> {
@@ -43,25 +61,33 @@ class LinkLongPressMovementMethod : LinkMovementMethod() {
         ) {
           cancelLongPress()
           isLinkTouchActive = false
+          pressedLink = null
         }
       }
 
-      MotionEvent.ACTION_UP, MotionEvent.ACTION_CANCEL -> {
+      MotionEvent.ACTION_UP -> {
+        cancelLongPress()
+        val tappedLink = pressedLink
+        isLinkTouchActive = false
+        pressedLink = null
+
+        // LinkSpan.onClick itself swallows the click that follows a completed
+        // long-press (and resets its internal flag), so it is always invoked
+        // for a tap that started and ended on the same link.
+        if (tappedLink != null && findLinkSpan(widget, buffer, event) === tappedLink) {
+          tappedLink.onClick(widget)
+          return true
+        }
+      }
+
+      MotionEvent.ACTION_CANCEL -> {
         cancelLongPress()
         isLinkTouchActive = false
-        if (widget.hasSelection()) {
-          Selection.removeSelection(buffer)
-        }
+        pressedLink = null
       }
     }
 
-    val result = super.onTouchEvent(widget, buffer, event)
-
-    if (event.action == MotionEvent.ACTION_DOWN) {
-      Selection.removeSelection(buffer)
-    }
-
-    return result
+    return super.onTouchEvent(widget, buffer, event)
   }
 
   private fun scheduleLongPress(
@@ -72,9 +98,6 @@ class LinkLongPressMovementMethod : LinkMovementMethod() {
 
     longPressRunnable =
       Runnable {
-        if (widget.hasSelection()) {
-          Selection.removeSelection(widget.text as Spannable)
-        }
         if (span.onLongClick(widget)) {
           widget.cancelLongPress()
         }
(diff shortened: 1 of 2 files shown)


packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
```diff
@@ -31,9 +49,9 @@ class LinkLongPressMovementMethod : LinkMovementMethod() {
         startX = event.x
         startY = event.y
 
-        val span = findLinkSpan(widget, buffer, event)
-        isLinkTouchActive = span != null
-        span?.let { scheduleLongPress(widget, it) }
+        pressedLink = findLinkSpan(widget, buffer, event)
+        isLinkTouchActive = pressedLink != null
+        pressedLink?.let { scheduleLongPress(widget, it) }
```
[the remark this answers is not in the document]
eszlamczyk: Fixed. It was actually a bit worse that described: the unguarded `charOffsetAt()` predicates the PR, but previously it only fed long-press scheduling and `isLinkTouchActive`, while link click went through other methods, which has its own out-of-bounds guards since Android 12 I believe. 

Ported the RN copy's guards to the UI copy

packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
```diff
@@ -2,24 +2,43 @@ package com.swmansion.enriched.markdown.utils.text.view
...
+ * getSelectionStart()/getSelectionEnd() == -1 mid-gesture. AOSP tolerates that
+ * state, but Samsung's One UI Editor caches the offsets during the gesture and
+ * replays them after select-word via semSetSelection ->
+ * Selection.setSelection(spannable, -1, -1), crashing with
+ * "IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0".
+ * Link clicks are therefore dispatched from touch offsets here instead of
+ * relying on LinkMovementMethod's selection-based implementation, and
+ * [ArrowKeyMovementMethod] keeps text selection behavior identical to a plain
+ * selectable TextView.
+ */
```
hryhoriiK97, line 21: The crash forensics are already well documented in the PR description and commit. Let's replace with something like this:
```
/**
 * Movement method that adds link tap / long-press handling on top of
 * [ArrowKeyMovementMethod], the method [setTextIsSelectable] installs.
 *
 * Must never mutate the buffer's Selection spans — the platform Editor
 * manages them during long-press gestures, and removing or overwriting
 * them mid-gesture crashes on some OEM skins.
 * See: https://github.com/software-mansion/react-native-enriched-markdown/issues/580
 */
```

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
```diff
@@ -13,12 +12,31 @@ import com.swmansion.enriched.markdown.spans.SpoilerSpan
...
+ * getSelectionStart()/getSelectionEnd() == -1 mid-gesture. AOSP tolerates that
+ * state, but Samsung's One UI Editor caches the offsets during the gesture and
+ * replays them after select-word via semSetSelection ->
+ * Selection.setSelection(spannable, -1, -1), crashing with
+ * "IndexOutOfBoundsException: setSpan (-1 ... -1) starts before 0".
+ * Link clicks are therefore dispatched from touch offsets here instead of
+ * relying on LinkMovementMethod's selection-based implementation, and
+ * [ArrowKeyMovementMethod] keeps text selection behavior identical to a plain
+ * selectable TextView.
+ */
```
hryhoriiK97, line 23: The same as above 🔝

hryhoriiK97 (approved):
Looks good! I left a few minor comments.

PR #597, opened 2026-07-27
feat: match rn textinput keyboard-dismiss behavior on taps outside the input

### LINKED ISSUE #577: Match RN TextInput keyboard-dismiss behavior on taps outside the input

With a stock React Native TextInput inside a ScrollView/FlatList, the default keyboardShouldPersistTaps="never" behavior comes for free: tapping outside the focused input dismisses the keyboard, while scrolling behaves according to keyboardDismissMode. EnrichedMarkdownTextInput doesn't participate in this — when it's focused and the user taps elsewhere on the screen (e.g. on the message list above a chat composer), the input keeps focus and the keyboard stays up. RN's responder system can't auto-blur it the way it does the built-in TextInput, presumably because focus/first-responder handling lives entirely in the custom native view and isn't registered with RN's TextInputState tracking.

Parity with the built-in TextInput: register the input's focus state with React Native's text-input state tracking (TextInputState / the Fabric focus registry) so that scroll containers' existing keyboardShouldPersistTaps logic treats it exactly like a native TextInput. Plain taps outside the input should blur it and dismiss the keyboard, scrolling should remain possible without dismissing (unless keyboardDismissMode says otherwise), and keyboardShouldPersistTaps="always"/"handled" should keep their usual meaning.

App-level workaround (what we do now): hold the input ref in a provider, expose a dismissKeyboard() action that calls ref.blur(), and wire a tap-gesture handler on the surrounding message list that calls it — while carefully not stealing scroll or item-press gestures. It works, but it's boilerplate every consumer embedding the input in a chat-style screen will need, and it's easy to get subtly wrong (taps on interactive children, tap-vs-scroll discrimination).
An opt-in prop (e.g. blurOnOutsideTap / dismissKeyboardOnTapOutside) implementing the same semantics natively would cover the common case, if full TextInputState integration isn't feasible.

**Additional context**

Observed with v0.7.4 in an Expo 55 / RN 0.83.6 chat screen where the input sits below a message list. Happy to provide a repro or test builds if useful.

eszlamczyk:
### What/Why?
Closes #577.

When `EnrichedMarkdownTextInput` was focused inside a `ScrollView`/`FlatList`, tapping outside it (e.g. on the message list above a chat composer) left the input focused and the keyboard up - unlike the built-in TextInput, which blurs per the scroll container's keyboardShouldPersistTaps setting.

RN's tap-to-dismiss behavior is implemented entirely in JS: ScrollView's responder handlers consult TextInputState (a JS-side registry of host instances) to decide whether a tap should dismiss the keyboard. A custom input that never registers is invisible to that logic — ScrollView concludes no text input is focused and does nothing.

We mirror behaviour of RN by:
  - register on mount / unregister (blurring first if focused) on unmount,
  - `focusInput`/`blurInput` on the native focus/blur events, so all focus sources are tracked (tap, ref.focus(), autoFocus).

### Testing
In the example app's Playground screen (its ScrollView uses keyboardShouldPersistTaps="handled"):
1. Focus the input so the soft keyboard opens.
2. Tap a non-touchable area (the Preview box, padding between button rows) → input blurs, keyboard dismisses. Taps on
buttons still fire without dismissing (correct 'handled' semantics).
3. Temporarily remove the keyboardShouldPersistTaps prop (default 'never') → the first tap on a button only dismisses the
keyboard, the second presses it - stock TextInput two-tap behavior.
4. Tap inside the focused input → caret moves, keyboard stays.
5. Scroll while focused → keyboard stays.

#### Screenshots 

https://github.com/user-attachments/assets/76bf9f58-c6cb-47f1-aa6f-c2cb9108ff78

https://github.com/user-attachments/assets/594a1d7f-8a16-41d6-8cc6-d17b86f9fd83

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/src/utils/textInputState.ts
+++ b/packages/react-native-enriched-markdown/src/utils/textInputState.ts
@@ -1,0 +1,43 @@
+/**
+ * Bridges EnrichedMarkdownTextInput into React Native's TextInputState focus
+ * registry so it participates in RN's keyboard-dismiss behavior (issue #577).
+ *
+ * RN's tap-to-dismiss logic (ScrollView's keyboardShouldPersistTaps handling
+ * and Keyboard.dismiss) is implemented entirely in JS. ScrollView's responder
+ * handlers consult TextInputState, a plain JS module holding the set of
+ * registered inputs and the currently focused one, to decide whether a tap
+ * outside the focused field should blur it. A custom native input that never
+ * registers is invisible to that logic: ScrollView concludes no text input is
+ * focused and leaves the keyboard up.
+ *
+ * Registering a non-TextInput host component is safe by design. When
+ * ScrollView dismisses, TextInputState.blurTextInput dispatches the built-in
+ * TextInput's "blur" command at the registered instance, and Fabric command
+ * dispatch is name-based; TextInputState.js states "commands don't actually
+ * care as long as the thing being passed in actually has a command with that
+ * name". EnrichedMarkdownTextInput ships focus/blur commands under those
+ * exact names, so the dispatch lands in the existing native handlers and no
+ * native changes are needed.
+ *
+ * The registry has no complete public export: TextInput.State exposes only
+ * blurTextInput/focusTextInput/currentlyFocusedInput and lacks the
+ * registerInput/focusInput/blurInput bookkeeping functions, so a deep import
+ * is required. This is the same approach @expensify/react-native-live-markdown
+ * uses. The import is typed here via a local interface; requiring at runtime
+ * keeps the internal module out of the public type surface.
+ */
+import type { HostInstance } from 'react-native';
+
+interface TextInputStateModule {
+  registerInput(input: HostInstance): void;
+  unregisterInput(input: HostInstance): void;
+  focusInput(input: HostInstance | null): void;
+  blurInput(input: HostInstance | null): void;
+  blurTextInput(input: HostInstance | null): void;
+  currentlyFocusedInput(): HostInstance | null;
+}
+
+export const TextInputState =
+  // eslint-disable-next-line @react-native/no-deep-imports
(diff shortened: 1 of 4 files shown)


packages/react-native-enriched-markdown/src/utils/textInputState.ts
```diff
@@ -0,0 +1,51 @@
...
+ * exact names, so the dispatch lands in the existing native handlers and no
+ * native changes are needed.
+ *
+ * The registry has no complete public export: TextInput.State exposes only
+ * blurTextInput/focusTextInput/currentlyFocusedInput and lacks the
+ * registerInput/focusInput/blurInput bookkeeping functions, so a deep import
+ * is required. This is the same approach @expensify/react-native-live-markdown
+ * uses. The import is typed here via a local interface; requiring at runtime
+ * keeps the internal module out of the public type surface.
+ */
```
hryhoriiK97, line 12: nit: This comment is doing a lot of heavy lifting - the "how ScrollView/TextInputState work" section is already well explained in the PR description and docs. Consider trimming to just the why + why-it's-safe.

```
/**
 * Bridges EnrichedMarkdownTextInput into React Native's TextInputState focus
 * registry so it participates in RN's keyboard-dismiss behavior (issue #577).
 *
 * Registering a non-TextInput host component is safe: TextInputState's
 * blurTextInput dispatches the built-in TextInput's "blur" command at the
 * registered instance, and Fabric command dispatch is name-based — it works
 * as long as the component ships a command with that name, which ours does.
 *
 * A deep import is required because the public TextInput.State API lacks
 * registerInput/focusInput/blurInput. This matches the approach used by
 * @expensify/react-native-live-markdown.
 */
```

hryhoriiK97 (approved):
LGTM! I've left one nit comment 👍

PR #601, opened 2026-07-27
feat: padding and broderradius for blockquotes

### LINKED ISSUE #103: Add padding and borderRadius option for blockquotes
Would be nice to add padding and border radius options to blockquotes ! Thank you so much for making such a great and useful library !

eszlamczyk:
### What/Why?
 
Closes #103.

Adds `padding` and `borderRadius` options to `markdownStyle.blockquote`, implemented the same way as the code block background.

- `borderRadius` (default `0`) - rounds the full-width blockquote background box. The accent stripe is clipped to the rounded shape, and each nested quote's stripe rounds against its own box, matching per-element `border-radius` on web.
- `padding` (default `0`) - inner top/bottom spacing between the background box edges and the quote content (also applies to the trailing edge on iOS and web, same platform behavior as `codeBlock.padding`). Applied at every nesting level, so nested quotes pad their own box and paddings stack at shared boundaries like nested elements on web. Leading-edge spacing is still `borderWidth` + `gapWidth`.

### Testing
Example app / storybook - Blockquote stories with the new `borderRadius` and `padding` controls, on iOS and Android - including nested quotes, quote as the last document element, and defaults left at `0` (unchanged rendering).

#### Screenshots

| | iOS | Android |
| - | - | - |
| Default (before) | image | image |
| `Padding: 10, borderRadius: 10` | image | image |

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/apps/react-native-example/.rnstorybook/stories/components/EnrichedMarkdownText/block/Blockquote.stories.tsx
+++ b/apps/react-native-example/.rnstorybook/stories/components/EnrichedMarkdownText/block/Blockquote.stories.tsx
@@ -67,6 +67,16 @@ const argTypes = {
     control: 'color',
     description: 'markdownStyle.blockquote.backgroundColor',
   },
+  borderRadius: numberControl('markdownStyle.blockquote.borderRadius', {
+    min: 0,
+    max: 16,
+    step: 1,
+  }),
+  padding: numberControl('markdownStyle.blockquote.padding', {
+    min: 0,
+    max: 32,
+    step: 2,
+  }),
 };
 
 function renderBlockquote(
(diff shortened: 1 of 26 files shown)


hryhoriiK97 (commented):
Looks good! I found one bug - looks like in some cases the bottom-left border radius is not applied on Android (attaching the screenshots). Can you look into this please?

Image

Image

eszlamczyk:
Aparently the issues were of-by one calculations for bottom border radius calculations, and the fact that we increased the spacer for padding by changing line heights, that broke when we had two calculations like that line by line. Fixed by implementing it the way iOS does, by adding additional marker that we manip for paddings.

image
image

hryhoriiK97 (approved):

PR #611, opened 2026-07-29
fix(android): inline code drawing range calculations

eszlamczyk:
### What/Why?
`CodeBackgroundSpan` measures character positions by building a throwaway `StaticLayout` from `text.subSequence(lineStart, lineEnd)`. That subsequence keeps all spans, including `LeadingMarginSpan`s (`BaseListSpan`, `BlockquoteSpan`), so the measuring layout already applies the list indent to `getPrimaryHorizontal`. `drawBackground` then added `leadingMargin` on top of that, shifting the box right by exactly the marker indent.

The fix trusts the measuring layout's margin for measured offsets and only adds `leadingMargin` explicitly in the branches where no layout is built (span starting at line start, continuation lines). Applied to both Android source trees:

- `packages/react-native-enriched-markdown/.../CodeBackgroundSpan.kt` -  fixes the rightward shift in lists
- `packages/android-enriched-markdown/.../CodeBackgroundSpan.kt` - this copy had no margin handling at all, so continuation lines of a wrapped code span were missing the indent

### Testing

In the example app (`apps/react-native-example`), open the **Text** screen and scroll to the any codeblock should hug the text exactly. Also verified in the **Playground** preview with inline code in a nested list (two stacked leading margins), in a plain paragraph, and inside a blockquote.

| Before | After |
| - | - |
| image | image |
| image | image |

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/spans/CodeBackgroundSpan.kt
+++ b/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/spans/CodeBackgroundSpan.kt
@@ -7,6 +7,7 @@ import android.graphics.RectF
 import android.text.Spanned
 import android.text.StaticLayout
 import android.text.TextPaint
+import android.text.style.LeadingMarginSpan
 import android.text.style.LineBackgroundSpan
 import com.swmansion.enriched.markdown.styles.StyleConfig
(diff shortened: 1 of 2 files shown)


hryhoriiK97 (commented):
LGTM! Thank you for the fix 🎖️

hryhoriiK97 (approved):

PR #612, opened 2026-07-30
fix(input): keep styles and list block when typing over a leading selection; stop iOS edit-menu jitter

wildseansy:
## Summary

Fixes three `EnrichedMarkdownTextInput` bugs reported on iOS (the first reproduces on Android too — the shared logic is mirrored there):

**1. Selecting a bold run at the beginning of a list item and typing over it lost the bold style and deleted the list item.**

Root cause is in the shared range edit adjustment: a replacement whose deletion starts exactly at a stored range's start classified as `FullyDeleted`/`ClippedStart` and either removed the range outright (bold lost) or moved it past the inserted text — shoving the block range off its line start, where `pruneOrphanedBlockAnchors` then deleted the whole list block before `normalizeToLineBounds` could self-heal it.

- `ENRMRangeEditAdjustment` / `RangeEditAdjustment.kt`: replacements starting at a range's start now let the inserted text join the range, matching UIKit's rule that replacement text inherits the attributes of the first replaced character. Links deliberately keep the old behavior — typing over a selected link/mention replaces it rather than extending it.
- Pending styles are now seeded from the selection start when the selection is non-empty, so a continued typing run stays styled through the post-edit grace period (previously only the first typed character kept the style). Per review feedback, Android now mirrors this: `onSelectionChanged` seeds `pendingStyles` the same way (and a ranged style toggle reseeds, since it changes what's active without moving the selection).
- Android `BlockStore` guards against re-adding a collapsed block that survived adjustment via inheritance.

**2. The selection edit menu repeatedly jumped up and down while selecting text.**

Three compounding causes in `EnrichedMarkdownTextInput.mm`, all on the selection-change hot path:

- The atomic-link selection snap wrote `selectedRange` on every frame of a live selection-handle drag; each programmatic write makes UIKit dismiss and re-present the edit menu while the gesture immediately reasserts the finger-derived range. The snap is now deferred until the gesture settles, and is clamped to the text length (a stale link range extending past the text end could previously drive an unbounded snap/clamp loop).
- `updateEmptyBulletMarker` unconditionally forced `layoutIfNeeded` + a full-view `setNeedsDisplay` on every selection-change delegate fire; it now early-returns when there is no marker to draw or clear, and only stamps the paragraph style into storage when it is actually missing.
- Typing attributes were rewritten on every selection change; they are now only synced at a caret (UIKit derives replacement attributes from the first selected character) and only assigned when they actually changed.

**3. iOS: `setValue()` rendered the imported markdown unstyled — no bullet, indent, or bold — until the first edit.**

Originally noted here as a pre-existing issue to file separately. We fixed it on this branch (`importMarkdown` applied formatting while the edit session was still in the `Importing` phase, which suppresses it), but #617 landed the identical fix on `main` in the meantime — after merging `main` back in, this PR's diff no longer carries that change. Verified here regardless: `setValue('- **boldword** rest')` renders the bullet, indent, and bold immediately.

## Testing

- Standalone unit checks of the adjustment function (12 replacement/deletion/insertion overlap cases, including the link exclusion and the pure-deletion paths) all pass.
- Maestro repro of the reported bug — set `- **boldword** rest`, double-tap-select the bold word, type over it — now serializes `- **typed** rest` (bold kept, list item intact). Verified on both platforms; on Android the replacement was typed through the IME key-by-key, confirming the pending-style seed (not just first-character range inheritance).
- New visual regression e2e (`list_bold_replace_selection_test`): screenshot-asserts the editor after typing over the selected leading bold word of a list item — covers the bullet marker and full-word bold visually rather than via serialization. iOS golden included; `ios-only` until an Android golden is captured on the CI AVD (same convention as `paste_markdown_test`).
- `setValue('- **boldword** rest')` now renders the bullet, indent, and bold immediately (previously plain until the first edit), matching the display component preview.
- Existing input e2e suite (inline + block flows, iOS simulator): identical results before/after the change — no regressions; the only failing flows fail on `main` as well (two missing iOS screenshot baselines for `ordered_list_test`/`list_toggle_test`, `link_test`'s `toolbar-link` sitting off-screen in the scrolling toolbar, and a sub-threshold pixel diff on a newer iOS runtime than the baselines).
- iOS example app builds; Android `compileDebugKotlin` + `ktlintCheck` pass; `tsc` clean.


eszlamczyk (approved):
hi @wildseansy, the fix seems sound and reasonable! One note: Im not sure why this `bump version` got included in this, even though this isn't a release commit. Im gonna clean up this and a bunch of other code nitpicks and merge this today!

eszlamczyk (changes requested):
Again thanks for this - the block-range reasoning is sound and the range-adjustment parity is carefully done. Unfortunately, testing it interactively, the inline-style behavior regresses, and I think the root cause is that the approach fixes the stored ranges / serialization but not the visual application - and the only automated coverage here asserts serialized markdown, which is exactly the part that stays correct in every failing case. So the green check doesn't tell us the feature works.

I'll keep this to the two regressions I've confirmed are introduced by this PR (there are some other visual glitches in this area, but they reproduce on `main` too, so they're out of scope here).

**Repro (both platforms):** `- **boldword** rest`, double-tap-select the bold word, type a replacement.

### Android - only the first typed char stays bold (`**t**est` instead of `**test**`)

Confirmed in code. The PR ports the range-inheritance half to Android but not the "seed pending styles from the selection" half that the iOS side got:

- `RangeEditAdjustment` inheritance makes the bold range `[start, start+1]` after the first char. But the insert-only branch doesn't grow a range when inserting at `range.end` (`RangeEditAdjustment.kt` - both conditions are false for an insert at `range.end`), so char 2+ fall outside it.
- Nothing carries the style forward: `EnrichedMarkdownTextInputView.onSelectionChanged` only *clears* `pendingStyles`/`pendingStyleRemovals` (~L377-378), it never seeds them from the selection start, unlike iOS `ENRMInputTypingAttributesController.rebuildFromContext`.

So the PR description's "the shared logic is mirrored there [Android]" only holds for the range-inheritance, not the pending-style seed. Android needs the equivalent seed in `onSelectionChanged` for a non-empty selection.

### iOS - the list bullet disappears when you replace the first word of a list item

Repro: in a list item, double-tap the first word and type over it -> the bullet marker vanishes (indent and serialized
markdown stay correct). This is tied to the new block-range inheritance path (`ENRMBlockStore` now passes `inheritsReplacementAtStart = YES`): the list block range is adjusted differently on that edit and the marker stops drawing. I haven't fully root-caused the internal mechanism (it needs on-device tracing of the block-range state after `adjustForEdit` + `normalizeToLineBounds` and what the marker drawer reads), but it's reliably reproducible and specific to this change.

### Ask before merge

1. Android: seed pending styles from the selection start (mirror iOS `rebuildFromContext`).
2. iOS: fix the bullet-marker regression on first-word replacement.
3. Add a **visual** regression check (screenshot/attribute assertion) rather than serialization-only - that's what let both
of these through.

wildseansy:
Thanks for the review @eszlamczyk, I'll take a look tonight and update with points 1-3.

wildseansy:
@eszlamczyk Thanks for the thorough review — pushed 4bbc4d3 addressing it:

1. **Android**: done — `onSelectionChanged` now seeds pending styles from the selection start (mirroring iOS `rebuildFromContext`), and a ranged toggle reseeds. Verified on emulator: typing over the selected bold word char-by-char yields `- typed **newword**`, fully bold.

2. **iOS bullet marker**: I couldn't reproduce this on a verified-fresh build, and found a likely explanation: with math enabled, the RN `spm_dependency` (RaTeX) integration generates a corrupt `Pods.xcodeproj` (missing PBXProject object). Xcode 26.x reports it damaged and silently skips the pod targets, so builds reuse whatever pod binaries are already in derived data — i.e. interactive testing can run *stale* natives while appearing freshly built. After forcing a true rebuild, the exact repro (`- **boldword** rest`, double-tap bold word, type) gives `- **typed** rest` with the bullet drawn and the block-type attribute stamped at paragraph start on every keystroke. Could you re-test on a build where the ReactNativeEnrichedMarkdown pod verifiably recompiled (e.g. `ENRICHED_MARKDOWN_ENABLE_MATH=0 pod install` first)? Happy to dig further if it still reproduces.

3. **Visual coverage**: added `list_bold_replace_selection_test` — screenshot assertion of the editor after typing over the selected leading bold word of a list item (bullet + bold), iOS golden included; `ios-only` pending a CI-AVD golden, matching `paste_markdown_test`.

Worth filing separately: the corrupt-Pods/SPM issue itself, the `setValue()` no-render-until-first-edit bug (already noted in the PR description), and two quirks I hit interactively on both platforms but didn't classify against main (pending toolbar toggle lagging one word; formatting ranges surviving `setValue('')`).

eszlamczyk (approved):
@wildseansy Looks like the bug I described is a separate issue, so we're approving this one. Thanks for your contribution!

Issue #51, opened 2026-01-21
Support Code Block Highliter


eeee0717:
This library is really awesome! I wonder if there are any plans to support code block highlighting.

Bartek532:
Hey, is this still in plans? Can we see the roadmap somewhere?

hryhoriiK97:
Hey @Bartek532, this is still in the plans. We plan to revisit code blocks and implement code block highlighting and their styling overall.

Regarding the roadmap - we have a [future plans section](https://github.com/software-mansion-labs/react-native-enriched-markdown?tab=readme-ov-file#future-plans), but I should update it. Thank you for mentioning this!

PR #624, opened 2026-08-03
feat: codeblock highlight colors

eszlamczyk:
### What/Why?

First part of the code block syntax highlighting stack for #51.

This PR wires up the **styling plumbing** for per-token code block colors end to end, without shipping the highlighting engine itself. It adds a `syntaxColors` object to `codeBlock` styles, keyed on tree-sitter highlight token types (`keyword`, `string`, `comment`, `function`, `type`, etc.), and threads it through:

- **TS types** — `CodeBlockSyntaxColors` on `CodeBlockStyle`, plus the internal style type.
- **Style normalization** — `normalizeMarkdownStyle` (and the `.web` variant) resolve the palette, applying the default GitHub-light colors and falling back to the code block's base `color` for `operatorColor`/`punctuation`/`variable`/`embedded`.
- **Codegen specs** — both native component specs expose the new props.
- **iOS** — `StyleConfig`, `ENRMCodeBlockHighlighter`, and `StylePropsUtils` consume the colors.
- **Android** — `CodeBlockStyle`, `CodeBlockRenderer`, and `CodeBlockHighlighter` consume the colors.
- **Docs** — `STYLES.md` documents `syntaxColors` and each token key.
- **Storybook** — a `CodeHighlight` story for the example app.

Notes:
- `operator` is named `operatorColor` in the public API because `operator` is a reserved word in the generated native (C++) struct.
- Colors only take **visible** effect once the optional syntax highlighting module (tree-sitter grammars) is compiled in — that lands in a later PR in the stack. Until then code blocks render uncolored, so this PR is a safe no-op at runtime.

This **addresses** #51 (does not close it) — it's the foundation for the highlighter that follows.

### Testing

- Ran the example app on iOS and Android; the new `CodeHighlight` Storybook story renders and accepts a `syntaxColors` palette without regressions to existing code block rendering.
- Verified omitted keys fall back to the default palette / base `color`.

#### Screenshots 
Screenshots are included in second PR (#625) as without real paring its not viable to do anything with those colors

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/src/types/MarkdownStyle.ts
+++ b/packages/react-native-enriched-markdown/src/types/MarkdownStyle.ts
@@ -50,12 +50,41 @@ interface ListStyle extends BaseBlockStyle {
   itemSpacing?: number;
 }
 
+/**
+ * Per-token syntax highlight colors for fenced code blocks, keyed on the
+ * tree-sitter highlight token types. Any key omitted falls back to the default
+ * palette (Operator/Punctuation/Variable/Embedded inherit the code block's
+ * base `color`). Colors only take visible effect when the optional syntax
+ * highlighting module is compiled in; otherwise code blocks render uncolored.
+ */
+interface CodeBlockSyntaxColors {
+  keyword?: string;
+  /**
+   * Color for operator tokens. Named `operatorColor` (not `operator`) because
+   * `operator` is a reserved word in the generated native (C++) struct.
+   */
+  operatorColor?: string;
+  punctuation?: string;
+  string?: string;
+  number?: string;
+  constant?: string;
+  comment?: string;
+  function?: string;
+  type?: string;
+  variable?: string;
+  property?: string;
+  tag?: string;
+  attribute?: string;
+  embedded?: string;
+}
+
 interface CodeBlockStyle extends BaseBlockStyle {
   backgroundColor?: string;
   borderColor?: string;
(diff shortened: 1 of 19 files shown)


packages/react-native-enriched-markdown/src/types/MarkdownStyle.ts
```diff
@@ -50,12 +50,41 @@ interface ListStyle extends BaseBlockStyle {
...
+ * base `color`). Colors only take visible effect when the optional syntax
+ * highlighting module is compiled in; otherwise code blocks render uncolored.
+ */
+interface CodeBlockSyntaxColors {
+  keyword?: string;
+  /**
+   * Color for operator tokens. Named `operatorColor` (not `operator`) because
+   * `operator` is a reserved word in the generated native (C++) struct.
+   */
+  operatorColor?: string;
```
hryhoriiK97: Should the public API use operator instead of `operatorColor`? operator is valid in TypeScript/JavaScript and would be consistent with the rest of the keys (keyword, string, comment, etc.). The C++ reserved-word constraint only applies to the codegen/native layer — `normalizeMarkdownStyle` could map operator → operatorColor when building the internal struct. Easier to rename now before consumers depend on it.

hryhoriiK97 (approved):
LGTM! I've left one comment 👍

PR #625, opened 2026-08-03
feat: tressiter for colored codeblocks

### LINKED ISSUE #51: Support Code Block Highliter

eszlamczyk:
### What/Why?

**Second part of #51 — closes #51.**

Part one landed the code-block segmentation/rendering seam. This PR wires the
actual native syntax highlighter behind it: fenced code blocks are now
highlighted on iOS and Android via a vendored [tree-sitter](https://tree-sitter.github.io/)
runtime plus a curated set of grammars, foreground-only (it only recolors
tokens, never changes text metrics, so a block's measured height always matches
its drawn height).

Highlighting is on by default with a curated language set, and can be trimmed or
fully disabled per platform to control binary size.

#### The vendoring problem it solves

Tree-sitter grammars ship as machine-generated `parser.c` parse tables — the full
supported set is **~177 MB / ~1.1M LOC** of generated C. Committing that would
bloat every clone, dominate PR diffs, and swamp GitHub's language stats. It also
has no business living in git: it is 100% reproducible from the pinned grammar
packages. The same is true of the tree-sitter runtime (~21k lines) and the
codegen'd registry — all machine-generated, all reproducible from pins.

So **none of the vendored C is committed**. The entire `vendor/` tree is gitignored
and reproduced on demand from `vendor/grammar-versions.json`:

- **`vendor/grammars/`** — the ~177 MB of `parser.c`/`scanner.c` tables, copied
  from the grammar **devDependencies** (version-pinned in the lockfile).
- **`vendor/tree-sitter/`** — the ~21k-line runtime, fetched and **sha256-verified**
  from the pinned GitHub release tarball (`runtime.tarball` + `runtime.sha256` in
  the manifest). No native build, no compiler — just `lib/src` + `lib/include`.
- **`vendor/generated/`** — the default registry, codegen'd from the vendored
  grammars by `gen-registry.mjs`.

Restore is idempotent (`.stamp` fingerprints per runtime pin and per grammar set),
wired into `prepare` (so `yarn install` self-heals) and `prepack` (so the published
tarball ships the full set — consumers install it prebaked and never fetch).

Net effect: the branch reads as **~2k lines of genuinely reviewable diff** — all
the vendored/generated C is out of git entirely — while a `git clone` +
`yarn install` (or an `npm install` of the published tarball) still ends up with
the complete set.

#### New scripts / infra

- **`vendor/vendor-grammars.mjs`** — one idempotent restore command. Fetches +
  sha256-verifies the runtime from the pinned tarball, copies each grammar's
  minimal source set (only `parser.c`, `scanner.c` when present, `tree_sitter/*.h`,
  `highlights.scm`, `LICENSE`) from the grammar devDependencies, and regenerates
  the default registry. Separate `.stamp` fingerprints (runtime ref+sha; pinned
  grammar versions) make repeated `yarn install`s a no-op — no re-fetch, no
  rewrite of 177 MB. `--force` rebuilds regardless; `--only a,b` scopes to specific
  grammars; `--runtime-src <lib>` vendors the runtime from a local checkout offline
  instead of the network.
- **`vendor/gen-registry.mjs`** — offline, build-time codegen. Emits
  `generated_queries.h` + `generated_registry.cpp` for exactly the selected
  language subset, so the link step is valid no matter how many grammars are
  vendored. Invoked by the Android CMake configure step, the iOS podspec, and by
  `vendor-grammars.mjs` to (re)generate the default set. Resolves
  `; inherits:` directives (e.g. cpp inherits c) by inlining parent highlights.
- **`vendor/grammar-versions.json`** — single pin manifest: the tree-sitter
  runtime ref (`v0.26.3`, ABI 13-15) + tarball URL + sha256, and every grammar's
  package/version/scanner flags plus which form the curated default set.
- **`scripts/prepare-npm-publish.sh`** — `prepack` restores the whole vendor tree,
  copies `core/cpp` into the RN package, and ships `gen-registry.mjs` +
  `grammar-versions.json` alongside so a consumer can compile a custom language
  set; `postpack` restores the symlink and cleans up.
- **`prepare` hook** now runs `vendor-grammars:install` before the bob build, so a
  fresh `yarn install` self-heals the entire gitignored vendor tree.

Consumers can pick their language set (and shrink the binary) via
`ENRICHED_MARKDOWN_CODE_HIGHLIGHT_LANGUAGES` (Podfile ENV) /
`enrichedMarkdown.codeHighlightLanguages` (gradle.properties) / the Expo config
plugin - see `docs/CODE_HIGHLIGHT.md`.

### Testing

Quick summary of what was verified:

- Fresh `yarn install` regenerates the gitignored `vendor/grammars/` from the
  devDependencies; a second install is a `.stamp` no-op.
- `gen-registry.mjs` produces a registry for both the default set and an arbitrary
  `--only`/`--languages` subset; the link step stays valid.
- Highlighting renders correctly in the example app on iOS and Android for the
  default languages, and respects `codeBlock.syntaxColors`.
- Language-subset overrides (Podfile ENV / gradle property / config plugin) build
  and drop the unselected grammars from the binary.
- `yarn workspace react-native-enriched-markdown pack` produces a tarball that
  **contains** the full vendored grammar set (prepack restore), while the git
  working tree does not — verified by installing the tarball into a scratch RN app
  and comparing sizes.
  
### Screenshots

| Case | iOS | Android |
| - | - | - |
| default | <video width="300" src="https://github.com/user-attachments/assets/18ee5b83-3a9a-4b66-b9d7-96d0b2004371" /> | <video width="300" src="https://github.com/user-attachments/assets/dbbd660c-5b24-4c85-89c6-ae618303a2e2" /> |
| selected few | <video width="300" src="https://github.com/user-attachments/assets/3e820991-c147-4db0-8606-28b9a9b525b6" /> | <video width="300" src="https://github.com/user-attachments/assets/72d1e468-cd14-43a0-9ef2-58b5acabdb51" /> |
| all | <video width="300" src="https://github.com/user-attachments/assets/2d63d27e-053a-42dc-a57e-2cc3495b1291" /> | <video width="300" src="https://github.com/user-attachments/assets/f4c8f467-eb47-4eaf-90af-9901554d5fe8" /> |
| disabled | <video width="300" src="https://github.com/user-attachments/assets/815fdf7a-542c-446e-916b-bda3e7ccb8d7" /> | <video width="300" src="https://github.com/user-attachments/assets/65c62ab1-b73e-49ba-8bac-0ba1025d4758" /> |

### PR Checklist

- [ ] E2E tests are passing

# Code-highlighting size impact

| Measure | Size |
|---|---|
| **node_modules** (installed package) | **177 MB** — of which grammars are 168 MB |
| **Published tarball** (npm download) | **9.5 MB** |

App binary - native lib `libreact_codegen_EnrichedMarkdownTextSpec.so` (arm64-v8a, stripped,
uncompressed), which carries all the highlighting code; normalized to the no-highlighting build:

| Config | Native lib | Net increase |
|---|---|---|
| **No highlighting** | 2.2 MB | — (baseline) |
| **Base** (14 default langs) | 10.6 MB | **+8.3 MB** |
| **All** (19 langs) | 25.5 MB | **+23.2 MB** |


docs/CODE_HIGHLIGHT.md
```diff
@@ -0,0 +1,146 @@
...
+| Default (on) | Opt-in (heavier) |
+|---|---|
+| json, html, css, markdown, yaml, go, java, javascript, python, c, rust, bash, typescript, tsx | cpp, swift, php, ruby, c-sharp |
+
+The default set is the smaller-footprint tier (~32 MB of grammar C source across the whole set).
+The opt-in grammars are larger (17-29 MB each) and are only compiled when you list them explicitly.
+A block whose language is not compiled in simply renders as plain (uncolored) code.
+
+### TODO: Kotlin
+
```
hryhoriiK97: Do we need to keep this information? 🤔

hryhoriiK97 (approved):

PR #641, opened 2026-08-07
fix: list marker dropped when inserting character

### LINKED ISSUE #630: iOS/Android: list block dropped when inserting a character at the start of a list item
**Describe the bug**
Inserting a character at the very first position of a list item's line drops the entire list block. The bullet/number marker stops drawing (a visual-only indent remains, with no real spaces or tabs in the text), the preview renders the line as a plain paragraph, and `getMarkdown()` loses the list marker.

This is pre-existing on `main` (not introduced by #612).

**To Reproduce**
1. In an `EnrichedMarkdownTextInput`, create a list item: `- test`
2. Place the caret after the first character: `t|est`
3. Delete the `t` -> `|est` (still a correct list item, marker intact, as expected)
4. Type any character before `e`, e.g. `a`
5. See the marker vanish; `getMarkdown()` returns `aest` instead of `- aest`

**Expected behavior**
The line stays a list item: marker keeps drawing and `getMarkdown()` returns `- aest`.

**Actual behavior**
- The marker disappears, replaced by a visual-only indent (no real whitespace in the buffer)
- The preview renders the line as a plain paragraph
- `getMarkdown()` returns `aest` (the block range is gone)

**Root cause (iOS)**
In `ENRMEditPipeline.adjustStoresForEditAtLocation` (`ENRMEditPipeline.mm:107`) three passes run in order on the insert:
1. `blockStore.adjustForEditAtLocation` -> insert-only branch of `ENRMAdjustRangeForEdit` (`ENRMRangeEditAdjustment.mm:79-84`): because `rangeStart(0) >= editLocation(0)`, the block **shifts** `[0,3) -> [1,4)` instead of growing, so its `location` leaves the line start.
2. `pruneOrphanedBlockAnchors` (`ENRMEditPipeline.mm:119-132`) runs **before** normalize: list items are persists-when-empty (`ENRMInputBlockType.h:72`), so it checks `atLineStart` — `paragraphStart(0) == location(1)` is false -> it **removes the block**.
3. `normalizeToLineBoundsInText` runs last and would have re-anchored the block to `[0,4)`, but it is already gone.

The trigger is narrow: only a pure insert at position 0 of the line pushes `location` off the line start into the prune kill-zone. Android has the same shape (insert shift at `RangeEditAdjustment.kt:96-99` plus a `pruneOrphanedAnchors` pass) and should be confirmed on device.

**Proposed fix**
Keep a block anchored on a leading insert: extend the "owns its start" invariant (blocks already pass `inheritsReplacementAtStart = YES`) into the insert-only branch so a leading insert **grows** the block instead of shifting it. Prefer a block-only flag (e.g. `growsAtStartOnInsert`) rather than reusing `inheritsReplacementAtStart`, since that flag is also set for inline styles and growing them at start would make a character typed before a bold run become bold — undesirable. Do not reorder prune-after-normalize (prune-before-normalize is intentional for line-merge cases). A visual/screenshot regression test is needed (serialization-only coverage would not catch the intermediate state).

 - Device: iOS Simulator (iPhone) and Android emulator
 - OS: iOS + Android
 - Version: react-native-enriched-markdown 0.7.4 (also reproduces on `main`)

**Additional context**
Found during review of #612. Out of scope for that PR, which addresses a different selection-replacement case; this bug is independent and pre-existing.

eszlamczyk:
### What/Why?

Fixes #630. 

Inserting a character at the very first position of a list item's line dropped the whole list block - the marker stopped drawing and `getMarkdown()` lost the marker (`- aest` -> `aest`). 

A leading insert **shifted** the block off the line start (`[0,N) -> [1,N+1)`), so the orphan-prune pass (which runs before normalize) saw the anchor was no longer at a line start and removed the block. Fix: a leading insert now **grows** the block instead of shifting it, keeping it anchored. Uses a new block-only `growsAtStartOnInsert` flag rather than reusing `inheritsReplacementAtStart` (also set for inline styles - growing those would make a char typed before a bold run become bold). Patched on iOS (`ENRMRangeEditAdjustment`) and Android (`RangeEditAdjustment`).

### Testing

Go to playground, set markdown to `- est`, caret to position 0, type `a`, assert the editor still renders a bulleted `aest`. 

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/55a80c3a-b550-4e43-a746-3a7d3d6bdbbe" width=300 /> | <video src="https://github.com/user-attachments/assets/77c7b2e7-d821-413f-b647-2b811cafbb29" width=300 /> |

  ### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
+++ b/packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
@@ -158,7 +158,7 @@ static NSRange paragraphBoundsForRange(NSRange range, NSString *text)
     }
 
     ENRMAdjustedRange adjusted =
-        ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength, YES);
+        ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength, YES, YES);
     if (adjusted.shouldRemove) {
       // A persisting block deleted exactly to its end collapses to a zero-length
       // anchor (the line's newline survived, so the line stays the block); a
--- a/packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
+++ b/packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
@@ -157,8 +157,9 @@ static NSRange paragraphBoundsForRange(NSRange range, NSString *text)
       continue;
     }
 
-    ENRMAdjustedRange adjusted =
-        ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength, YES, YES);
+    ENRMAdjustedRange adjusted = ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength,
+                                                        /* inheritsReplacementAtStart */ YES,
+                                                        /* growsAtStartOnInsert */ YES);
     if (adjusted.shouldRemove) {
       // A persisting block deleted exactly to its end collapses to a zero-length
       // anchor (the line's newline survived, so the line stays the block); a
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/formatting/BlockStore.kt
@@ -116,7 +116,14 @@ class BlockStore {
         it.type in BlockType.ANCHORED && it.start >= editLocation && it.end == deleteEnd
       }
 
-    RangeEditAdjustment.adjustForEdit(ranges, editLocation, deletedLength, insertedLength) { true }
+    RangeEditAdjustment.adjustForEdit(
+      ranges,
+      editLocation,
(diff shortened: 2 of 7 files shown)


packages/react-native-enriched-markdown/ios/input/ENRMBlockStore.mm
```diff
@@ -158,7 +158,7 @@ - (void)adjustForEditAtLocation:(NSUInteger)editLocation
     }
 
     ENRMAdjustedRange adjusted =
-        ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength, YES);
+        ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength, YES, YES);
```
hryhoriiK97: Nit: two trailing `YES` values with no label - easy to accidentally swap if more params are added. Consider inline comments for clarity:

```objc
ENRMAdjustRangeForEdit(blockRange.range, editLocation, deletedLength, insertedLength,
                       /* inheritsReplacementAtStart */ YES,
                       /* growsAtStartOnInsert */ YES);
```

hryhoriiK97 (approved):

PR #646, opened 2026-08-09
fix(ios): reset typing attributes before importing markdown

hsource:
### What/Why?

When `setValue` reloads markdown via `importMarkdown`, iOS assigns plain text using `textView.text = ...`. UITextView stamps the current `typingAttributes` onto all of that text — including any list paragraph `headIndent` left over from where the cursor was before blur.

If the user blurs while focused in a nested list item, non-list paragraphs incorrectly pick up that list indent after `setValue`, even though the parsed markdown and block store are correct.

Reset base typing attributes before `ENRMSetPlainText` so import only applies block styling from the parsed markdown (via `applyFormatting`), not stale composer state from the previous cursor position.

### Testing

1. Open an editable `EnrichedMarkdownTextInput` with mixed content: a normal paragraph, a nested list, and another normal paragraph.
2. Place the cursor in a nested list item.
3. Blur the field while JS calls `setValue` with the current markdown (e.g. on blur to refresh link formatting).
4. **Before:** non-list paragraphs render with the nested list indent.
5. **After:** non-list paragraphs stay at normal indent; list items still render at the correct depth.

Tested on iOS

| Before | After |
| - | - |
| Before | After |

### PR Checklist

- [X] Code compiles and runs on iOS
- [X] Code compiles and runs on Android
- [X] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
+++ b/packages/react-native-enriched-markdown/ios/input/EnrichedMarkdownTextInput.mm
@@ -666,6 +666,10 @@ static const NSTimeInterval kENRMAtomicSnapPollInterval = 0.1;
   [_editSession enterPhase:ENRMEditPhaseImporting];
 
   [_editSession enterPhase:ENRMEditPhaseFormatting];
+  // UITextView stamps typingAttributes (including list headIndent from the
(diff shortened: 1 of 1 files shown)


eszlamczyk (approved):
hi @hsource thanks for your contribution! Change looks good, this should fit into today's nightly release

PR #648, opened 2026-08-10
feat: codeblock streaming

eszlamczyk:
### What/Why?

Adds **streaming support for fenced code blocks** (GitHub flavor), mirroring the existing table/math streaming model, plus a measurement fix that surfaced while building it.

**New `streamingConfig.codeBlockMode` prop** (`'progressive'` default | `'hidden'`), only active with `streamingAnimation`:
- **`progressive`** — the code streams in line-by-line. Its header (language label + copy button) stays visible but **non-interactive** (copy is disabled), and **syntax highlighting is deferred** until the closing fence arrives, so it applies once instead of flickering on every token.
- **`hidden`** — the whole block stays hidden until its closing fence arrives, then appears complete (same all-or-nothing
behavior as block math).

### Additional fix
Yoga calls `measureContent` several times per layout pass. The first call did the exact measure and bumped `lastExactMeasurementCounter`; later calls in the same pass took the streaming fast-path and returned the view's *not-yet-committed* (stale) mailbox size, clobbering the fresh measurement. With content that changes every tick (a streaming code block), this froze the committed height permanently → the region below the code block went invisible / twitched. Fixed by caching the last exact size and returning that from the fast-path. This also makes existing
text/table/math streaming measurement strictly more correct.

### Testing
Run example app's streaming test

#### Screenshots
| | iOS | Android |
| - | - | - |
| progresive (default) | <video src="https://github.com/user-attachments/assets/e53369f4-5ded-4f7f-9e88-8f40695aa360" width="300" /> | <video src="https://github.com/user-attachments/assets/43a3664b-8ae5-4d60-9fe1-8e107acb8bf8" width="300" /> |
| hidden | <video src="https://github.com/user-attachments/assets/d50d7a74-c1c2-4681-9134-dda426c0957a" width="300" /> | <video src="https://github.com/user-attachments/assets/60abba49-dd74-4d6d-a657-d92fb5acb079" width="300" /> |

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/StreamingMarkdownFilter.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/StreamingMarkdownFilter.kt
@@ -11,20 +11,11 @@ enum class CodeBlockStreamingMode {
 }
 
 /**
- * Pre-parse filter that hides incomplete trailing tables, block math and
- * fenced code blocks during streaming. A table is considered complete only
- * after a blank separator line follows it; a block math (`$$`) is complete
- * only when a closing `$$` exists; a fenced code block is complete only when
- * a matching closing fence exists.
- *
- * A still-open fenced code block is handled first because its content may
- * contain lines that look like `$$` delimiters or `|` table rows; those must
- * not be treated as pending math/table blocks. In HIDDEN mode the open block
- * is truncated (like math); in PROGRESSIVE mode it is kept verbatim and only
- * the region before it is passed through the math/table filters.
- * endsInsideOpenCodeFence reports whether the (already filtered) markdown ends
- * inside such a block, so the renderer can defer highlighting and header
- * chrome on that trailing block until its closing fence arrives.
+ * Pre-parse filter that hides incomplete trailing tables, block math and code
+ * blocks while streaming. An open fenced code block is handled first so its
+ * body (which may hold `$$` or `|` lines) is not mistaken for a pending
+ * math/table block: HIDDEN truncates it, PROGRESSIVE keeps it and only filters
+ * the region before it. endsInsideOpenCodeFence flags that trailing open block.
  */
 object StreamingMarkdownFilter {
   fun renderableMarkdownForStreaming(
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/StreamingMarkdownFilter.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/StreamingMarkdownFilter.kt
@@ -11,6 +11,15 @@ enum class CodeBlockStreamingMode {
 }
 
 /**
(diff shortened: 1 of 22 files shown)


eszlamczyk:
Fixed streaming highlighting:


https://github.com/user-attachments/assets/fa50bf49-8481-454f-8b94-c45548755d30

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/StreamingMarkdownFilter.kt
```diff
@@ -5,23 +5,94 @@ enum class TableStreamingMode {
...
+ * math/table block: HIDDEN truncates it, PROGRESSIVE keeps it and only filters
+ * the region before it. endsInsideOpenCodeFence flags that trailing open block.
  */
 object StreamingMarkdownFilter {
   fun renderableMarkdownForStreaming(
     markdown: String,
     tableMode: TableStreamingMode = TableStreamingMode.PROGRESSIVE,
+    codeBlockMode: CodeBlockStreamingMode = CodeBlockStreamingMode.PROGRESSIVE,
+  ): String {
+    val lines = markdown.split("\n")
```
hryhoriiK97, line 35: The markdown is split into lines 3 times per streaming tick: here, again inside `removePendingTablesAndMath` (same string in the no-open-fence case), and once more in `endsInsideOpenCodeFence` (which re-parses to answer a question we already know from `openFenceIndex`). On a hot path this adds up. Consider returning a small result object like `StreamingFilterResult(markdown, endsInsideOpenCodeFence)` to eliminate both redundant splits. Same applies to the iOS side.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/views/CodeBlockContainerView.kt
```diff
@@ -230,14 +241,15 @@ class CodeBlockContainerView(
   }
 
   private fun showContextMenu(anchor: View) {
+    if (pending) return
```
hryhoriiK97: nit: When `pending` is true, `showContextMenu` returns early but the long-press listener still returns `true` and consumes the event - the user gets no feedback at all. Might be worth either not consuming it or giving some visual hint that the block isn't interactive yet.

packages/react-native-enriched-markdown/ios/internals/EnrichedMarkdownShadowNode.h
```diff
@@ -24,7 +24,8 @@ class EnrichedMarkdownShadowNode : public ConcreteViewShadowNode<EnrichedMarkdow
 
   Size measureContent(const LayoutContext &layoutContext, const LayoutConstraints &layoutConstraints) const override;
 
-  static ShadowNodeTraits BaseTraits() {
+  static ShadowNodeTraits BaseTraits()
+  {
```
hryhoriiK97, line 28: Not related change - please revert.

hryhoriiK97 (approved):

PR #651, opened 2026-08-11
feat: add enableTaskListItemToggle prop to EnrichedMarkdownText

### LINKED ISSUE #609: Ability to make task list checkboxes non-interactive
My app allows users to edit markdown in a TextInput, and then displays it with EnrichedMarkdownText. I have `flavor="github"` enabled so task lists are supported. When a user taps a task list checkbox in EnrichedMarkdownText, it toggles. However, this is a read-only view, so I don't want the checkbox to toggle.

I would like a prop on EnrichedMarkdownText to indicate whether task list item toggling is enabled or not. It can default to enabled to match current behavior. I would suggest the name `enableTaskListItemToggle` to match existing prop `enableLinkPreview`.

I looked into whether `onTaskListItemPress` could cancel the toggle by returning a value or calling a function, but it does not seem to do so.

I used a coding agent to patch `react-native-enriched-markdown` to add this flag, and it seems to be a pretty simple addition. Because I am not a past `react-native-enriched-markdown` contributor I don't have a high degree of confidence in my assessment of the fix. If you would like me to contribute this patch as a PR, I'd be happy to do so. Links to the parts of the patch that change the behavior (the rest of the patch is plumbing):

- Android: https://gitlab.com/needbee/docapp-client/-/blob/main/patches/react-native-enriched-markdown@0.7.3.patch?ref_type=heads#L190-192
- iOS: https://gitlab.com/needbee/docapp-client/-/blob/main/patches/react-native-enriched-markdown@0.7.3.patch?ref_type=heads#L273-275

**Additional context**
n/a

CodingItWrong:
### What/Why?

Adds an `enableTaskListItemToggle` prop to `EnrichedMarkdownText` and `EnrichedMarkdown`, so apps can render GFM task list checkboxes that display their markdown state but cannot be toggled by the user.

Closes #609.

The prop defaults to `true`, so existing behavior is unchanged. When `false`, the tap is **fully inert** — no visual toggle and no `onTaskListItemPress` emission — while text selection and links in the same row keep working.

Per platform:

- **iOS** (`EnrichedMarkdown.mm`, `EnrichedMarkdownText.mm`) — the tap handler short-circuits before `handleTaskListTapWithSharedLogic`, so the gesture is never consumed and falls through to the text view unchanged.
- **Android** (`CheckboxTouchHelper`) — `onTouchEvent` returns `false` immediately when disabled, so checkbox hit-testing is skipped and no gesture is consumed.
- **Web** (`ListRenderers.tsx`) — the checkbox is marked `readOnly` / `aria-disabled` rather than `disabled`, keeping its normal appearance so it stays visually consistent with iOS and Android, and it remains keyboard-focusable. `handleChange` guards against toggling, and the checkbox is additionally pointer-inert (`pointer-events: none`) so the browser paints no hover or active state on a checkbox that cannot be toggled.

This addresses both review notes from the issue:

1. **Web support** — implemented in `src/web/renderers/ListRenderers.tsx`, so platforms stay consistent.
2. **Docs** — the fully-inert contract is stated explicitly in the JSDoc on both the native (`MarkdownTextProps.ts`) and web (`MarkdownTextProps.web.ts`) prop types, and the prop is documented in `docs/API_REFERENCE.md` and `docs/WEB.md` (plus `docs/TEXT.md`).

Not included: the optional Storybook story/control, which you offered to add during review. Happy to take that on here instead if you'd prefer.

### Testing

Tested manually in both example projects — `apps/react-native-example` on iOS and Android, and `apps/react-native-web-example` on web.

`yarn typecheck` and `yarn lint` both pass for the changed files.

**On E2E coverage** — no Maestro flow is included, and I'd appreciate your steer before adding one. Two things made me want to check first rather than guess at a pattern:

- Task list checkboxes are drawn spans rather than views, so they carry no `testID`. Every existing flow taps by ID, and none use coordinate/point taps, so exercising a checkbox tap would introduce a new pattern to the suite.
- Flipping the prop needs an affordance in the example app that doesn't exist yet. Nothing currently shipped in `apps/react-native-example` lets you switch `enableTaskListItemToggle` at runtime, so a flow can't reach the disabled state at all until one is added — presumably the Playground control or Storybook story mentioned above. The manual testing above used a throwaway toggle that isn't part of this PR.

The assertion itself looks straightforward, since Playground already surfaces an `Alert` on `onTaskListItemPress` (`assertVisible` when enabled, `assertNotVisible` when disabled). Happy to add the flow, plus iOS/Android screenshot baselines, in whatever shape you'd like — just didn't want to invent a convention here unilaterally.

### PR Checklist

- [ ] E2E tests are passing
  - I did try the existing suite locally, without a usable result:
    - **Android — 1/92 passed.** The emulator drops offline partway through (`device offline` in 83 of the 91 failures); the remaining 8 time out waiting for the app's first screen. Reproducible across runs, including on a single isolated flow.
    - **iOS — 0/94 passed.** `setup-ios-simulator.sh` pins `IOS_VERSION="26.5"`, and that runtime was not downloadable when I tried:

      ```
      $ xcodebuild -downloadPlatform iOS -buildVersion 26.5
      Finding content...
      iOS 26.5 is not available for download.
      ```

      Pointing it at the 26.4 runtime I do have, the flows that ran failed only on the pixel-exact screenshot assertions (`thresholdPercentage: 100`) at 97.2–98.7% similarity, which looks like the runtime delta rather than a rendering change.
  - Failures sit mostly in `enrichedMarkdownInput` flows this PR doesn't touch, so both read as environmental. Relying on CI for the real signal.

---

**Note on authorship:** the code in this PR was authored by Claude. I reviewed it myself and tested it by hand on iOS, Android, and web.


eszlamczyk (approved):
Hi @CodingItWrong everything looks great! 

Thanks for your contribution!

PR #656, opened 2026-08-12
feat(ios): put styled HTML on the clipboard when copying

maksg:
System Copy now writes a public.html flavor alongside plain text, so pasting into rich-text targets keeps headings, inline styles, lists, blockquotes, code blocks, links, and images. The HTML is generated by walking the custom rendering attributes and emitting semantic tags with inline CSS derived from the style config.

The system attributed-string HTML exporter was deliberately not used: it drops image URLs, loses list markers and block chrome (drawn by decoration views, absent from the text), and emits unstable CSS.

The pasteboard is injectable on MarkdownTextView because headless test processes are not authorized to access UIPasteboard.general.

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/ios-enriched-markdown/Sources/EnrichedMarkdown/Extraction/MarkdownHTMLGenerator.swift
+++ b/packages/ios-enriched-markdown/Sources/EnrichedMarkdown/Extraction/MarkdownHTMLGenerator.swift
@@ -1,0 +1,570 @@
+import UIKit
+
+/// Generates semantic HTML with inline styles from the rendered attributed
+/// string by walking the custom `MarkdownAttribute` keys. Used by Copy to
+/// put a rich `public.html` flavor on the pasteboard alongside plain text.
+///
+/// The system attributed-string HTML exporter is deliberately not used: it
+/// drops image URLs, loses list markers and block chrome (drawn by
+/// decoration views, absent from the text), and emits unstable Apple CSS.
+///
+/// Hard line breaks (U+2028 within a paragraph) emit `<br>`.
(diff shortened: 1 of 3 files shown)


packages/ios-enriched-markdown/Sources/EnrichedMarkdown/Extraction/MarkdownHTMLGenerator.swift
```diff
@@ -0,0 +1,549 @@
...
+        case .normal:
+            closeCodeBlockIfOpen(&html, state: &state, styles: styles)
+            closeAllBlockquotes(&html, state: &state)
+            closeListsIfOpen(&html, state: &state)
+            html += "<p style=\"margin: 0 0 \(styles.paragraphMarginBottom)px 0; "
+                + "color: \(styles.paragraphColor); "
+                + "font-size: \(styles.paragraphFontSize)px;\">\(inline)</p>"
+            state.previousWasBlockquote = false
+        }
+    }
```
eszlamczyk, line 172: Nit: I don't like the fact that list and blockquote have neat function calls that look clean while other stuff does not. If possible extract the html generation in heading and normal to separate helpers

eszlamczyk (approved):

packages/ios-enriched-markdown/Sources/EnrichedMarkdown/Extraction/MarkdownHTMLGenerator.swift
maksg: Agreed. It's done now

PR #664, opened 2026-08-14
fix: inputs line height adjustments 

### LINKED ISSUE #615: Android: editor lineHeight broken — render mixes dp/px AND auto-grow measurement ignores it
**Describe the bug**

On **Android**, supplying a `lineHeight` to the editor's `style` is broken in two coupled ways. Together they make multi-line editing unusable when `lineHeight` is set — the same "line-height collapses / text is compressed" symptom from #359 (closed once the `lineHeight` prop was added), plus an auto-grow problem. iOS is unaffected in both cases.

There are **two separate code paths** for line height on Android — the render path and the measurement (auto-grow) path — and they disagree.

---

### Bug 1 — render: `setLineHeight` mixes dp and px

`EnrichedMarkdownTextInputManager.setLineHeight` computes the extra spacing as `value - view.textSize`:

```kotlin
// android/.../input/EnrichedMarkdownTextInputManager.kt
@ReactProp(name = "lineHeight", defaultFloat = 0f)
override fun setLineHeight(view: EnrichedMarkdownTextInputView?, value: Float) {
  if (value > 0 && view != null) {
    view.setLineSpacing(value - view.textSize, 1f)   // unit mismatch
  }
}
```

`value` (the RN `lineHeight`) arrives as a **raw dp** number — `@ReactProp` floats are not density-converted. But `setLineSpacing(add, mult)` and `view.textSize` are both in **px**. So:

- `lineHeight = 24` (dp) → passed as `24f`
- `view.textSize` ≈ `fontSize(16) × density(2.75)` ≈ `44px`
- `setLineSpacing(24 − 44, 1f)` → `setLineSpacing(-20px, 1f)` → **negative** spacing → lines overlap.

**Fix:** convert dp → px before subtracting:

```kotlin
import com.facebook.react.uimanager.PixelUtil
view.setLineSpacing(PixelUtil.toPixelFromDIP(value) - view.textSize, 1f)
```

### Bug 2 — measure: auto-grow ignores `lineHeight`, so the input grows too slowly

Once bug 1 is fixed (render now adds correct positive spacing), a second bug surfaces: the height the editor reports for auto-grow is measured **without** that spacing, so the input grows slower than its content and the content scrolls internally instead of expanding.

The measurement path builds its `StaticLayout` with a hardcoded `setLineSpacing(0f, 1f)`:

```kotlin
// android/.../input/layout/InputMeasurementStore.kt  (measure())
StaticLayout.Builder
  .obtain(content, 0, content.length, paint, widthPx)
  .setIncludePad(true)
  .setLineSpacing(0f, 1f)   // ignores lineHeight entirely
```

So the reported height counts each line without the extra spacing the view actually renders. Per line it falls short by `(lineHeight − textSize)`, and the shortfall accumulates as lines are added.

**Fix:** measure with the same spacing the view renders. The cleanest way to guarantee the two paths never diverge is to use the view as the single source of truth — `InputLayoutManager` already has the view when it triggers a re-measure, so pass `view.lineSpacingExtra` / `view.lineSpacingMultiplier` into `InputMeasurementStore.store()` and use them in `measure()` instead of the hardcoded `(0f, 1f)`. For the first-pass `initialMeasure` (before the view exists) derive the spacing from props using the same `PixelUtil.toPixelFromDIP(lineHeight) - textSize` formula as `setLineHeight`.

---

**To Reproduce**

1. Render `EnrichedMarkdownTextInput` on Android with a base `style` like `{ fontSize: 16, lineHeight: 24 }`.
2. Type several lines separated by Enter.
3. Before bug 1 is fixed: lines overlap. After fixing only bug 1: spacing is correct but the input grows too slowly and its content becomes internally scrollable. iOS renders correctly in both cases.

**Expected behavior**

`lineHeight: 24` with `fontSize: 16` on Android should render non-overlapping lines **and** auto-grow at the correct rate, matching iOS and matching the read-only renderer.

**Environment**

- `react-native-enriched-markdown` **0.7.4**
- `react-native` 0.81.5, Expo SDK 54
- Platform: **Android** (iOS unaffected)

Happy to open a PR covering both — render (dp→px) plus the view-sourced measurement.

eszlamczyk:
### What/Why?
Fixes #615 

The editor's base `style.lineHeight` on `EnrichedMarkdownTextInput`.

 ##### Android — two coupled bugs (the issue)
 1. *Render:* `setLineHeight` computed `value - view.textSize`, mixing a raw **dp**
    prop with a **px** text size, producing negative spacing → overlapping lines.
    Now converts via `PixelUtil.toPixelFromDIP` and centralizes spacing on a stored
    base font size (`applyLineHeight()`), so it's immune to prop-apply order and the
    transient heading cursor-size sync. Also adds the missing `invalidateLayout()`.
  2. *Measure (auto-grow):* the `StaticLayout` used for height hardcoded
    `setLineSpacing(0f, 1f)`, so the input grew slower than its content and scrolled
    internally. Measurement now uses the view as the single source of truth
    (`view.lineSpacingExtra/Multiplier`), cached across width-change re-measures, and
    derives spacing from props for the pre-view first pass.

  ##### iOS — was unsupported entirely 
the `lineHeight` prop reached native but nothing
  consumed it (only `markdownStyle.*.lineHeight` worked, and only for the read-only
  renderer). Implemented it for the editor via `NSParagraphStyle`:
  - New `baseLineHeight` on `ENRMInputFormatterStyle`, read in `applyInputStyleProps`.
  - Applied to existing text through a new formatter pass (`applyBaseLineHeightToTextView`)
    that runs after the block pass so it also covers plain paragraphs, and to typed
    text via typing attributes (`resetBaseTypingAttributes`, `syncWithCursorBlock`,
    `clearListParagraphStyle`).
  - Parity with Android's compression: sets `maximumLineHeight` (clamp) for body text
    so values below the natural line height tighten the lines, while leaving oversized
    lines (headings, attachments) unclamped so they're never clipped.
  - Fixed a `updateProps` gate that only re-formatted when inline markdown ranges
    existed, so runtime `style.lineHeight`/`fontSize`/color changes now apply to plain
    text too.

  Auto-grow needs no extra work on iOS — `measureSize` measures the attributed text,
  which now carries the paragraph line height.

  ### Testing

1. Created throwaway screen in example app. It renders a `fontSize 16 / lineHeight 24` input with multi-line `defaultValue` in a bordered box, plus a no-`lineHeight` control and +/-4 steppers.
2. Verified on **both platforms**:
   - Even, non-overlapping lines and correct auto-grow (no internal scroll).
   - Steppers change spacing live; values below ~16 compress body text.
   - Regression: no `lineHeight` set behaves as before; headings on empty lines
     still size the cursor correctly and aren't clipped by a small `lineHeight`.

#### Screenshots

 | | Before | After |
 | - | - | - |
 | ios | <video src="https://github.com/user-attachments/assets/2134e429-17aa-457f-8d60-2e91f6274dc6" width=300 /> | <video src="https://github.com/user-attachments/assets/4ee2d5f9-bacc-4124-b556-a075a71643b1" width=300 /> |
 | android | <video src="https://github.com/user-attachments/assets/aaa32570-4033-48e4-982b-2ecb5a8234bb" width=300 /> | <video src="https://github.com/user-attachments/assets/1c493b31-e647-4d4d-a647-896bece4dfd6" width=300 /> |

  ### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputManager.kt
@@ -227,9 +227,7 @@ class EnrichedMarkdownTextInputManager :
     view: EnrichedMarkdownTextInputView?,
     value: Float,
   ) {
-    if (value > 0 && view != null) {
-      view.setLineSpacing(value - view.textSize, 1f)
-    }
+    view?.setLineHeightFromProps(value)
   }
 
   @ReactProp(name = "fontFamily")
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -127,6 +127,9 @@ class EnrichedMarkdownTextInputView(
   private var headingOverrideBaseSizePx: Float? = null
   private var baseHintColor: Int? = null
 
+  private var lineHeightDip: Float = 0f
+  private var baseFontSizePx: Float = ceil(PixelUtil.toPixelFromSP(16f))
+
   // Number of ZWSP empty-list anchors believed live in the buffer. Bumped on insert,
   // made exact on every strip scan (which recounts what it keeps), and reset when the
   // buffer is replaced. Incoming markdown is scrubbed of U+200B at parse time, so in
@@ -950,8 +953,24 @@ class EnrichedMarkdownTextInputView(
   fun setFontSizeFromProps(size: Float) {
     if (size <= 0f) return
     val sizePx = ceil(PixelUtil.toPixelFromSP(size))
+    baseFontSizePx = sizePx
     setTextSize(TypedValue.COMPLEX_UNIT_PX, sizePx)
+    applyLineHeight()
     layoutManager.invalidateLayout()
+  }
+
+  fun setLineHeightFromProps(value: Float) {
+    lineHeightDip = value
+    applyLineHeight()
+    layoutManager.invalidateLayout()
+  }
+
+  private fun applyLineHeight() {
+    if (lineHeightDip > 0f) {
+      setLineSpacing(PixelUtil.toPixelFromDIP(lineHeightDip) - baseFontSizePx, 1f)
+    } else {
+      setLineSpacing(0f, 1f)
+    }
   }
 
   fun setColorFromProps(colorInt: Int?) {
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputLayoutManager.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputLayoutManager.kt
@@ -14,7 +14,8 @@ class InputLayoutManager(
     val text = view.text
     val paint = view.paint
 
-    val needUpdate = InputMeasurementStore.store(view.id, text, paint)
+    val needUpdate =
+      InputMeasurementStore.store(view.id, text, paint, view.lineSpacingExtra, view.lineSpacingMultiplier)
     if (!needUpdate) return
 
     val state = Arguments.createMap()
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/layout/InputMeasurementStore.kt
@@ -23,6 +23,8 @@ object InputMeasurementStore {
     val cachedSize: Long,
     val text: CharSequence?,
     val paintParams: PaintParams,
+    val spacingExtra: Float,
+    val spacingMult: Float,
   )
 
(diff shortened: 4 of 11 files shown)


hsource:
@eszlamczyk and @hryhoriiK97 Thanks for this PR! While working on fixing this internally at Wanderlog, I originally took the same approach but ran into some issues: the inputs had very different height layout behavior compared to base React Native TextInputs, which we were trying to replace. This made it harder to use as drop-in- replacement.

Because of this, I ended up adopting a solution that was more similar to the base React Native code.

See #885 and #886 for the Android and iOS approaches: they use spans to actually use line-heights instead of deriving line spacing for line-height.

I basically (with the help of LLMs) looked at a lot of React Native code for TextInput (EditText) and tried to port their logic over. Would you be potentially interested in using that approach instead?

eszlamczyk:
> @eszlamczyk and @hryhoriiK97 Thanks for this PR! While working on fixing this internally at Wanderlog, I originally took the same approach but ran into some issues: the inputs had very different height layout behavior compared to base React Native TextInputs, which we were trying to replace. This made it harder to use as drop-in- replacement.
> 
> Because of this, I ended up adopting a solution that was more similar to the base React Native code.
> 
> See #885 and #886 for the Android and iOS approaches: they use spans to actually use line-heights instead of deriving line spacing for line-height.
> 
> I basically (with the help of LLMs) looked at a lot of React Native code for TextInput (EditText) and tried to port their logic over. Would you be potentially interested in using that approach instead?

Hi @hsource Thanks for your work, Im gonna look into this now

eszlamczyk:
Closing due to https://github.com/software-mansion/enriched-markdown/pull/885 and https://github.com/software-mansion/enriched-markdown/pull/886

PR #674, opened 2026-08-17
feat: gfm blockquotes

### LINKED ISSUE #608: Rework the way blockquotes are handled for GFM
With current setup we draw background separately from the attributed string. We can get better configurability for the blockquote if we move it to the component based setup we are implementing for codeblock in #607

### LINKED ISSUE #621: Blockquotes "padding right" does not apply on Android
**Describe the bug**
After installing the nightly version (^0.8.0-nightly-20260731-e189caeda) to access the new `padding` and `borderRadius` properties for blockquotes, I saw that the vertical padding was fine but not the padding right on android

**To Reproduce**
Type something inside a blockquote and add character by character until reaching the right and side border.

**Expected behavior**
It should break line and let a padding visible on the right hand side

**Screenshots**

Image

 - Device: Pixel 9 Pro XL
 - OS: Android
 - Version ^0.8.0-nightly-20260731-e189caeda

eszlamczyk:
### What/Why?

Closes #608; Closes #621. 

Reworks GFM blockquotes (`flavor="github"`) from inline spans into a **recursive container view**, attempting to mirror how code blocks already work (#607). 

Previously a top-level blockquote was folded into a single attributed-string segment and drawn with spans so it couldn't carry per-instance state or be extended with refs/richer styling. Now each blockquote is a first-class object:

- New reusable **`ContainerNodeView`** (Android `ViewGroup` / iOS `ENRMContainerNodeView`) that reconciles a list of rendered segments into child views, stacks them vertically with per-segment margins, and reports height. Both the document root and every blockquote use it. This is done mainly to have abstraction on top of `non-leaf AST nodes`
- `BlockquoteContainerView` subclasses it and draws the box itself - accent bar, `backgroundColor`, and `padding`, with `borderRadius` (the accent bar is clipped to the rounded box so its corners follow the radius).
- A blockquote splits its own content into segments recursively, so a code block, table, or nested quote inside a quote becomes a *real nested block component* - not inline text. This is what enables future styling extension and gives real box padding on all sides (fixes the #621 right-padding gap).
- Inner text is rendered as blockquote content (quote font/color/line-height). Paragraphs inside a quote now emit block margins so spacing between blocks matches the top-level document, instead of the old tight single-newline path shared with lists.
- Nested quotes carry no extra vertical margin (only the outermost quote does), matching commonmark.
- Web already rendered recursive `<blockquote>`s; one left-inset parity fix so it matches the native box model.

### Commonmark note

The box rendering of the commonmark path is unchanged, but it is not strictly untouched: `ParagraphRenderer` no longer uses the tight single-newline path for blockquotes (it still does for lists). Paragraphs inside a quote now emit block margins so spacing matches the top-level document. Because the commonmark inline blockquote renderer shares the same `blockquoteDepth` signal, blockquotes rendered inline (e.g. nested inside a list item) pick up the same paragraph spacing. This is intentional - spacing is now consistent with the document.

### Testing
Verified with maestro tests (new test) and with example app's storybook (new stories)

### Screenshots
| | iOS | Android |
| - | - | - |
| Commonmark | image | image |
| GFM | image | image |
| nested GFM | image | image |
| GFM with other block elements | <video src="https://github.com/user-attachments/assets/310d1ee7-96e8-4b2a-9c1b-c2d7805dbf17" width="300" /> | <video src="https://github.com/user-attachments/assets/89c6b379-03c9-4221-a78c-4d6257f21431" width="300" /> |

### PR Checklist

- [x] E2E tests are passing


packages/react-native-enriched-markdown/ios/segments/ENRMBlockquoteContainerView.mm
```diff
@@ -0,0 +1,302 @@
...
+#import "ENRMMathContainerView.h"
+#endif
+
+// Inner content is rendered with trailing margins disabled and the default line
+// break strategy; the quote's own vertical inset provides the bottom gap, and
+// per-segment margins between children still apply (allowTrailingMargin only
+// gates the last child's bottom margin).
+static NSArray<ENRMRenderedSegment *> *ENRMRenderBlockquoteChildren(MarkdownASTNode *node, StyleConfig *config)
+{
+  return ENRMRenderSegmentsFromAST(node, config, /*allowTrailingMargin*/ NO, /*allowFontScaling*/ YES,
```
hryhoriiK97: `allowFontScaling` is hardcoded to `YES` and `lineBreakStrategy` to `NSLineBreakStrategyNone` -- blockquote content will ignore these props if the consumer overrides them. Should thread the actual values through like the root path does. Same `allowFontScaling` hardcoding in `ENRMSegmentHeightMeasurer.mm`.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/renderer/ParagraphRenderer.kt
```diff
@@ -22,8 +22,8 @@ class ParagraphRenderer(
   ) {
     val context = factory.blockStyleContext
 
-    // If nested (e.g., inside a list or blockquote), render content simply with a newline
-    if (context.isInsideBlockElement()) {
+    // Inside a list, render tight with a single newline - lists own their item spacing.
+    if (context.listDepth > 0) {
```
hryhoriiK97, line 26: The old `isInsideBlockElement()` covered both lists and blockquotes with the tight-newline path. Now paragraphs inside a blockquote get block margins instead. This also affects the commonmark inline path where `BlockquoteRenderer` bumps `blockquoteDepth` -- is that intentional? If so the PR description ("commonmark untouched") should be updated.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/span/SpanFlags.kt
```diff
@@ -15,3 +15,16 @@ const val SPAN_FLAGS_EXCLUSIVE_EXCLUSIVE = SpannableString.SPAN_EXCLUSIVE_EXCLUS
...
+ * `SPAN_EXCLUSIVE_EXCLUSIVE` with the maximum span priority, for a block-wide
+ * LineHeightSpan (e.g. a blockquote's uniform line height) that must be iterated
+ * BEFORE the per-block [MarginBottomSpan]s it overlaps. LineHeightSpans are
+ * applied in iteration order and higher priority iterates first, so this span
+ * sets the base line height and the lower-priority margin span then adds its gap
+ * on top instead of being normalized away. Without it a line-height pass applied
+ * after the margins (e.g. in BlockquoteTextRenderer.postProcess) would erase them
+ * and clip the first line's ascent.
+ */
+const val SPAN_FLAGS_LINE_HEIGHT_PRIORITY =
```
hryhoriiK97, line 29: Same value as `SPAN_FLAGS_CONTAINER_BACKGROUND` above. Intentional semantic alias or should they have different priorities?

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -5,457 +5,452 @@ import android.content.res.Configuration
...
-      // A just-closed block has unchanged content, so the reconciler reuses it
-      // without an update; sync pending here to trigger its deferred highlight.
-      segmentViews.forEachIndexed { index, view ->
-        if (view is CodeBlockContainerView) {
-          view.pending = pendingCodeBlockSegment != null && index == segmentViews.size - 1
-        }
-      }
+  private fun applyRenderedSegments(
+    renderedSegments: List<RenderedSegment>,
+    @Suppress("UNUSED_PARAMETER") style: StyleConfig,
```
hryhoriiK97: Nit: `style` is passed from every call site but never read. Either drop it or add a comment on why it's kept.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/BlockquoteContainerView.kt
```diff
@@ -0,0 +1,250 @@
...
+
+  companion object {
+    /**
+     * View-free height of a blockquote node at the given outer content width.
+     * The children are summed at the reduced inner width (outer minus horizontal
+     * inset) and the vertical inset is added; nested quotes recurse through this
+     * same function via SegmentHeightMeasurer. Math child heights use the shared
+     * estimate to avoid a main-thread measure from the layout pass.
+     */
+    fun measureBlockquoteNodeHeight(
```
hryhoriiK97, line 219: Minor: `splitASTIntoSegments` + `render` runs here for measurement and again in `applyBlockquoteNode` for display, so each nesting level renders content twice. Not urgent but could cache the rendered segments so the view path reuses the measurement result.

packages/react-native-enriched-markdown/ios/renderer/ENRMBlockquoteTextRenderer.m
```diff
@@ -0,0 +1,47 @@
...
+- (instancetype)initWithConfig:(StyleConfig *)config
+{
+  self = [super init];
+  if (self) {
+    _config = config;
+  }
+  return self;
+}
+
+- (void)pushOnContext:(RenderContext *)context
```
hryhoriiK97, line 19: Cross-platform note: Android's `push()` bumps `blockquoteDepth` (and pops it), iOS deliberately doesn't (to avoid double-indenting lists). Both work, but paragraph rendering branches differently per platform because of it (`blockquoteDepth > 0` vs `currentBlockType == BlockTypeBlockquote`). Just flagging so it's on the radar for future changes.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/renderer/ParagraphRenderer.kt
eszlamczyk: Right, with commonmark untouched I meant more of a "no visible regression seen", updated the PR desc

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/text/span/SpanFlags.kt
eszlamczyk: Yes this is intentional, they're the same packed value but govern two independent span domains. CONTAINER_BACKGROUND orders BlockquoteSpan against other background/drawing spans, while LINE_HEIGHT_PRIORITY orders the block-wide LineHeightSpan against the per-block MarginBottomSpans. Each need max priority within its own domain, and its already made this way. Keeping the two separate flags documents distinct usage of them.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/BlockquoteContainerView.kt
eszlamczyk: I've added a comment that can be picked uppon while we work on performance improvements

hryhoriiK97 (approved):

Issue #691, opened 2026-08-19
Make admonition title text configurable/localizable


eszlamczyk:
Admonition headers currently render fixed English titles derived from the type (Note, Tip, Important, Warning, Caution). Apps should be able to override or localize these labels, likely via the existing accessibilityLabels-style config. Blocked by #606.

PR #692, opened 2026-08-19
feat: gfm blockquote admonitions

### LINKED ISSUE #606: Add github admonitions to GFM
md4c supports admonitions, like bellow, add them to GFM (first rework how blockquote work

> [!NOTE]
> Useful information that users should know, even when skimming content.

> [!TIP]
> Helpful advice for doing things better or more easily.

> [!IMPORTANT]
> Key information users need to know to achieve their goal.

> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.

> [!CAUTION]
> Advises about risks or negative outcomes of certain actions.

> Test

eszlamczyk:
### What/Why?

Closes #606.

Adds support for **Github-style admonitions / alerts** ( `> [!NOTE]`, `> [!TIP]`, `> [!IMPORTANT]`, `> [!WARNING]`, `> [!CAUTION]`), for iOS, Android and Web.

Part of the _blockquote rework_ stack. Reuses previous pr's recursive container-segment rendering, parametrized by node type. An admonition is a blockquote whose first line is an alert marker, so the same container renderer draws the callout box; this PR layers on top the _type-specific_ title, tint color and leading icon.

What it introduces:
- New `admonition` AST node type, plumbed through the md4c parser bridge, the RN spec, and the web parser.
- Shared Octicon icon paths per alert type, rendered natively via svg strings.
- Themable admonition styles (title, tint, icon, container) exposed through `MarkdownStyle` props.

### Testing
- iOS/Android native unit tests updated (so they don't break)
- Added new Storybook entries for the admonitions.
- Verified rendering of all five alert types via example app.

 #### Screenshots

| | iOS | Android |
| - | - | - |
| All variants | image | image |
| Nested | image | image |
| With block elements | image | image |

### PR Checklist

- [x] E2E tests are passing


packages/react-native-enriched-markdown/ios/segments/ENRMSVGPath.m
eszlamczyk, line 1: Note: unfortunately it seems that there isn't any viable options for SVG rendering on ios, and to keep the icons the same created this minimal svg parser

packages/react-native-enriched-markdown/ios/segments/ENRMAdmonitionIcons.m
```diff
@@ -0,0 +1,79 @@
...
+  return data;
+}
+
+CGPathRef ENRMAdmonitionIconPath(NSString *type)
+{
+  NSString *pathData = ENRMAdmonitionIconPathData()[type];
+  if (!pathData) {
+    return NULL;
+  }
+  return ENRMCreateCGPathFromSVGPath(pathData);
```
hryhoriiK97: The icon path gets re-parsed from the SVG string every draw call. Since these 5 icons never change, caching the parsed result per type would be a cheap win. Same on Android in `AdmonitionIcons.kt` — `PathParser.createPathFromPathData()` runs every `onDraw` too.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/parser/MarkdownASTNode.kt
```diff
@@ -38,6 +38,7 @@ data class MarkdownASTNode(
     Subscript,
     Highlight,
     SoftBreak,
+    Admonition,
```
hryhoriiK97, line 42: Looks like `Admonition` is missing from `isTopLevelBlock()` below — it's a top-level block just like `Blockquote`, so `InputParser` won't preserve blank lines around it correctly.

packages/react-native-enriched-markdown/src/web/renderers/BlockRenderers.tsx
```diff
@@ -35,6 +42,52 @@ function BlockquoteRenderer({ node, styles, renderChildren }: RendererProps) {
...
+    alignItems: 'center',
+    gap: Math.round(iconSize * 0.4),
+    marginBottom: Math.round(blockquote.fontSize * 0.4),
+    color: tint,
+    fontWeight: 'bold',
+    fontSize: blockquote.fontSize,
+    lineHeight: 1.2,
+  };
+  return (
+    <div style={boxStyle}>
```
hryhoriiK97: Since this renders as a `<div>` instead of `<blockquote>`, the global `blockquote > *:last-child { margin-bottom: 0 }` rule won't kick in. The last child keeps its bottom margin, so admonitions end up with more bottom space than plain blockquotes.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/BlockquoteContainerView.kt
```diff
@@ -46,6 +49,22 @@ class BlockquoteContainerView(
...
 
+  // Set from the applied node: null for a plain quote, the admonition type
+  // ("note"/"tip"/…) otherwise. Drives the header + per-type theming.
+  private var admonitionType: String? = null
+
+  private val iconSizePx: Int = ceil(blockquoteStyle.fontSize).toInt()
+
+  private val titlePaint =
+    Paint(Paint.ANTI_ALIAS_FLAG).apply {
+      typeface = Typeface.DEFAULT_BOLD
```
hryhoriiK97: Android always uses `Typeface.DEFAULT_BOLD` and macOS always uses `boldSystemFontOfSize:` for the title, so a custom blockquote `fontFamily` won't show up there. iOS (non-macOS) does pick up the configured font.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/BlockquoteContainerView.kt
```diff
@@ -138,6 +182,40 @@ class BlockquoteContainerView(
...
     }
+
+    if (type != null) {
+      drawAdmonitionHeader(canvas, type, tint)
+    }
+  }
+
+  // Draws the admonition header (tinted octicon + capitalized title) in the band
+  // reserved at the top of the view by the enlarged top padding.
+  private fun drawAdmonitionHeader(
```
hryhoriiK97, line 205: The header is drawn directly on the canvas, so VoiceOver/TalkBack won't see it. Might be worth exposing the admonition type ("Note", "Warning", etc.) via `accessibilityLabel`/`contentDescription` on the container.

eszlamczyk: Most of the accessibility stuff is marked to be done in #691

hryhoriiK97 (approved):

PR #699, opened 2026-08-22
fix(android): clean up text render executor

4shutosh:
## Summary

Each `EnrichedMarkdownText` created a background executor that stayed alive after React Native dropped the view. This change shuts it down and clears pending render work and animations.


eszlamczyk (approved):
Hi @4shutosh Thanks for contribution, everything looks good to me!

PR #700, opened 2026-08-24
feat: support HTML line break tags

### LINKED ISSUE #52: Add support for HTML `<br>` tags
## Problem

Currently, `react-native-enriched-markdown` does not support HTML `<br>` tags for line breaks. This makes it difficult to render content that requires explicit line breaks without creating full paragraph breaks (which add extra spacing).

## Use Case

When rendering markdown content (e.g., in chat messages, ticket comments, or user-generated content), we often need to preserve line breaks from the source text.

## Current Behavior

When `<br>` tags are present in the markdown content, they are **treated as plain text** and displayed literally as `<br>` in the rendered output, rather than being converted to actual line breaks.

## Desired Behavior

Support HTML `<br>` tags to render as actual line breaks in the rendered output, similar to how they work in standard HTML and many markdown implementations.

## Technical Context

- **`<br>` is an HTML tag** but is also **commonly used in markdown** for line breaks, as it's a simple and explicit way to create line breaks without the extra spacing of paragraph breaks
- **md4c** (which `react-native-enriched-markdown` likely uses) supports HTML tags including `<br>`
- Many markdown libraries (GitHub Flavored Markdown, etc.) support `<br>` tags as a standard way to create line breaks

## Example

**Input:**
```markdown
Line one<br>Line two<br>Line three
```

**Expected Output:**
```
Line one
Line two
Line three
```

(With line breaks but without the extra spacing of paragraph breaks)

huytdps13400:
### What/Why?

Support attribute-free HTML line break tags (`<br>`, `<br/>`, `<br />`, including case variants) in the shared Markdown parser.

The parser previously disabled all HTML tokenization, so these tags were emitted as literal text. This change keeps HTML blocks disabled, enables inline HTML tokenization, maps only supported `br` tokens to the existing `LineBreak` AST node, and appends every other inline HTML token as literal text. Tags with attributes and tags inside code spans remain literal.

This also covers `br` inside GFM table cells.

Closes #52.

### Testing

- RED/GREEN standalone C++ parser harness, including paragraph, table, unsupported HTML, and code-span cases
- Full iOS `EnrichedMarkdown` XCTest suite on iPhone 17 Pro simulator
- Android parser native build for armv7, arm64, x86, and x86_64
- Android instrumentation-test APK compilation
- Android unit tests and ktlint
- C++ clang-format check
- `yarn lint`
- `yarn typecheck`
- `yarn test` — 6/6 tests
- `yarn prepare`

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/android-enriched-markdown/parser/src/androidTest/java/com/swmansion/enriched/markdown/parser/ParserTest.kt
+++ b/packages/android-enriched-markdown/parser/src/androidTest/java/com/swmansion/enriched/markdown/parser/ParserTest.kt
@@ -131,6 +131,62 @@ class ParserTest {
   }
 
   @Test
+  fun parsesHtmlBreakTagsAsLineBreaks() {
+    for (tag in listOf("<br>", "<br/>", "<br />", "<BR>")) {
+      val ast = requireNotNull(parser.parseMarkdown("line one${tag}line two"))
+      val paragraph = ast.assertHasChildType(MarkdownASTNode.NodeType.Paragraph)
+
+      assertEquals(
+        "failed for $tag",
+        listOf(
+          MarkdownASTNode.NodeType.Text,
+          MarkdownASTNode.NodeType.LineBreak,
+          MarkdownASTNode.NodeType.Text,
+        ),
+        paragraph.children.map { it.type },
+      )
+      assertEquals("line one", paragraph.children.first().content)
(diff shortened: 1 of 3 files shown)


eszlamczyk (commented):
Hi @huytdps13400 we will have to investigate this feature further if this is the way we are 100% sure we want to take, I'll circle to it back in the future

PR #478, opened 2026-07-01
fix: better list children rendering logic

eszlamczyk:
### What/Why?

Fixes #243 — lists containing block children (fenced code blocks, nested sublists) rendered broken: markers
were lost or drawn inside code blocks, checked-task strikethrough bled across code, and code blocks ignored
list indentation.

Root cause: list styling was applied **post-hoc** — children rendered list-ignorant, then `ListItemRenderer`
stamped indent/marker metadata over the result, which broke whenever an item's first child wasn't plain text.
This PR moves list layout to a **top-down** model:

**Shared parser (`packages/core/cpp`)**
- `wrapListItemInlineRuns` post-pass: md4c emits bare inline children for tight list items; they're now wrapped
in synthetic `Paragraph` nodes (marked `tight`), so renderers only ever see block children. One change covers
iOS, Android (JNI) and web (rebuilt `md4c.js`; web renderers needed no changes).

**iOS**
- `BaseRenderer` is now a template method (`renderNode:` = snapshot scope → `renderNodeContent:` → restore).
`RenderContext.snapshotScope/restoreScope` replaces the per-renderer `@try/@finally` partial restores and stops
block-style state leaking between sibling blocks.
- `RenderContext.accumulatedIndent`: `ListItemRenderer` pushes the item indent around its children;
`CodeBlockRenderer` folds it into its paragraph style and `CodeBlockBackground` insets the background rect —
code blocks in list items now render indented under the item content (CommonMark behavior).
- Markers are self-contained `ENRMListMarkerDescriptor`s planted on the item's **first content character**
(appending when one already exists). `ListMarkerDrawer` draws every descriptor at its own gutter — fixes
missing markers on code-block-first items, draws both markers (browser-style) for sublist-first items, and no
longer duplicates markers per paragraph in loose items.

**Android (mirror)**
- `BlockStyleContext.accumulatedIndent` → `CodeBlockSpan.leadingIndent` (leading margin + background inset).
- `BaseListSpan` split into segments around code blocks (keeps the metric-affecting span off code text), single
`drawsMarker` owner per item; when the item starts with a code block/sublist, a zero-margin
`ListMarkerAnchorSpan` draws the marker instead.
- `applyCheckedDecorations` excludes code-block ranges — no more strikethrough across code in checked tasks;
inline-code backgrounds and line-height spans are list-margin aware.

Not in scope (pre-existing): ordered-list `start` is ignored by the parser (`1.` + nested `2. x` renders `1. 1.
x`).

### Testing

[...]

PR #704, opened 2026-08-25
fix(ios): list markers next to emoji not rendered properly

### LINKED ISSUE #702: iOS: emoji dropped from list items + only last item renders a marker (regression since 1.0.0)
## Before submitting a new issue

- [x] I tested using the latest version of the library at the time of reporting (`1.0.1`), and confirmed the bug is present there.
- [x] I checked for possible duplicate issues (searched `bullet`, `marker`, `emoji`) — found none.
- [x] I bisected against the previous stable release (`0.7.4`) and confirmed the bug is **not** present there — this is a regression.

## Bug summary

On iOS (New Architecture / Fabric), rendering an unordered (`-`) list where an item's text starts with an **emoji immediately after the marker** produces two related rendering bugs:

1. **The emoji glyph is dropped entirely** from the rendered list item — even though the underlying markdown/text data still contains it (confirmed by logging the string right before it's handed to the renderer).
2. **Only the last item in the list renders its bullet/marker dot.** Every preceding item in the same list renders no marker at all. This is visible in the fully-settled final render, not just during a transient/streaming state.

Both issues disappear completely when the library is pinned back to `0.7.4` with the exact same markdown input and `list` style props. Removing the emoji (keeping plain text after the marker) does **not** fix bug 2 on `1.0.1` — so this isn't purely an emoji-decoding issue, it's specifically the combination of "marker directly followed by a leading glyph" that breaks marker rendering for anything but the last item.

## Library version

`1.0.1` (bug also present in `1.0.0`; not present in `0.7.4`)

## Environment info

```
React Native: 0.85.3
React: 19.2.3
Expo SDK: ~56.0.15
Node: 22.15.0
Xcode: 26.6 (17F113)
iOS Simulator: iPhone 13 mini, iOS 26.1
Architecture: New Architecture (Fabric) enabled
Platform: iOS only (not verified on Android)
```

## Steps to reproduce

1. Install `react-native-enriched-markdown@1.0.1` on iOS with the New Architecture enabled.
2. Render the following markdown through the component, with a `list` style that sets at least `bulletColor`, `markerColor`, `markerMinWidth`, and `gapWidth`:
   ```md
   **Top categories:**
   - 🛒 Groceries: $320
   - 🚗 Transport: $110
   - 🍔 Dining out: $215
   - 🎬 Entertainment: $85
   - 💊 Health & pharmacy: $60
   ```
3. Observe on the rendered output:
   - The 🛒 🚗 🍔 🎬 💊 glyphs are missing from every item's text.
   - Only the last item ("Health & pharmacy: $60") shows a marker dot; the first four items show no marker at all.
4. Downgrade the dependency to `0.7.4` (no other code changes, same markdown string, same style props) and re-render — both issues are gone: all emoji render, and every item shows its marker.
5. As a secondary check, on `1.0.1`, setting `itemSpacing: 0` (or omitting it) does not affect either symptom — ruling out the new `itemSpacing` ("space between list items") feature as the cause.

## Reproducible example repository

I don't have a public standalone repro repo to share right now (this was found while investigating a regression in an internal app's markdown list rendering), but the snippet above combined with any minimal RN 0.85 + New Architecture app should be enough to reproduce it. Happy to put together a minimal repro repository if that would help narrow it down further — just let me know.

## Additional context

Since `1.0.0` bundles the parser port to iOS, the new theme system, and a list-rendering rewrite all in one release (no intermediate stable npm version between `0.7.4` and `1.0.0`), we weren't able to bisect further via published versions. Given the "marker directly followed by a leading glyph" trigger, our best guess is something in the new iOS list/marker renderer indexing content by UTF-16 code units instead of grapheme-cluster boundaries when locating where the marker attachment/marker range ends and the item's own text begins — but we haven't confirmed this against the `1.0.x` source, just flagging it as a possible lead.

eszlamczyk:
### What/Why?

Fixes #702.

On iOS, an unordered/ordered list item whose text begins with an emoji (e.g. `- 🛒 Groceries`) rendered two bugs, both regressions introduced in 1.0.0:
1. The leading emoji glyph was dropped entirely.
2. Only the **last** item in the list drew its marker; every preceding item drew no bullet/number/checkbox.

##### Root Cause 
In `ListItemRenderer`, each item's marker (`ListItemMarkerStartAttribute`) is anchored on the item's first content character so that items opening with a code block or nested sublist still get a marker (introduced in #478). The anchor was applied over `NSMakeRange(anchorLocation, 1)` - a single UTF-16 code unit. Emoji are multi-unit graphemes (a surrogate pair like 🛒, or longer ZWJ/flag/variation-selector sequences), so a length-1 attribute run splits the grapheme in two. That split makes `NSLayoutManager` fail to form the emoji glyph (dropping it), and shifts the paragraph's first mapped character so the marker drawer's "draw only at paragraph start" check skips every item but the last.

##### Fix 
Anchor the marker over the full composed character sequence instead of one code unit

### Testing
Added a dedicated "Emoji & Grapheme List Rendering" section to the example app's sampleMarkdown.ts (rendered by the Text screen) covering the issue's exact input plus harder graphemes: ZWJ family sequence, regional-indicator flag, and base+variation-selector, across unordered, ordered, nested, and task lists.

### Screenshots

| Before | After |
| - | - |
| image | image |

### PR Checklist

- [x] E2E tests are passing


hryhoriiK97:
@eszlamczyk fyi - pushed a fix in [555c2e5](https://github.com/software-mansion/enriched-markdown/pull/704/commits/555c2e5ac22670f9829b4931863fa2b0d4cc77d7) - ListMarkerDrawer now looks past the leading emoji to use the actual text font's metrics for marker placement.

Zrzut ekranu 2026-08-29 o 12 45 32

hryhoriiK97 (approved):

PR #708, opened 2026-08-25
feat: allow disabling block context menus

### LINKED ISSUE #685: Option to disable long-press copy menu on code blocks and tables
On Android, `selectable={true}` on `EnrichedMarkdownText` makes regular text selectable (native text-selection menu on longpress), but `CodeBlockContainerView.kt` and `TableContainerView.kt` hardcode `setOnLongClickListener { ... showContextMenu(...) }`, which opens the library's custom popup (copy code / copy as markdown, or copy table) instead of the native menu. Code blocks also already expose a copy button in the header (`copyButton`), so for those the longpress popup is redundant with the builtl in copy affordance.

There's no way to disable the popup or fall back to the native selection menu. The `selectable` prop only affects regular text segments.

Would be useful to turn this off per instance.. e.g. a prop on `EnrichedMarkdownText` or a flag in the style object.

Working around it locally with patch-package (stripping the listeners) for now. Happy to open a PR if a prop is the right direction.

Additional notes:

- Math blocks follow the same pattern — `isMathContainerView` also receives `setCopyLabel`/`copyAsMarkdownLabel` and shows the same popup, so a fix should cover tables, code blocks, and math.
- `selectionMenuConfig` labels already apply to the table and math block copy menus, so extending `enabled: false` on its items to hide the whole block popup might be the least new API path.

huytdps13400:
### What/Why?

Fixes #685.

Adds a top-level enableBlockContextMenu boolean for native block views. Setting it to false disables the long-press copy popup on GitHub-flavor code blocks, tables, and block math on iOS, Android, and macOS.

The toggle defaults to true, updates already-mounted block views, and does not affect the code-block header copy button, VoiceOver/TalkBack copy action, or platform text-selection menu. Keeping it outside SelectionMenuConfig matches the existing enable* interaction-toggle API and avoids mixing block interactions with labeled text-selection actions.

### Testing

Using Node 22.20.0 and Yarn 4.11.0:

- TDD regression: the new top-level default test failed before the review fix and now passes
- yarn workspace react-native-enriched-markdown test --runInBand — 8 tests passed
- yarn typecheck
- yarn lint
- yarn prepare — module, types, and Codegen build passed
- pre-commit owner hooks — TypeScript, ESLint, clang-format, ktlint passed
- Android: :react-native-enriched-markdown:compileDebugKotlin passed
- iOS: ReactNativeEnrichedMarkdown arm64 simulator target built successfully with Xcode 26.6
- macOS target attempt is blocked before the library target by the existing RN macOS 0.81.4 fmt 11.0.2 consteval failure on Xcode 26.6

### PR Checklist

- [ ] E2E tests are passing
- [x] Required regression tests have been added


eszlamczyk (changes requested):
Hi @huytdps13400 thanks for tackling this, and for the test coverage. The behaviour looks decent. Although I believe there are some things that needs to be polished:

1. Wrong place in the public API
`SelectionMenuConfig` models the native text-selection menu - the action you get when you select text. Those items are in the shape `{enabled?, label?}` objects specifically because they each carry a localizable label. The block long-press popup is a different interaction (a context menu on non-text block views), and it needs only an on/off. 

2. It's shiped as Android-only, but iOS has the identical popup. 
iOS renders the same long-press copy menu on all three block types via `UIContextMenuInteraction`:
- `ENRMCodeBlockContainerView.m` (`configurationForMenuAtLocation:`)
- `TableContainerView.m`
- `ENRMMathContainerView.m`

So @platform android here documents a gap rather than a platform limitation — iOS already reads every other field of selectionMenuConfig; this would be the one field it parses on Android and silently drops on iOS. A user setting it would reasonably expect it to work on both.

### Requested change
Replace selectionMenuConfig.blockContextMenu with a top-level boolean, matching the existing enable* interaction-toggle convention (enableTaskListItemToggle, enableLinkPreview):
```
/**
 * Controls the long-press context menu on block views (code blocks, tables,
 * and block math). When false, long-pressing a block does not open the copy
 * popup. Does not affect the code-block header copy button, the
 * VoiceOver/TalkBack copy action, or the system text-selection menu.
 * @default true
 * @platform ios, android, macos
 */
enableBlockContextMenu?: boolean;
```

huytdps13400:
Addressed the requested API and platform changes in e89b42a.

The option is now the top-level enableBlockContextMenu boolean, SelectionMenuConfig is back to text-selection actions only, and the native prop updates existing block views as well as newly created code/table/math views. Android returns false from disabled long-press listeners; iOS context-menu delegates return no configuration; macOS menu providers return no menu. Header and accessibility copy paths are unchanged.

Fresh validation on the pushed commit: Jest 8/8, TypeScript, ESLint, package build/Codegen, owner clang-format and ktlint hooks, Android compileDebugKotlin, and the iOS ReactNativeEnrichedMarkdown arm64 simulator target all pass. I also attempted the macOS target; Xcode 26.6 fails earlier in the existing RN macOS 0.81.4 fmt 11.0.2 dependency with consteval errors, before compiling this library.

eszlamczyk (approved):
Looks good! Thanks for your contribution, this should be released with today's nightly release

huytdps13400:
Thank you for the review and for the guidance on the API placement and platform parity. Glad to hear the updated implementation is ready for the nightly release.

hryhoriiK97 (approved):

PR #713, opened 2026-08-26
docs: finished documentation page

eszlamczyk:
This pull request is to track overall documentation implementation

> [!WARNING]
> Merging this pull request will release the documentation page. **This should not be done unless consulted with whole team**


eszlamczyk (commented):
Todo: add information to docs about possibly failing postinstall scripts, that they warn instead of erroring and to be caucious

PR #721, opened 2026-08-27
feat(ios): tap-to-toggle task list checkboxes

maksg:
### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):

hryhoriiK97 (commented):
**Note (non-blocking):** The `taskMarkerRegex` only matches unordered list markers (`-`, `*`, `+`), same as the React Native package's `TaskListTapUtils` on both iOS and Android. GFM also allows ordered-list task items (`1. [ ] task`), which would toggle visually (the hit-test uses attributes, not regex) but `togglingSource` wouldn't rewrite the markdown, leaving source out of sync with the visual state.

This is a pre-existing limitation shared with the RN package, so not a blocker for this PR, but worth tracking as a follow-up for both. cc @eszlamczyk @maksg

hryhoriiK97 (approved):

PR #723, opened 2026-08-28
fix(ios): accessibility path follows correct position in scroll view

### LINKED ISSUE #696: Incorrect accessibilityPath positions in ScrollView (iOS)
**Describe the bug**
In our project we use `<EnrichedMarkdownText />` inside of a `<ScrollView  />` which renders a chat bot conversation. 
When new elements are being rendered in that conversation and you start VoiceOver on iOS the focus ring for the text of `<EnrichedMarkdownText />` are not rendered at the right position and thus not properly being scrolled to when navigating through the elements.

**To Reproduce**
Render `<EnrichedMarkdownText />` inside of a `<ScrollView  />` and dynamically add new `<EnrichedMarkdownText />` elements to the scroll view container.

**Expected behavior**
When the screen reader reads out the content of a text element inside of `<EnrichedMarkdownText />` the focus ring should be located at the right position and it should be scrolled to the right position.

**Screenshots**
Image

**Additional context**
The patch in the attached screenshot actually fixed the issue for us. So I would assume the issue here is that the paths are being cached and not recalculated?

 - Device: iPhone17
 - OS: iOS26
 - Version: 1.0.1

eszlamczyk:
### What/Why?
Fixes #696.

On iOS, VoiceOver drew the focus ring for EnrichedMarkdownText text at the wrong position (and failed to scroll to it) when the component was inside a ScrollView - e.g. a chat feed that appends messages.

Multi-line text ranges described their focus ring with an accessibilityPath in absolute screen coordinates, computed once and cached. Screen coordinates go stale the moment the ScrollView scrolls, so the ring pointed at the text's old position. This drops the screen-space path and describes every accessibility element with `accessibilityFrameInContainerSpace`, which UIKit re-projects to the screen on every query, so the geometry stays correct under scroll. (This matches the fix the reporter proposed.)

> Note: because of this change we changed paragraphs to be one block instead of block per line.

### Testing
- Ran the example app's Text screen (EnrichedMarkdownText inside a ScrollView with long, multi-line content) on iOS. With VoiceOver / Accessibility Inspector, scrolled the content and navigated between paragraphs: the focus ring now sits on the visible text and VoiceOver scrolls each element into view. Verified single-line paragraphs, links, images, and headings still focus correctly.

##### Before
https://github.com/user-attachments/assets/d9dce0ee-3453-42f8-86f5-d2dd6a62094a

##### After 
https://github.com/user-attachments/assets/c1d9dea9-d95c-41d0-8980-5125887cfc72

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/ios/utils/MarkdownAccessibilityElementBuilder.m
+++ b/packages/react-native-enriched-markdown/ios/utils/MarkdownAccessibilityElementBuilder.m
@@ -161,12 +161,7 @@ static const CGFloat kFocusRectPadding = 2.0;
   if (text.length > 0)
     element.accessibilityLabel = text;
 
-  UIBezierPath *path = [self accessibilityPathForRange:range inTextView:textView];
-  if (path) {
-    element.accessibilityPath = path;
-  } else {
-    element.accessibilityFrameInContainerSpace = [self frameForRange:range inTextView:textView container:container];
-  }
+  element.accessibilityFrameInContainerSpace = [self frameForRange:range inTextView:textView container:container];
 
   if (type == ElementTypeImage) {
     element.accessibilityTraits =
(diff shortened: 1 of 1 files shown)


hryhoriiK97 (approved):
LGTM 🎖️

PR #729, opened 2026-08-30
fix: keep pipe tables visible in commonmark

### LINKED ISSUE #645: EnrichedMarkdownText renders nothing when content contains `|` (spoiler syntax cannot be disabled)
## Version

`react-native-enriched-markdown@0.7.4`, React Native 0.8x via Expo SDK 57, Android.

## What happens

Any content containing ASCII pipes renders as a **completely empty view** — no text, no error,
no warning. A markdown table, a shell command, prose using a vertical bar: all vanish.

For a chat app this is silent data loss. A table-only answer from an LLM arrives as a blank
message bubble, and the user has already paid for the tokens that produced it.

## Reproduction

```jsx
// Renders nothing at all
<EnrichedMarkdownText flavor="commonmark">
  {'| Model | Speed |\n|---|---|\n| Claude | Fast |'}
</EnrichedMarkdownText>

// Renders fine
<EnrichedMarkdownText flavor="commonmark">
  {'Model: Claude, Speed: Fast'}
</EnrichedMarkdownText>
```

Measured on a physical device by rendering the raw string beside the markdown:

| Content | Length | Pipes | Rendered |
|---|---|---|---|
| Model comparison, table form | 593 chars | 36 | **nothing** |
| Same content, prose form | 575 chars | 0 | correctly |

Length is not the factor; the pipes are.

## Cause (as far as we could determine)

The `||text||` spoiler syntax. The docs state that spoiler parsing is **always enabled**, and
the available `md4cFlags` cover underline / superscript / subscript / latexMath / highlight —
there is no flag to turn spoilers off. A table row (`| a | b |`) is full of pipe pairs.

Two things we tried that did **not** help, which may narrow it down:

1. Styling the spoiler visible (`spoiler: { color: 'transparent' }`) changes nothing — the
   block appears to be dropped before styling is applied.
2. `flavor="github"` does render tables, but segments the view per block, and the docs note
   that "text selection cannot span across segments" — which reintroduces a different bug for
   us, so it is not a usable workaround.

## What would help

Any one of these:

- An `md4cFlag` to disable spoiler parsing.
- Requiring spoilers to be well-formed, so an unmatched or table-shaped pipe run falls through
  as literal text.
- Failing visibly rather than silently — rendering the raw text, or warning in dev — so this is
  discoverable without a device experiment.

The silent part is what makes it expensive: there is no error to search for, and the code reads
as correct.

## Our workaround

`mobile-rn/src/lib/mdTable.js` converts GFM tables to ASCII-bordered text inside a code fence
and replaces every remaining ASCII pipe with U+2758 (LIGHT VERTICAL BAR) before rendering. Its
test suite asserts "no ASCII pipe reaches the renderer" as the core invariant.

Side effect we accept: a shell command inside a fence shows `❘` instead of `|`. A mangled glyph
beats a blank reply.

huytdps13400:
### What/Why?

`flavor="commonmark"` used the single-TextView renderer, but the shared MD4C parser still enabled the GFM tables extension. A table-shaped string therefore became a `Table` AST node that the CommonMark renderer cannot represent, so the entire message rendered as an empty view.

This change makes table parsing an internal flavor policy:

- CommonMark disables the MD4C tables extension, preserving pipe-table source as selectable literal text.
- GitHub flavor keeps table parsing and the segmented table renderer unchanged.
- The shared native parser exposes the low-level tables flag consistently for Android and iOS consumers.
- Regression coverage checks JS flavor wiring, the shared parser behavior, and an Android Maestro flow against the exact issue input.

Plain prose and shell pipelines containing a single `|` already parsed as text and remain unchanged. Spoiler parsing (`||text||`) also remains enabled and unchanged.

Closes #645.

### Testing

- `yarn lint`
- `yarn typecheck`
- `yarn test --runInBand` — 19/19 passed
- `yarn prepare` — TypeScript, module build, and native Codegen passed
- iOS native package tests via `xcodebuild test` — 274/274 passed
- SwiftLint 0.65.0 — 0 violations
- Android native `testDebugUnitTest`, `ktlintCheck`, `compileDebugAndroidTestKotlin`, and `assembleRelease` — passed
- React Native Android example/Fabric build — passed
- React Native iOS example/Fabric build — passed
- Android Maestro `commonmark_pipe_table_test.yaml` on API 35 — passed; the preview exposed `| Model | Speed | |---|---| | Claude | Fast |`
- Repository pre-commit hooks (`lint`, `typecheck`, `clang-format`, `ktlint`) — passed

### PR Checklist

- [x] E2E tests are passing


.maestro/enrichedMarkdownText/flows/basic/block_elements/commonmark_pipe_table_test.yaml
eszlamczyk, line 1: While Im usually fine with more tests, this change does not affect visual apperance in its spirit. The unit tests in the packages itselves are enough to cover this, so I'd want you to remove this and the `PlaygroundScreen` changes as well.

eszlamczyk (changes requested):
Thanks @huytdps13400 for your pull request!

While this change could work, I believe there is a cleaner way to do it. Since tables are part of the strict GFM specification, gating them behind their own dedicated `tables` flag threaded through the whole stack is unnecessary - and it leaves the mode half-done, since strikethrough and task lists are still parsed unconditionally, so `flavor="commonmark"` isn't really CommonMark.

Instead of putting a flavor-derived value inside the `Md4cFlags` struct, I'd rather thread the information about the flavor being GFM through to the parser and resolve the extension flags there - not just tables, but the whole GFM set (strikethrough, task lists) in one place:

```cpp
std::shared_ptr<MarkdownASTNode> MD4CParser::parse(
    const std::string &markdown, const Md4cFlags &md4cFlags, bool isGFM) {
  unsigned flags = MD_FLAG_NOHTML | MD_FLAG_SPOILERS; 
  if (isGFM) {
    flags |= MD_FLAG_TABLES | MD_FLAG_STRIKETHROUGH | MD_FLAG_TASKLISTS;
  }
  // ...
```
On the JS side this stays what you already have, just as its own signal rather than member of md4cFlags, e.g. isGFM: flavor === 'github'. On iOS and Android you might need to therad this a bit, confirm that you need or reuse already existing infrastructure.

When you're done please tag me here and I'll re-review. Please address my other review comments as well.

.maestro/enrichedMarkdownText/flows/basic/block_elements/commonmark_pipe_table_test.yaml
huytdps13400: Removed the Maestro flow and the PlaygroundScreen additions in 9da8a83. Coverage now stays at the parser/unit layers, as requested.

eszlamczyk (approved):
LGTM - the core change is exactly the shape I wanted: a single `isGFM` signal gating the whole GFM extension set (tables + strikethrough + task lists) in the parser, driven off `flavor === 'github'` at the RN layer, with no `tables` member left in `Md4cFlags`. Threading through the RN iOS and Android paths (including measurement and streaming) is thorough. 

One small nitpick, which I'll take care of myself so we don't hold this up: the standalone native packages (`enriched-markdown-ios`, `android-enriched-markdown`) don't render GFM yet - there's no table/task-list/strikethrough renderer in them - so exposing an `isGFM` toggle on their public parse API (and documenting it in the iOS README) advertises something they can't do. I'll drop the `isGFM` parameter from those two packages and just pin CommonMark at their bridge call, and remove the GFM mention from the iOS README. That keeps GFM entirely a concern of the RN layer until the dedicated iOS GFM rendering work lands. Thanks for the solid rework here!

PR #731, opened 2026-08-30
fix(android): remove the accessibility layout listener on detach

### LINKED ISSUE #730: [Android] EnrichedMarkdownText retains its whole view hierarchy: the accessibility OnGlobalLayoutListener is never removed

- [x] I checked for possible duplicate issues.
- [x] I tested using a supported version of React Native (0.86.2, New Architecture).
- [ ] I tested using the latest version of the library. **Measured on 0.7.4** - but the two files
  involved are byte-identical in 1.0.2, and identical again on `main` at
  00423aa8110a7a69b26b0debb1844b3f7fa60541, so the bug is present in the current code. Details below.

## Bug summary

On Android, an `EnrichedMarkdownText` that is mounted and later unmounted **retains its entire
enclosing view hierarchy, permanently**. Not just the markdown view: because a retained `View`
pins `mParent`, and a parent pins all of its children, one leaked markdown view retains the whole
screen it lived on.

In our app that was ~225 `View` objects plus an unrelated `WebView` per screen open, linear, never
reclaimed. Six open/close cycles of one screen:

| | Views | WebViews |
|---|---|---|
| Baseline | 327 | 0 |
| After 6 cycles | 1907 (+225/cycle) | 7 (+1/cycle) |
| After 6 cycles, with the one-line fix below | **327** | **0** |

All readings on a **release** build (Pixel 11 Pro, Android 16), taken from `dumpsys meminfo`'s
Objects block after forcing a GC by backgrounding the app, so these are retained references and
not uncollected garbage.

## Root cause

`MarkdownAccessibilityHelper` defers its TalkBack item rebuild by registering a
`ViewTreeObserver.OnGlobalLayoutListener`, and that listener is only ever removed by **its own
callback**:

https://github.com/software-mansion/enriched-markdown/blob/00423aa8110a7a69b26b0debb1844b3f7fa60541/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt#L68-L101

Three things combine:

1. **The deferred branch is the normal path, not an edge case.** The JS wrapper always sends
   resolved `accessibilityLabels`, so `setAccessibilityLabels` -> the `labels` setter ->
   `invalidateAccessibilityItems()` runs at prop-set time, when `textView.layout` is still `null`.
   Every instance therefore registers a listener.

2. **The listener ends up on the WINDOW's observer.** At prop-set time the view is not attached
   yet, so `textView.viewTreeObserver` returns the view's own *floating* observer. On attach,
   `View.dispatchAttachedToWindow` **merges** the floating observer's listeners into
   `AttachInfo.mTreeObserver`, which belongs to `ViewRootImpl`. That outlives every screen and is
   a GC root.

3. **Nothing else ever removes it.** `removePendingLayoutListener()` is called from exactly one
   place - inside the listener. The other branch of `invalidateAccessibilityItems()` (the one
   taken once the view *does* have a layout) rebuilds immediately and never clears a listener
   registered earlier, and neither `onDetachedFromWindow` nor
   `EnrichedMarkdownTextManager.onDropViewInstance` clears one.

So every instance whose callback does not fire leaves:

```
ViewRootImpl.mTreeObserver -> OnGlobalLayoutListener -> MarkdownAccessibilityHelper -> the TextView
  -> mParent -> ... -> the entire screen
```

**This is not about rendering.** The clearest evidence: a component rendered with `markdown=""`,
which returns from `setMarkdownContent` before `scheduleRender` and never parses or renders
anything, leaks at exactly the same rate. That rules out the parser, the spans, `MeasurementStore`
and the render executor, and leaves only what happens when the view is constructed and its props
are set.

Two smaller consequences worth noting:

- The per-view `Executors.newSingleThreadExecutor()` in `EnrichedMarkdownText` never dies while a
  view is retained (thread count climbed ~4 per screen open here, and went flat with the fix).
  It is a symptom, not a separate bug - `FinalizableDelegatedExecutorService.finalize()` shuts the
  pool down once the view is collectable - but it does mean a leaked view also costs a thread.
- `MarkdownAccessibilityHelper` is created in the shared base class
  `AccessibleMarkdownTextView`, so **`EnrichedMarkdownInternalText` is affected too**, and with it
  the `EnrichedMarkdown` component.

## Suggested fix

Remove the listener on detach, in the base class so both subclasses are covered.
`onDetachedFromWindow` is the right place because it is the last moment `getViewTreeObserver()`
still returns the window's observer - doing it from `onDropViewInstance` would silently remove
from the view's floating observer instead, and leave the leak in place.

```diff
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
@@
     override fun onFocusChanged(
       gainFocus: Boolean,
       direction: Int,
       previouslyFocusedRect: Rect?,
     ) {
       super.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
       accessibilityHelper.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
     }
+
+    override fun onDetachedFromWindow() {
+      accessibilityHelper.cleanup()
+      super.onDetachedFromWindow()
+    }

--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
@@
+  fun cleanup() {
+    removePendingLayoutListener()
+  }
+
   private fun removePendingLayoutListener() {
```

That is the exact change measured in the table above (carried as a `patch-package` patch in our
app). TalkBack behaviour is unaffected: the listener exists only to rebuild items after a pending
layout, and a detached view has no items to announce.

I am happy to open a PR for this - just say the word, or point me at a different shape if you'd
prefer the cleanup somewhere else.

## Library version

0.7.4 (measured). Present unchanged in 1.0.2 and on `main`:
`MarkdownAccessibilityHelper.kt` is **byte-identical** between 0.7.4 and 1.0.2, and
`onDetachedFromWindow` in `EnrichedMarkdownText.kt` still does not clean up on `main`.

## Environment info

```
React Native 0.86.2 (New Architecture / Fabric)
Expo SDK 57
react-native-enriched-markdown 0.7.4
Android 16 (API 36), Pixel 11 Pro, release build (R8 enabled)
```

## Steps to reproduce

1. Put an `<EnrichedMarkdownText markdown="..." markdownStyle={...} />` anywhere inside a screen
   that can be pushed and popped (a `react-native-screens` stack route, in our case expo-router).
   One instance is enough - no list, no recycling.
2. Push the screen, pop it. Repeat six times.
3. Background the app for ~10s so a GC runs, then
   `adb shell dumpsys meminfo <your.package>` and read `Views` from the **Objects** block.

`Views` climbs by roughly one screen's worth per cycle and never comes back down. Take two samples
a few seconds apart and trust the second - the first can still be pre-collection.

The same procedure with the markdown component swapped for a plain `<Text>` returns to baseline
exactly, which is the A/B that isolates it.

## Reproducible example repository

I do not have a standalone repo to link, and I would rather say so than link something I have not
run. What I do have is the measurement above from a real release build, plus the source reasoning,
which I hope is enough to act on given the fix is one line. The reproduction is small enough that
the example app in this repo should show it by navigating to a markdown screen and back while
watching `dumpsys meminfo` - I have not run that myself, so I am flagging it as derived rather
than verified. If a standalone repro repo would help, I am glad to put one together.

Tyler-V:
### What/Why?

Fixes #730.

On Android, an `EnrichedMarkdownText` that is mounted and later unmounted retains its **entire
enclosing view hierarchy**, permanently. `MarkdownAccessibilityHelper` defers its TalkBack item
rebuild by registering a `ViewTreeObserver.OnGlobalLayoutListener`, and that listener is only ever
removed by its own callback:

https://github.com/software-mansion/enriched-markdown/blob/00423aa8110a7a69b26b0debb1844b3f7fa60541/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt#L78-L101

Three things make that leak once per instance:

1. **The deferred branch is the normal path.** The JS wrapper always sends resolved
   `accessibilityLabels`, so the `labels` setter calls `invalidateAccessibilityItems()` at
   prop-set time, when `textView.layout` is still `null`. Every instance registers a listener.
2. **The listener ends up on the window's observer.** At prop-set time the view is not attached,
   so `textView.viewTreeObserver` is the view's own *floating* observer;
   `View.dispatchAttachedToWindow` then merges it into `AttachInfo.mTreeObserver`, which belongs
   to `ViewRootImpl` and outlives every screen.
3. **Nothing else removes it.** `removePendingLayoutListener()` has exactly one caller, inside the
   listener. The immediate branch of `invalidateAccessibilityItems()` never clears a listener
   registered earlier, and neither detach nor `onDropViewInstance` does.

Result: `ViewRootImpl.mTreeObserver -> listener -> helper -> TextView -> mParent -> the whole
screen`.

`onDetachedFromWindow` is the right place for the removal specifically because
`getViewTreeObserver()` returns the window's observer only while the view is attached. Doing it
from `onDropViewInstance` would silently remove from the view's floating observer and leave the
leak untouched.

The cleanup goes in `AccessibleMarkdownTextView` rather than in `EnrichedMarkdownText`, because
that is where the helper is created, so `EnrichedMarkdownInternalText` (and with it the
`EnrichedMarkdown` component) is covered by the same change.

### Testing

Measured in a production app, on a **release** build with R8 enabled: Pixel 11 Pro, Android 16,
React Native 0.86.2 on the New Architecture, expo-router / react-native-screens navigation.
Six push/pop cycles of one screen, `Views` and `WebViews` read from `dumpsys meminfo`'s Objects
block after forcing a GC by backgrounding the app, two samples per reading:

| Six push/pop cycles | Views | WebViews |
| - | - | - |
| Before | 327 -> 1907 (+225/cycle) | 0 -> 7 |
| After | 327 -> **327** | 0 -> **0** |

The `WebViews` column is not a typo: an unrelated `WebView` elsewhere on the same screen was being
retained through the same chain, which is what makes this expensive rather than merely untidy. At
about 23 MB PSS per retained screen, this was the whole of a leak we had been chasing for two
weeks.

Two A/Bs isolate it to this component and this mechanism, same install and same content:

- Replacing the markdown component with a plain `<Text>` returns to baseline exactly.
- A component rendered with `markdown=""`, which returns from `setMarkdownContent` before
  `scheduleRender` and never parses or renders anything, leaks at the **identical** rate. That
  rules out the parser, the spans, `MeasurementStore` and the render executor, and points at
  construction and prop-set, which is where the listener is registered.

One incidental observation while measuring: the per-view `Executors.newSingleThreadExecutor()` in
`EnrichedMarkdownText` climbed about 4 threads per screen open before this change and goes flat
after it. That is a symptom rather than a second bug (the pool's `finalize()` shuts it down once
the view is collectable), but it does mean a retained view also costs a live thread.

### PR Checklist

- [ ] E2E tests are passing

Being straight about the unticked boxes: the change is Kotlin-only and Android-only, and it was
built, run and measured in a real release app rather than in this repo's example app, so I have
not run the Maestro suites or an iOS build. I did not add an E2E test because the behaviour is not
user-visible - it is an object count in `dumpsys meminfo`, which Maestro cannot assert. Happy to
adjust the shape of this if you would rather the cleanup lived somewhere else, and happy to put
together a standalone reproduction repo if that would help.

### DIFF
--- a/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
+++ b/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
@@ -30,4 +30,14 @@ abstract class AccessibleMarkdownTextView
       super.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
       accessibilityHelper.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
     }
+
+    override fun onDetachedFromWindow() {
+      // Before super, and here rather than in a ViewManager's
+      // onDropViewInstance: getViewTreeObserver() returns the WINDOW's observer
+      // only while the view is attached. Once detached it hands back the view's
+      // own floating observer, so a removal then would target the wrong
+      // observer and leave the listener on the window.
+      accessibilityHelper.cleanup()
+      super.onDetachedFromWindow()
+    }
   }
--- a/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
+++ b/packages/android-enriched-markdown/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
@@ -83,6 +83,17 @@ class MarkdownAccessibilityHelper(
     observer.addOnGlobalLayoutListener(listener)
   }
 
+  /**
+   * Drops the pending global-layout listener, if any is still registered.
+   *
+   * Must run while the view is attached: `textView.viewTreeObserver` returns
+   * the window's observer only until detach, and the listener was merged into
+   * it from the view's floating observer when the view attached.
+   */
+  fun cleanup() {
+    removePendingLayoutListener()
+  }
+
   private fun removePendingLayoutListener() {
     val listener = pendingLayoutListener ?: return
     pendingLayoutListener = null
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/AccessibleMarkdownTextView.kt
@@ -30,4 +30,14 @@ abstract class AccessibleMarkdownTextView
       super.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
       accessibilityHelper.onFocusChanged(gainFocus, direction, previouslyFocusedRect)
     }
+
+    override fun onDetachedFromWindow() {
+      // Before super, and here rather than in a ViewManager's
+      // onDropViewInstance: getViewTreeObserver() returns the WINDOW's observer
+      // only while the view is attached. Once detached it hands back the view's
+      // own floating observer, so a removal then would target the wrong
+      // observer and leave the listener on the window.
+      accessibilityHelper.cleanup()
+      super.onDetachedFromWindow()
+    }
   }
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
@@ -91,6 +91,17 @@ class MarkdownAccessibilityHelper(
     observer.addOnGlobalLayoutListener(listener)
   }
 
+  /**
+   * Drops the pending global-layout listener, if any is still registered.
+   *
+   * Must run while the view is attached: `textView.viewTreeObserver` returns
+   * the window's observer only until detach, and the listener was merged into
+   * it from the view's floating observer when the view attached.
+   */
+  fun cleanup() {
+    removePendingLayoutListener()
+  }
+
   private fun removePendingLayoutListener() {
     val listener = pendingLayoutListener ?: return
     pendingLayoutListener = null


eszlamczyk (approved):
Hi @Tyler-V thanks for the genuinely thorough investigation!

I reproduced the leak and confirmed the fix on our side. Thanks for your contribution.

One small thing for next time: this repo carries a second, parallel native source tree at `packages/android-enriched-markdown/`, and it has the same leaking listener with no detach cleanup, so a change like this generally needs to be mirrored there too. If you could in the future just run a agent to check, wether this fix needs to be applied as well there. I will fix this myself to speed it up, but please be aware in the future.

Thanks again for the high-quality report and fix!

PR #736, opened 2026-09-01
feat(android): add support for strikethrough and underline

jnowakow:
### What/Why?

This PR ports underline and strikethrough implementation from RN package to plain Android package. 

### Test 1
1. Go to "Playground" screen
2. Mark `Underline`
3. Paste following example: 
```
__huhu__

~~hehe~~

~~__haha__~~

__~~haha~~__

> ~~__hihi__~~
```

| Before | After |
| - | - |
| image | image |

### Test 2
1. Open "Text" screen
2. Scroll to "Priority Actions"
3. Verify that "Ignore the problem" renders correctly

| Before | After |
| - | - |
| image | image |

### PR Checklist

- [ ] Code compiles and runs on iOS (not applies)
- [ ] E2E tests are passing


packages/android-enriched-markdown/compose/build.gradle
```diff
@@ -36,7 +36,7 @@ android {
 }
 
 dependencies {
-  implementation(project(":ui"))
+  api(project(":ui"))
```
jnowakow, line 39: This change was required in order to expose `Md4cFlags`. I've used `typealias` to allow users to import directly from `compose` package instead of `parser`: 
```kotlin
import com.swmansion.enriched.markdown.compose.Md4cFlags
```

Alternative solution I've considered is adding new class in `compose` package that would mirror `Md4cFlags` from `parser` and map it internally. Drawback is that it would require manual maintenance.

eszlamczyk (approved):
LGTM!

hryhoriiK97 (approved):
LGTM! 🎖️

PR #738, opened 2026-09-02
fix(android): long press crash

### LINKED ISSUE #728: [Android] EnrichedMarkdownTextInput crashes in Editor.performLongClick on long press
**Describe the bug**

`EnrichedMarkdownTextInput` can crash Android during a long press with an uncaught framework `NullPointerException`:

```
NullPointerException: Attempt to invoke virtual method
'void android.widget.Editor$InsertionPointCursorController.show()'
on a null object reference
    at android.widget.Editor.performLongClick(Editor.java)
```

We have observed 8 production crashes affecting 7 users. Seven events were on Android 16 and one was on Android 13. Devices include Samsung phones and tablets, a Pixel 7 Pro, and an Oppo device.

Every inspected event occurred immediately after the user touched or long-pressed an `EnrichedMarkdownTextInput`. The affected inputs were mounted in several contexts, including a normal screen, a modal editor, and inside a parent `ScrollView`. This does not appear limited to a single screen-navigation lifecycle.

**Environment**

- react-native-enriched-markdown: 1.0.1
- React Native: 0.86.2
- Expo: 57
- New Architecture / Fabric: enabled
- Android: 13 and 16

**Usage**

```tsx
<EnrichedMarkdownTextInput
  defaultValue={value}
  editable
  multiline
  scrollEnabled={false}
  selectionMenuConfig={{
    format: { enabled: true },
    copyAsMarkdown: { enabled: true },
  }}
/>
```

**To reproduce**

The crash is intermittent in production and we do not yet have a deterministic standalone reproduction.

1. Render an editable multiline `EnrichedMarkdownTextInput`.
2. Focus the input.
3. Long-press the text or empty space near the insertion cursor.
4. Android calls `Editor.performLongClick()`.
5. `getInsertionController()` returns null and Android calls `.show()` on it.

The crash has also occurred immediately after opening a fresh editor, so remounting the input when a screen regains focus does not cover all observed cases.

**Expected behavior**

Long-press should open the native selection or insertion

eszlamczyk:
### What/Why?

Fixes #728.

`EnrichedMarkdownTextInputView` could crash Android on a long press with an uncaught framework NPE:

```
NullPointerException: Attempt to invoke virtual method
'void android.widget.Editor$InsertionPointCursorController.show()'
on a null object reference
    at android.widget.Editor.performLongClick(Editor.java)
    at android.widget.TextView.performLongClick(TextView.java)
```

This is an AOSP bug, not something we can fix in the framework. `Editor.performLongClick`
evaluates `mInsertionControllerEnabled` (true), then inside that guarded branch calls
`Selection.setSelection(...)`, whose span/watcher callbacks can re-run
`prepareCursorControllers()` and flip `mInsertionControllerEnabled` to false before the
immediately following `getInsertionController().show()` - which then dereferences null:

```java
// android.widget.Editor.performLongClick (android16-release)
if (!handled && !isPositionOnText(lastDownX, lastDownY)
        && !mTouchState.isOnHandle() && mInsertionControllerEnabled) {   // guard: true
    getOffsetForPosition(...);
    Selection.setSelection(text, offset);      // <- SpanWatcher -> our onSelectionChanged can disable the controller here
    getInsertionController().show();           // <- now returns null -> NPE
}
```

The mid-call flip happens because `Selection.setSelection` moves the selection span, whose
`SpanWatcher` callback runs our `onSelectionChanged` override synchronously; that override
re-runs `prepareCursorControllers()` and flips `mInsertionControllerEnabled` to false. Combined
with our block/formatting spans, this is why the input trips a framework bug that most
`EditText`s never hit.

The fix overrides `performLongClick()` and absorbs that specific NPE - but only when its top
stack frame is `android.widget.Editor.performLongClick`; any other `NullPointerException`
(including one from our own `onSelectionChanged`, `syncEmptyListAnchor`, or event emitters) is
rethrown so real regressions aren't masked. It logs a warning under our own tag rather than
swallowing silently, and returns `true` so the long press counts as handled (otherwise
`View.CheckForLongPress` leaves `mHasPerformedLongPress` unset and a spurious click fires on
`ACTION_UP`). The long press just skips the insertion UI on that frame instead of crashing. This
mirrors RN core's own precedent of swallowing framework bugs it cannot fix (e.g. the
`IndexOutOfBoundsException` catch in `ReactEditText`; RN's `ReactSoftExceptionLogger` is
`internal`, so we use a plain `Log.w`).

<details>
<summary>Deterministic reproduction (forced-race debug harness - do NOT feel obliged to run this)</summary>

The production crash is a timing/reentrancy race, so it cannot be reproduced by a plain
scripted long press - even the issue reporter has no deterministic standalone repro. To prove
the fix, I forced the exact race window with a debug-only Activity: a `SpanWatcher` disables
the insertion controller the instant `Selection.setSelection` moves the selection span inside
`performLongClick`. This is faithful to the production mechanism but artificially triggered, so
a reviewer would have to recreate the whole harness to see it. Not worth anyone's time to
re-run; included only for the record.

Setup: Android 13+ device/emulator, `adb shell settings put global hidden_api_policy 1` (the
harness reflects into `Editor.mInsertionControllerEnabled`), a debug-only Activity + manifest
entry, then `adb shell am start -n <app>/.ReproActivity` and tap the button.

```kotlin
// app/src/debug/java/.../ReproActivity.kt  (NOT committed - reference only)
class ReproActivity : Activity() {
  private val tag = "Issue728"

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
    val input = EnrichedMarkdownTextInputView(this).apply {
      setText("hello"); isLongClickable = false
    }
    val trigger = Button(this).apply {
      text = "Trigger #728"
      setOnClickListener { fireLongPressRace(input) }
    }
    setContentView(LinearLayout(this).apply {
      orientation = LinearLayout.VERTICAL
      addView(input, LinearLayout.LayoutParams(MATCH_PARENT, 240))
      addView(trigger)
    })
  }

  private fun fireLongPressRace(input: EnrichedMarkdownTextInputView) {
    input.requestFocus()
    val editable = input.text ?: return
    val watcher = object : SpanWatcher {
      override fun onSpanAdded(t: Spannable, w: Any, s: Int, e: Int) = maybeFlip(input, w)
      override fun onSpanRemoved(t: Spannable, w: Any, s: Int, e: Int) = Unit
      override fun onSpanChanged(t: Spannable, w: Any, os: Int, oe: Int, ns: Int, ne: Int) = maybeFlip(input, w)
    }
    editable.setSpan(watcher, 0, editable.length, Spanned.SPAN_INCLUSIVE_INCLUSIVE)

    input.post {
      // record a down in empty space so performLongClick takes the "empty space" branch
      val now = SystemClock.uptimeMillis()
      val down = MotionEvent.obtain(now, now, MotionEvent.ACTION_DOWN,
        (input.width - 4).toFloat(), (input.height - 4).toFloat(), 0)
      input.dispatchTouchEvent(down); down.recycle()

      setInsertionControllerEnabled(input, true)  // re-arm so every tap is a real trial
      Log.i(tag, "performLongClick(); enabled=${insertionControllerEnabled(input)}")
      val handled = input.performLongClick()
      Log.i(tag, "returned $handled (no crash)")
      editable.removeSpan(watcher)
    }
  }

  private fun maybeFlip(input: TextView, what: Any) {
    if (what === Selection.SELECTION_END || what === Selection.SELECTION_START) {
      setInsertionControllerEnabled(input, false)  // reflect Editor.mInsertionControllerEnabled = false
      Log.i(tag, "selection span moved -> disabled insertion controller mid-performLongClick")
    }
  }

  // reflection helpers into TextView.mEditor / Editor.mInsertionControllerEnabled omitted for brevity
}
```

Observed logs, unpatched vs patched, under identical `enabled=true` + selection-span-flip conditions:

```
# unpatched
I Issue728: performLongClick(); enabled=true
I Issue728: selection span moved -> disabled insertion controller mid-performLongClick
E AndroidRuntime: FATAL EXCEPTION: main
E AndroidRuntime: java.lang.NullPointerException: ... Editor$InsertionPointCursorController.show() on a null object reference
E AndroidRuntime:   at android.widget.Editor.performLongClick(Editor.java:1499)
E AndroidRuntime:   at android.widget.TextView.performLongClick(TextView.java:15311)

# patched
I Issue728: performLongClick(); enabled=true
I Issue728: selection span moved -> disabled insertion controller mid-performLongClick
I Issue728: returned true (no crash)
```

</details>

### Testing

Verified A/B on an Android 16 (API 36) emulator with the forced-race harness above:

- Without the fix: `FATAL EXCEPTION: main` - the exact #728 stack. App dies.
- With the fix: `performLongClick returned true (no crash)` - NPE caught, activity stays up.

Note: please do not feel you need to reproduce this. It is a reentrancy/timing
race in the platform, only reproducible via the artificial harness above; a normal long press on
a stock emulator will not trip it. The change is a narrowly scoped `try/catch(NullPointerException)`
around a single framework call that can only turn a crash into a no-op - it cannot alter the
happy-path long-press behavior (selection/insertion UI still works normally).

### PR Checklist

- [ ] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -290,6 +290,16 @@ class EnrichedMarkdownTextInputView(
 
   override fun performClick(): Boolean = super.performClick()
 
+  // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
+  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE;
+  // the long-press just skips the insertion UI instead of crashing.
+  override fun performLongClick(): Boolean =
+    try {
+      super.performLongClick()
+    } catch (e: NullPointerException) {
+      false
+    }
+
   // In auto-grow mode (scrollEnabled=false) TextView's internal bringPointIntoView
   // scrolls content before Fabric has resized the view, causing a visible flicker.
   override fun scrollTo(
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -17,6 +17,7 @@ import android.view.inputmethod.EditorInfo
 import android.view.inputmethod.InputConnection
 import android.view.inputmethod.InputMethodManager
 import androidx.appcompat.widget.AppCompatEditText
+import com.facebook.react.bridge.ReactSoftExceptionLogger
 import com.facebook.react.common.ReactConstants
 import com.facebook.react.uimanager.BackgroundStyleApplicator
 import com.facebook.react.uimanager.PixelUtil
@@ -291,12 +292,14 @@ class EnrichedMarkdownTextInputView(
   override fun performClick(): Boolean = super.performClick()
 
   // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
-  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE;
+  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE
+  // and log it as a soft exception (mirroring ReactEditText's known-framework-crash handling);
   // the long-press just skips the insertion UI instead of crashing.
   override fun performLongClick(): Boolean =
     try {
       super.performLongClick()
     } catch (e: NullPointerException) {
+      ReactSoftExceptionLogger.logSoftException(ReactConstants.TAG, e)
       false
     }
 
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -8,6 +8,7 @@ import android.graphics.Color
 import android.os.Build
 import android.text.Editable
 import android.text.InputType
+import android.util.Log
 import android.util.TypedValue
 import android.view.Gravity
 import android.view.KeyEvent
@@ -17,7 +18,6 @@ import android.view.inputmethod.EditorInfo
 import android.view.inputmethod.InputConnection
 import android.view.inputmethod.InputMethodManager
 import androidx.appcompat.widget.AppCompatEditText
-import com.facebook.react.bridge.ReactSoftExceptionLogger
 import com.facebook.react.common.ReactConstants
 import com.facebook.react.uimanager.BackgroundStyleApplicator
 import com.facebook.react.uimanager.PixelUtil
@@ -293,13 +293,13 @@ class EnrichedMarkdownTextInputView(
 
   // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
   // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE
-  // and log it as a soft exception (mirroring ReactEditText's known-framework-crash handling);
-  // the long-press just skips the insertion UI instead of crashing.
+  // and log it (mirroring ReactEditText's known-framework-crash handling); the long-press just
+  // skips the insertion UI instead of crashing.
   override fun performLongClick(): Boolean =
     try {
       super.performLongClick()
     } catch (e: NullPointerException) {
-      ReactSoftExceptionLogger.logSoftException(ReactConstants.TAG, e)
+      Log.w(ReactConstants.TAG, "Absorbed framework NPE in performLongClick (issue #728)", e)
       false
     }
 
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
@@ -291,16 +291,26 @@ class EnrichedMarkdownTextInputView(
 
   override fun performClick(): Boolean = super.performClick()
 
-  // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
-  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE
-  // and log it (mirroring ReactEditText's known-framework-crash handling); the long-press just
-  // skips the insertion UI instead of crashing.
+  // Framework bug (issue #728): inside Editor.performLongClick, the guarded branch calls
+  // Selection.setSelection, whose SpanWatcher callback runs our onSelectionChanged override
+  // synchronously; that can re-run prepareCursorControllers and flip mInsertionControllerEnabled
+  // to false before the immediately following getInsertionController().show(), which then
+  // dereferences null. Absorb only that specific NPE (thrown from Editor.performLongClick) and
+  // rethrow anything else so unrelated regressions in our own callbacks aren't masked. We log it
+  // rather than swallowing silently (RN's ReactSoftExceptionLogger is internal to RN, so we use a
+  // plain warning under our own tag), and return true so the long press counts as handled (else
+  // View.CheckForLongPress leaves mHasPerformedLongPress unset and a spurious click fires on
+  // ACTION_UP); the long press just skips the insertion UI on that frame instead of crashing.
   override fun performLongClick(): Boolean =
     try {
       super.performLongClick()
     } catch (e: NullPointerException) {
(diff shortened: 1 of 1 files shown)


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -290,6 +290,16 @@ class EnrichedMarkdownTextInputView(
...
 
+  // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
+  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE;
+  // the long-press just skips the insertion UI instead of crashing.
+  override fun performLongClick(): Boolean =
+    try {
+      super.performLongClick()
+    } catch (e: NullPointerException) {
+      false
+    }
```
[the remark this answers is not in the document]
eszlamczyk: This is too broad, replacing this with something that mirrors RN's known-framework-crash convention (log a soft exception rather than swallow silently)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -290,6 +291,18 @@ class EnrichedMarkdownTextInputView(
...
   override fun performClick(): Boolean = super.performClick()
 
+  // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
+  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE
+  // and log it (mirroring ReactEditText's known-framework-crash handling); the long-press just
+  // skips the insertion UI instead of crashing.
+  override fun performLongClick(): Boolean =
+    try {
+      super.performLongClick()
+    } catch (e: NullPointerException) {
```
hryhoriiK97, line 307: This catches every `NullPointerException` from the entire `super.performLongClick()` call chain — including your own `onSelectionChanged` override, `syncEmptyListAnchor`, event emitters, etc. A bug in any of those would be silently swallowed and logged as a "framework NPE".

Can we rethrow unless the top frame is actually `Editor.performLongClick`?

```kotlin
} catch (e: NullPointerException) {
  val top = e.stackTrace.firstOrNull()
  if (top == null || !top.className.startsWith("android.widget.Editor") || top.methodName != "performLongClick") throw e
  Log.w(TAG, "Absorbed framework NPE in performLongClick (issue #728)", e)
  true
}
```

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -290,6 +291,18 @@ class EnrichedMarkdownTextInputView(
...
+      Log.w(ReactConstants.TAG, "Absorbed framework NPE in performLongClick (issue #728)", e)
+      false
```
hryhoriiK97: This should be `true`. Returning `false` tells the framework the long-press wasn't handled, so `View.CheckForLongPress` won't set `mHasPerformedLongPress` and a spurious click fires on ACTION_UP. Returning `true` suppresses that.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -290,6 +291,18 @@ class EnrichedMarkdownTextInputView(
 
   override fun performClick(): Boolean = super.performClick()
 
+  // Framework bug (issue #728): Editor.performLongClick can call getInsertionController().show()
```
hryhoriiK97: Nit: the comment (and PR description) mentions `MarkdownEditable` and `MarkdownTextWatcher` as the trigger, but neither is involved. There's no `MarkdownEditable` class, and `MarkdownTextWatcher` is a `TextWatcher` — `Selection.setSelection()` fires `SpanWatcher` callbacks, not `TextWatcher`. The actual synchronous path is `Selection.setSelection()` → `spanChange()` → your `onSelectionChanged()` override.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/EnrichedMarkdownTextInputView.kt
```diff
@@ -290,6 +291,18 @@ class EnrichedMarkdownTextInputView(
...
+  // on a null controller when our editable's span callbacks disable it mid-call. Absorb that NPE
+  // and log it (mirroring ReactEditText's known-framework-crash handling); the long-press just
+  // skips the insertion UI instead of crashing.
+  override fun performLongClick(): Boolean =
+    try {
+      super.performLongClick()
+    } catch (e: NullPointerException) {
+      Log.w(ReactConstants.TAG, "Absorbed framework NPE in performLongClick (issue #728)", e)
```
hryhoriiK97: Nit: `ReactConstants.TAG` is `"ReactNative"` - this'll look like an RN-core warning in log dashboards. Maybe use `"EnrichedMarkdown"` or similar so it's easy to filter.

hryhoriiK97 (approved):

PR #740, opened 2026-09-02
fix(android): non-latin symbols incorrectly shrinking line-height

### LINKED ISSUE #643: Android: "ᕱ⑅ᕱ ྀི" shifts the whole text line upward and clips its top
### Bug summary

On Android, adding the visible kaomoji suffix `ᕱ⑅ᕱ ྀི` to a single-line `EnrichedMarkdownText` changes the line metrics. With `fontSize: 14` and `lineHeight: 20`, the entire line—including the English text—is shifted upward and clipped at the top.

Removing the trailing `ྀི` restores the expected vertical alignment. These two trailing combining marks are `U+0F80` and `U+0F72`.

### Version

```text
react-native-enriched-markdown: 0.7.4
react-native: 0.86.0
expo: 57.0.7
platform: Android
```

### Minimal reproduction

```tsx
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import { EnrichedMarkdownText } from 'react-native-enriched-markdown';

const markdownStyle = {
  paragraph: {
    color: '#111111',
    fontSize: 14,
    lineHeight: 20,
    marginTop: 0,
    marginBottom: 0,
  },
};

export default function App() {
  return (
    <View style={styles.screen}>
      <Text>Control</Text>
      <EnrichedMarkdownText
        markdown="TALL ENGLISH TEXT ᕱ⑅ᕱ"
        markdownStyle={markdownStyle}
        flavor="commonmark"
        allowTrailingMargin={false}
        containerStyle={styles.textBox}
      />

      <Text>Reproduction</Text>
      <EnrichedMarkdownText
        markdown="TALL ENGLISH TEXT ᕱ⑅ᕱ ྀི"
        markdownStyle={markdownStyle}
        flavor="commonmark"
        allowTrailingMargin={false}
        containerStyle={styles.textBox}
      />
    </View>
  );
}

const styles = StyleSheet.create({
  screen: {
    flex: 1,
    justifyContent: 'center',
    padding: 24,
    gap: 8,
  },
  textBox: {
    width: '100%',
    borderWidth: StyleSheet.hairlineWidth,
    borderColor: 'red',
  },
});
```

### Steps to reproduce

1. Run the example on Android.
2. Compare the two bordered text views.
3. In `TALL ENGLISH TEXT ᕱ⑅ᕱ ྀི`, the whole line is positioned higher and clipped at the top.
4. Remove only `ྀི`; the expected alignment returns.

### Expected behavior

The complete line should remain visible and vertically aligned.

### Actual behavior

The line containing `ᕱ⑅ᕱ ྀི` is shifted upward and clipped at the top.

eszlamczyk:
### What/Why?

Fixes #643.

On Android, a line of `EnrichedMarkdownText` taller than its `lineHeight` (e.g. a trailing tall glyph like stacked Tibetan/Japanese combining marks) was clipped at the top and dragged the whole line up.

**Cause.** `spans/LineHeightSpan.kt` distributes leading as `lineHeight - ((-ascent) + descent)`. When content is taller than `lineHeight` this value is negative, so it *shrinks* ascent/descent below the glyph's natural size and the overflow gets clipped.

**Fix.** Make the clamp expand-only - when `leading <= 0`, keep the natural metrics so the line grows to fit instead of clipping. Normal (positive-leading) lines are unchanged. Applied to both Android source trees.

**Note (intentional divergence from RN).** RN's own `CustomLineHeightSpan` treats `lineHeight` as a hard ceiling and clips the same glyph (a plain RN `<Text>` was verified to clip identically). We deliberately grow the line instead. Trade-off: a line with an oversized-metric font can end up taller than `lineHeight`, since a `LineHeightSpan` only sees font metrics, not visible ink.

### Testing

Repro is font-dependent (the report's orphan `U+0F80 U+0F72` only render tall on some fonts). Verified on the emulator with a tall Tibetan syllable (`ཧ`, 14/20) and font-independent ASCII (24/12): tall rows fully visible after the fix, normal rows unchanged, plain RN `<Text>` still clips.

#### Screenshots

| Before | After |
| - | - |
| image |  |

### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (approved):
LGTM! Have you run Maestro tests to verify there are no regressions with the line height change?

PR #746, opened 2026-09-03
feat: onCodePress prop

### LINKED ISSUE #210: Callback when elements like code block is pressed
If someone wants to copy the code, we can enable the select but then they can select everything in the page and selecting just the code block isn't easy.

If we could have a callback like `onCodeBlockPress` just like we have for links, it would be great to build features like copy to clipboard.

No alternative :( 

**Additional context**
No

eszlamczyk:
### What/Why?

Closes #210.

Adds an `onCodeBlockPress` prop that fires with `{ code, language }` when a fenced code block is tapped/clicked anywhere in its body - useful for building actions like copy-to-clipboard or opening the code in a viewer.

- Off by default: the block only becomes tappable when the callback is provided (gated by an internal `enableCodeBlockPress` prop, mirroring `onImagePress`). Text selection, the GitHub-flavor header copy button, and the long-press context menu are unchanged.
- A tap never fires while text is being selected - selection stays separate from the press.
- Works across the board: `flavor="commonmark"` and `flavor="github"` on iOS and Android, and on web (the code block becomes a clickable, keyboard-activatable element with a button role).

### Testing

- Storybook: **Block → Code Block → Events** wires `onCopyPress` and `onCodeBlockPress` together; flip the `flavor` control to test commonmark vs github. The story intercepts `onCodeBlockPress` and alerts if the payload is ever an array, otherwise logs `{ code, language }` to the Actions panel.
- Example app: `TextScreen` (GFM) wires `onCodeBlockPress`.
- Verified tapping the header, the code text, and the box background/padding all fire; the copy button still only copies; scrolling and long-press are unaffected.

| iOS | Android |
| - | - |
| <video src="https://github.com/user-attachments/assets/1086c77b-d5c6-492b-9e14-e46425c27680" width=300 />  | <video src="https://github.com/user-attachments/assets/45fb5269-0851-4fee-a76b-573ca85aded0" width=300 />  |

### PR Checklist

- [ ] E2E tests are passing


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/EnrichedMarkdown.kt
```diff
@@ -345,6 +356,12 @@ class EnrichedMarkdown(
     }
   }
 
+  private fun pushCodeBlockPressToSegments() {
+    segmentViews.filterIsInstance<CodeBlockContainerView>().forEach {
```
hryhoriiK97, line 369: iOS's `pushCodeBlockPressToSegments` recurses into `ENRMBlockquoteContainerView` to propagate the flag to nested code blocks, but this only hits top-level `CodeBlockContainerView`s. If someone toggles `onCodeBlockPress` at runtime, code blocks inside blockquotes on Android won't update.

(Same gap exists for `pushBlockContextMenuToSegments` on both platforms, so it's a pre-existing pattern - just worth keeping in mind.)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/spans/CodeBlockSpan.kt
```diff
@@ -28,6 +28,9 @@ class CodeBlockSpan(
   private val context: Context,
   private val styleCache: SpanStyleCache,
   private val leadingIndent: Int = 0,
+  // Recovered by a tap on this span for onCodeBlockPress; unused by drawing.
+  val code: String = "",
```
hryhoriiK97, line 32: nit: Storing the full source text on the span means we keep the code string twice in memory (rendered glyphs + this payload). Fine for normal blocks, just something to be aware of if someone pastes a massive data dump into a fenced block.

apps/react-native-example/.rnstorybook/stories/components/EnrichedMarkdownText/block/CodeBlock.stories.tsx
```diff
@@ -129,3 +132,41 @@ export const CopyEvent: TextStory<CodeBlockStyleControls> = {
...
+    );
+    // Alert if the payload is ever an array, else forward the object to Actions.
+    const logCodeBlockPress = action('onCodeBlockPress');
+    return (
+      <EnrichedMarkdownTextStory
+        title="Code Block — Events"
+        description="onCodeBlockPress fires on a tap anywhere in the body (both flavors); onCopyPress fires from the header copy button / long-press → Copy (flavor=github only). onCodeBlockPress is intercepted here: an array payload triggers an Alert, otherwise it is logged to the Actions panel as { code, language }. Flip the flavor control to test commonmark vs github."
+        {...rest}
+        onCodeBlockPress={(event: CodeBlockPressEvent) => {
+          if (Array.isArray(event)) {
```
hryhoriiK97, line 164: nit: This looks like a leftover from debugging - `CodeBlockPressEvent` is always an object. Might confuse someone reading the story later; consider dropping it or adding a short comment why it's there.

hryhoriiK97 (approved):
LGTM! I've left a few comments

PR #753, opened 2026-09-03
feat(ios): add a render-plugin seam for optional modules

maksg:
Optional sibling modules plug into the render pipeline through a package-level MarkdownRenderPlugin (renderer lookup consulted before the built-ins, parser-flag adjustment, root-level block wrapping) plus a MarkdownPluginAttachment protocol that keeps Copy as Markdown and fixed line-height handling generic. The latexMath parser flag leaves the public API: math parsing can only be enabled by a plugin, so base needs no math renderer at all and $…$ stays literal text without one. No public API is added; the whole seam is SE-0386 package access.

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):

hryhoriiK97 (approved):
LGTM 🎖️

PR #765, opened 2026-09-06
feat: add native video playback support

hryhoriiK97:
### What/Why?
Adds video playback as a new block element. Videos are embedded using the HTML `<video>` tag — the standard way to include video in markdown. The parser recognizes block-level `<video>` tags and promotes them to native `Video` AST nodes. Other HTML tags are silently ignored (only `<video>` is allowlisted). Inline HTML remains disabled.

```markdown
<video src="https://example.com/ocean.mp4"></video>
```

**Parser (C++/md4c):**
- New `Video` AST node type, appended last to keep enum ordinals stable.
- Added `video` to md4c's type 6 block-level HTML tag list (vendored patch, [upstreamed to md4c](https://github.com/mity/md4c/pull/407)).
- Switched md4c flags from `MD_FLAG_NOHTML` to `MD_FLAG_NOHTMLSPANS` — block HTML enabled, inline HTML disabled.
- Lightweight HTML tag parser (`parseHtmlOpenTag`) extracts tag name and attributes; `tryPromoteHtmlBlock` converts allowlisted tags to AST nodes.
- Videos inside lists are promoted out as standalone blocks (list renderer doesn't support embedded native views). Videos inside blockquotes and admonitions are fully supported.
- Serialization round-trips Video nodes as `<video src="url"></video>`.

**iOS:**
- `ENRMVideoContainerView` wraps `AVPlayerViewController` via a thin host controller (`ENRMVideoHostController`) for correct child VC containment.
- Custom play icon overlay (semi-transparent circle + white triangle via `UIBezierPath`/`CAShapeLayer`) — hides on first tap, resets when the video URL changes. Added because `AVPlayerViewController` auto-hides native controls after ~3s with no API to prevent it.
- Proper appearance transitions in `dealloc` and `didMoveToWindow` cleanup when the view is removed without dealloc.
- macOS is gated out via `ENRMFeatureFlags.h` (`#undef ENRICHED_MARKDOWN_VIDEO` when `TARGET_OS_OSX`), so the guard is centralized instead of scattered across files.

**Android:**
- `VideoContainerView` uses ExoPlayer/Media3 `PlayerView` with rounded corners via `OutlineProvider`.
- Lives in a conditional source set (`src/video/`) — only compiled when `enableVideo` is set in `build.gradle`.
- Loaded reflectively via `Class.forName()`, with a ProGuard `-keep` rule in `consumer-rules.pro`.
- `releasePlayer()` resets `currentUrl` so `applyVideoNode` re-creates the player after navigate-back.
**Web:**
- `<video>` element with `controls`, `playsInline` (prevents iOS Safari forced fullscreen), and `display: block` (eliminates inline descender-gap whitespace).

**Styles:**
- `VideoStyle` with `marginTop`, `marginBottom`, `borderRadius`, `aspectRatio`, `backgroundColor`.
- Defaults: 16:9 ratio, black background, consistent across all platforms.

**Config:**
- `enableVideo` defaults to `true`, matching `enableCodeHighlight` and `enableMath`. Uses the `!= false` comparison pattern consistent with other features.

**Limitations:**
- macOS video playback is not yet supported (documented in `docs/MACOS.md`, planned for a follow-up PR).
- Videos inside lists are rendered as standalone blocks without bullet/marker. Supporting list markers requires making lists a container segment type (like `BlockquoteContainerView`), planned for a future release.

#### Screenshots

| iOS | Android |
| - | - |
| <video src="https://github.com/user-attachments/assets/1aa1ebe5-a277-4549-b5bd-2e9a8b09bde6" width=300 /> | <video src="https://github.com/user-attachments/assets/00a28588-1d53-4568-97bd-51a7161aee5a" width=300 /> |

### PR Checklist

- [ ] E2E tests are passing


apps/react-native-example/ios/EnrichedMarkdownExample/Info.plist
```diff
@@ -29,7 +29,7 @@
 	<key>NSAppTransportSecurity</key>
 	<dict>
 		<key>NSAllowsArbitraryLoads</key>
-		<false/>
+		<true/>
```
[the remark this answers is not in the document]
hryhoriiK97: This is an example/dev app, not a production app shipped to the App Store.

apps/react-native-example/ios/EnrichedMarkdownExample/Info.plist
```diff
@@ -56,6 +56,10 @@
 		<string>UIInterfaceOrientationPortrait</string>
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 	</array>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>audio</string>
+	</array>
```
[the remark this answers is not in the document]
hryhoriiK97: This is an example/dev app, not a production app shipped to the App Store.

packages/core/cpp/parser/MD4CParser.cpp
```diff
@@ -669,6 +740,7 @@ std::shared_ptr<MarkdownASTNode> MD4CParser::parse(const std::string &markdown,
 
   if (impl_->root) {
     promoteDisplayMathFromParagraphs(*impl_->root);
+    promoteVideoImages(*impl_->root);
     wrapListItemInlineRuns(*impl_->root);
   }
```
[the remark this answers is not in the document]
hryhoriiK97: [2887c51](https://github.com/software-mansion/enriched-markdown/pull/765/commits/2887c5101823acb3fcb165cffb46688f7bdd037a)

packages/react-native-enriched-markdown/android/src/video/java/com/swmansion/enriched/markdown/segments/VideoContainerView.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    addView(playerView, LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT))
+  }
+
+  fun applyVideoNode(node: MarkdownASTNode) {
+    val url = node.getAttribute("url") ?: return
+    if (url == currentUrl) return
+    currentUrl = url
+
+    releasePlayer()
```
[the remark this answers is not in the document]
hryhoriiK97: [a5fb342](https://github.com/software-mansion/enriched-markdown/pull/765/commits/a5fb3421019efaeb3ae27c577b3ae138dba5afed)

packages/core/cpp/parser/MD4CParser.cpp
```diff
@@ -525,12 +526,82 @@ bool isBlockNode(const MarkdownASTNode &node) {
...
+
+  for (size_t i = 0; i < children.size(); ++i) {
+    auto &child = children[i];
+
+    // Recurse into block containers so nested videos are also detected.
+    if (child->type == NodeType::Blockquote || child->type == NodeType::Admonition ||
+        child->type == NodeType::UnorderedList || child->type == NodeType::OrderedList ||
+        child->type == NodeType::ListItem) {
+      promoteVideoImages(*child);
+      continue;
```
eszlamczyk: This might be hard to maintain, how expensive would it be to just recurse all children (or maybe to be safer do the oposite case to stop the recursing if we are in a child we know it won't have any video/image shenanigas)

packages/core/cpp/parser/MD4CParser.cpp
```diff
@@ -525,12 +526,82 @@ bool isBlockNode(const MarkdownASTNode &node) {
...
+    }
```
jnowakow: Videos in block containers are promoted but they lack renderers. Here's list example on Android: 
```
- ![alpha](https://interactive-examples.mdn.mozilla.net/media/cc0-videos/flower.mp4)
- plain second item
```

Image

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/SegmentSignature.kt
```diff
@@ -13,6 +13,7 @@ object SegmentSignature {
   internal const val MATH_KIND_SALT = 0x6D61746800000000L // "math"
   internal const val CODE_BLOCK_KIND_SALT = 0x63626C6B00000000L // "cblk"
   internal const val BLOCKQUOTE_KIND_SALT = 0x6271746500000000L // "bqte"
+  internal const val VIDEO_KIND_SALT = 0x7669646F00000000L // "vido"
```
jnowakow:
```suggestion
  internal const val VIDEO_KIND_SALT = 0x7669646F00000000L // "video"
```

packages/core/cpp/parser/MD4CParser.cpp
```diff
@@ -525,12 +526,82 @@ bool isBlockNode(const MarkdownASTNode &node) {
...
+  // Lower-case the extension (including the dot) for comparison.
+  std::string ext;
+  ext.reserve(pathEnd - dotPos);
+  for (size_t i = dotPos; i < pathEnd; ++i) {
+    ext += static_cast<char>(std::tolower(static_cast<unsigned char>(url[i])));
+  }
+
+  static const char *videoExtensions[] = {
+      ".mp4", ".mov", ".webm", ".m4v", ".avi", ".mkv", ".ogv", ".3gp",
+  };
```
jnowakow: `.webm` doesn't work on iOS, as it lacks codecs: 
```
![w](https://interactive-examples.mdn.mozilla.net/media/cc0-videos/flower.webm)
```

Image

hryhoriiK97: I've actually changed the approach in the latest push - we no longer detect videos by file extension from image syntax. Instead, videos are now embedded using the HTML <video> tag directly:

```
<video src="https://example.com/video.mp4"></video>
```

Can you give another look to this PR?

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/segments/SegmentSignature.kt
hryhoriiK97: [e933255](https://github.com/software-mansion/enriched-markdown/pull/765/commits/e933255436300294f8f15dbdb2f13e37da6038d7)

packages/core/cpp/parser/MD4CParser.cpp
hryhoriiK97: @jnowakow good catch! Videos in lists are tricky - lists are rendered as attributed strings, so we can't embed a native video player view inside them. For now we promote the video out of the list and render it as a standalone block above the remaining items [86e09f9](https://github.com/software-mansion/enriched-markdown/pull/765/commits/86e09f91da77e8de3a762257e4580dbd50d3c41c). The bullet marker won't show for video items.

To support this properly, lists would need to become a container segment type (like BlockquoteContainerView) where each list item is its own native view that can host either text or a video + bullet marker. That's essentially rewriting the list renderer, so we'll revisit this later.

I've documented this limitation.

packages/core/cpp/md4c/md4c.c
```diff
@@ -6213,7 +6214,7 @@ md_is_html_block_start_condition(MD_CTX* ctx, OFF beg)
      * tree to speed-up the search. */
     static const TAG* map6[26] = {
         a6, b6, c6, d6, xx, f6, xx, h6, i6, xx, xx, l6, m6,
-        n6, o6, p6, xx, xx, s6, t6, u6, xx, xx, xx, xx, xx
+        n6, o6, p6, xx, xx, s6, t6, u6, v6, xx, xx, xx, xx
```
hryhoriiK97, line 6217: I've upstreamed this change to md4c - [mity/md4c#407](https://github.com/mity/md4c/pull/407). For now we carry the patch in our vendored copy.

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/serialization/MarkdownASTSerializer.kt
```diff
@@ -316,6 +316,13 @@ object MarkdownASTSerializer {
         buffer.append(")")
       }
 
+      NodeType.Video -> {
+        val url = node.getAttribute("url") ?: ""
+        buffer.append("<video src=\"")
+        buffer.append(url)
+        buffer.append("\"></video>")
```
eszlamczyk, line 323: Nit: I think it would be cleaner, to make video self closing here, as I believe this is more common case:


```suggestion
        buffer.append("\" />")
```

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/serialization/MarkdownASTSerializer.kt
```diff
@@ -316,6 +316,13 @@ object MarkdownASTSerializer {
         buffer.append(")")
       }
 
+      NodeType.Video -> {
+        val url = node.getAttribute("url") ?: ""
+        buffer.append("<video src=\"")
+        buffer.append(url)
```
eszlamczyk, line 322: the `src` here is being emited unescaped - a `"` in the URL breaks re-parsing (value truncates at the quote). Entity-escaping won't round-trip since the parser doesn't decode entities. We could either emit single quotes when URL contains a `"` or document this (as url with a `"` doesn't seem that common)

packages/react-native-enriched-markdown/ios/utils/MarkdownASTSerializer.m
```diff
@@ -95,6 +95,12 @@ static void serializeNode(MarkdownASTNode *node, NSMutableString *buffer)
       break;
     }
 
+    case MarkdownNodeTypeVideo: {
+      NSString *url = node.attributes[@"url"] ?: @"";
+      [buffer appendFormat:@"<video src=\"%@\"></video>", url];
```
eszlamczyk, line 100: same comments as for android

eszlamczyk (approved):
Overall okay, just found a some issues in my tests:

1. The mute button on ios does not work (on android there isn't even a mute button present)

2. There is no longpress action, maybe its worth adding (same old `copy` and `copy as markdown` should be enough)

Rest looks good, so Im approving

jnowakow (approved):
It looks good for me 💯

packages/react-native-enriched-markdown/ios/utils/MarkdownASTSerializer.m
hryhoriiK97: [da3532d](https://github.com/software-mansion/enriched-markdown/pull/765/commits/da3532d6f697064d09af4bf6bbce34264e83c3dd)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/serialization/MarkdownASTSerializer.kt
hryhoriiK97: [da3532d](https://github.com/software-mansion/enriched-markdown/pull/765/commits/da3532d6f697064d09af4bf6bbce34264e83c3dd)

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/utils/common/serialization/MarkdownASTSerializer.kt
hryhoriiK97: [da3532d](https://github.com/software-mansion/enriched-markdown/pull/765/commits/da3532d6f697064d09af4bf6bbce34264e83c3dd)

PR #772, opened 2026-09-07
feat(android): add table support to android-enriched-markdown

jnowakow:
### What/Why?

Adds GFM table rendering to the standalone `enriched-markdown-android` package. Cells keep their inline styles (bold, italic, strikethrough, inline code, links), and tables round-trip back to their markdown source, so a selection containing a table can be copied with "Copy as Markdown".

Tables are drawn by a `ReplacementSpan` inside the existing `TextView` rather than by a nested view, so a rendered document stays a single `TextView` and selection, copying and accessibility keep working across it. The span takes zero horizontal advance and reports the table height through the line’s font metrics, so the table gets a line of its own.

A table keeps its natural width instead of being squeezed to fit, and scrolls horizontally on its own: because a `ReplacementSpan` never receives touch events, `TableScrollController` claims the drag/fling and feeds it to the span, which redraws its content through a viewport-sized window with the frame pinned in place. A table that fits its viewport never claims a gesture, so vertical scrolling of the document is unaffected.

### Test 1
1. Go to "Playground" screen
2. Paste following example:
```
| Style | Rendered in cell | Marker |
| --- | --- | --- |
| Bold | **bold text** | `**` |
| Italic | *italic text* | `*` |
| Strikethrough | ~~struck out~~ | `~~` |
| Inline code | `fun main()` | backticks |
| Link | [Software Mansion](https://swmansion.com) | inline link |
| Combined | ***bold italic*** plus ~~**struck bold**~~ | nested |
```

| Before | After |
| - | - |
|  1-test1-before-playground | 2-test1-after-playground |

### Test 2
1. Open "Text" screen
2. Scroll to "Forest Statistics by Type"

| Before | After |
| - | - |
| 3-test2-before-text-screen | 4-test2-after-text-screen |

### Test 3
1. Open "Playground" screen
2. Paste the same example as in Test 1
3. Select the table and click "Copy as Markdown"
4. Paste copied text to a notebook and verify it matches the original.

### PR Checklist

- [ ] Code compiles and runs on iOS (not applies)
- [x] Code compiles and runs on Android
- [x] Updated documentation/README if applicable
- [x] Ran example app to verify changes
- [ ] E2E tests are passing
- [ ] Required E2E tests have been added (if applicable)

PR #774, opened 2026-09-08
feat(android): add admonition support to enriched-markdown-android

jnowakow:
### What/Why?

Adds GitHub admonitions (alerts) to the standalone `enriched-markdown-android` package. A blockquote opening with `> [!NOTE]`, `[!TIP]`, `[!IMPORTANT]`, `[!WARNING]` or `[!CAUTION]` renders as a themed callout — the usual blockquote geometry plus a header row with a tinted octicon and a bold title. Callouts round-trip to their markdown source, so "Copy as Markdown" works, and the header is announced by screen readers ahead of the body. Requires `Md4cFlags(admonitions = true)`. Admonitions are reusing `BlockquoteRenderer` since an admonition is just a themed quote.

### Test 1
1. Go to "Playground" screen
2. Paste the following example:
```
> [!NOTE]
> Useful information that users should know.

> [!TIP]
> Helpful advice for doing things better.

> [!IMPORTANT]
> Key information users need to know.

> [!WARNING]
> Urgent info needing immediate attention.

> [!CAUTION]
> Advises about risks or negative outcomes.
```

| Before | After |
| - | - |
| playground_admonitions_top | playground_admonitions_top |

### Test 2
1. Open "Text" screen
2. Scroll to "Field Notes"

The sample document gains this section in this PR, so there is no "before": on `main` the screen has no admonitions. Shown in two overlapping frames as it is taller than one viewport.

| Before | After |
| - | - |
| N/A | text_field_notes_1 |

### PR Checklist

- [ ] Code compiles and runs on iOS (not applies)
- [ ] E2E tests are passing


apps/android-example/app/src/main/java/swmansion/enriched/markdown/android/example/ExampleMarkdownStyles.kt
```diff
@@ -81,6 +81,15 @@ val CustomMarkdownStyle: MarkdownStyle =
       backgroundColor = Color(0xFFF9FAFB)
       gapWidth = 16.dp
       marginBottom = 16.dp
+      // Accent colors already default to the GitHub palette; this adds the matching background
+      // tints, which are opt-in so a callout is unfilled unless asked for.
```
jnowakow: Is this comment necessary?

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
```diff
@@ -530,7 +532,101 @@ internal data class BlockquoteStylePatch(
...
       gapWidth = gapWidth?.let(units::dp) ?: base.gapWidth,
       backgroundColor = backgroundColor?.let(units::color) ?: base.backgroundColor,
+      admonitions = applyAdmonitions(base, units),
     )
+
+  /**
+   * Merges the patched types over the palette already in [base], so overriding one alert type
+   * leaves the other four on their defaults. Within a type, an unset field keeps its previous
+   * value; a type touched for the first time inherits the plain blockquote border color.
+   */
```
jnowakow: Feels unnecessary?

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spans/AdmonitionIcons.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  private val PATHS: Map<String, Path> by lazy {
+    PATH_DATA.mapValues { (_, data) -> PathParser.createPathFromPathData(data) }
+  }
+
+  /** The octicon path for [type] in the 16x16 icon space. */
+  fun path(type: String): Path? = PATHS[type]
+
+  /** Capitalized header title for [type] (e.g. "note" -> "Note"). */
+  fun title(type: String): String = TITLES[type] ?: type.replaceFirstChar { it.uppercase() }
```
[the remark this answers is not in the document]
jnowakow: False positive: the receiver here is a `Char`, not a `String`. `Char.uppercase()` uses invariant-locale mapping — the locale-sensitive variant is the `Char.uppercase(locale)` overload, which we do not call. The fallback is unreachable anyway, since `TITLES` has an entry for every key in `PATH_DATA` and `TYPES` is derived from those keys.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
```diff
@@ -37,11 +38,19 @@ class MarkdownAccessibilityHelper(
     val linkUrl: String? = null,
     val listInfo: ListItemInfo? = null,
     val imageAltText: String? = null,
+    /**
+     * Bounds taken from a span's own geometry rather than from a text range — the band an
+     * admonition header is painted into.
+     */
+    val customBounds: Rect? = null,
+    /** Set for an admonition's header row; carries the alert type it announces. */
+    val admonitionType: String? = null,
```
eszlamczyk: Refer to two comments in #772

eszlamczyk (approved):

eszlamczyk (approved):

PR #775, opened 2026-09-08
feat: onLatexError callback

### LINKED ISSUE #749: Feature request: report LaTeX commands the engine could not render

There is no way to learn that a formula failed to render. When RaTeX cannot handle a command, the glyph is silently dropped — the app does not crash, nothing is logged, and the user is simply looking at a broken formula.

Our workaround is an app-side allowlist of ~272 commands, checked against the message text before rendering; anything unknown is sent to our error tracker. It has aged badly, in both directions:

- **False positives** — 90 days of production data gave 46 distinct commands, of which 22 are ones RaTeX demonstrably supports (`\implies`, `\underbrace`, `\checkmark`, `\overrightarrow`, `\varnothing`, `\stackrel`, `\phantom`, `\textcolor`, `\cancel`, …). Noise we triage for nothing.
- **False negatives** — a command that *is* in our list but that the engine cannot draw produces no signal at all. That is the case we actually care about, and the design cannot detect it.

The allowlist is fundamentally a guess about your engine's capabilities, maintained by people who do not own it, and it goes stale on every RaTeX bump with no signal that it has.

A callback on `EnrichedMarkdownText`, invoked when a math expression fails to parse or render — e.g.

```ts
onLatexError?: (e: { command?: string; source: string; message?: string }) => void
```

The engine already knows this; today the information stops at the native boundary. The current public surface exposes only `latexMath: boolean`.

With such a callback, consumers could report exactly what failed and delete the allowlist entirely — no false positives, no false negatives, and no staleness on engine upgrades.

- *Maintaining the allowlist properly* — what we do today. It has to be re-derived from RaTeX's source on every bump, and nothing detects when it is stale.
- *Deriving the list from RaTeX's source automatically* — our current plan. Better, but still an inference about the engine rather than a report from it, and complicated by the version skew reported separately.
- *Runtime rendering probes* — would need a visual oracle; not viable.

**Additional context**

React Native app on 1.0.2, math enabled on both platforms. Related: the RaTeX version skew reported in https://github.com/software-mansion/enriched-markdown/issues/748. Happy to prototype the JS side if the native plumbing would be welcome.

---
(Human written disclaimer: This was created with help of an AI agent 🤖 )

eszlamczyk:
### What/Why?

closes #749

Adds an `onLatexError` callback to `EnrichedMarkdownText`. Today, when the RaTeX engine can't parse or render a math expression it drops it silently, so apps have no way to know a formula broke. This callback surfaces the failure so consumers can report it (e.g. to an error tracker) instead of maintaining a brittle, stale-prone command allowlist.

- Payload: `onLatexError?: (e: { source: string; message?: string; displayMode: boolean }) => void`. The whole expression is the unit of failure — the engine rejects an expression atomically and only returns a free-form message — so there is intentionally **no `command` field**; `source` is the full failing inline span or block.
- Wired end-to-end on both platforms and both flavor view classes.
- De-duplicated per component instance and the cache persists across `markdown`/streaming updates, so a given failure fires once; a remount reports again (documented with a caution).
- Drive-by fix: a pre-existing crash where isolated block math (`$$…$$`) in **commonmark on Android** threw `requireBlockStyle` and blanked the entire document — now falls back gracefully (iOS already did).

### Testing

Verified on both platforms with a new Storybook story (`Props / Latex Error`): a document with two broken formulas (`\foo{x}` inline, `\bogus{y}` block) fires exactly two `onLatexError` calls; valid formulas stay silent. Switching the `flavor` control to `commonmark` still renders (crash fix).

#### Screenshots

| iOS | Android |
| - | - |
| <video src="https://github.com/user-attachments/assets/e9d101ce-2a4d-4865-8d1c-c47b541e254b" width=300 /> | <video src="https://github.com/user-attachments/assets/3c9cd1d1-18b9-4e70-bc59-fccf1be414d3" width=300 /> |

### PR Checklist

- [ ] E2E tests are passing


packages/react-native-enriched-markdown/ios/EnrichedMarkdownText.mm
```diff
@@ -790,6 +798,61 @@ - (void)emitContextMenuItemPress:(NSString *)itemText
     });
 }
 
+- (void)wireLatexErrorReporters:(NSArray<id<ENRMLatexErrorReporting>> *)reporters
```
hryhoriiK97: The five new latex-error methods (`wireLatexErrorReporters:`, `reportLatexErrorWithSource:...`, `emitLatexError:...`, `flushPendingLatexErrors`, `updateEventEmitter:`) and the two ivars are identical between `EnrichedMarkdownText.mm` and `EnrichedMarkdown.mm` — only the emitter cast differs. Could we pull the shared logic into a small helper to avoid them drifting apart over time?

packages/react-native-enriched-markdown/ios/EnrichedMarkdown.mm
```diff
@@ -1428,6 +1439,61 @@ - (void)emitContextMenuItemPress:(NSString *)itemText
...
+}
+
+- (void)reportLatexErrorWithSource:(NSString *)source message:(NSString *)message displayMode:(BOOL)displayMode
+{
+  NSString *key = [NSString stringWithFormat:@"%@ %@", displayMode ? @"B" : @"I", source];
+  if ([_reportedLatexErrors containsObject:key])
+    return;
+  [_reportedLatexErrors addObject:key];
+  if (![self emitLatexError:source message:message displayMode:displayMode]) {
+    [_pendingLatexErrors addObject:@{@"source" : source, @"message" : message, @"displayMode" : @(displayMode)}];
```
hryhoriiK97: Pending errors are stored/read with string keys (`@"source"`, `@"message"`, `@"displayMode"`). A typo in any key silently gives `nil` at flush time with no compiler help. Maybe worth a tiny struct or at least `static` key constants?

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/math/LatexErrorReporter.kt
```diff
@@ -0,0 +1,9 @@
+package com.swmansion.enriched.markdown.math
+
+fun interface LatexErrorReporter {
+  fun report(
+    source: String,
+    message: String,
```
hryhoriiK97: Minor: `message` is non-null `String` here, required `string` in codegen, but optional `string?` in the public TS type. It works because of the `?: ""` / `|| undefined` bridge, but the three layers disagree on the contract — could be confusing later. Would it make sense to use `String?` on the Kotlin side to match the public API?

hryhoriiK97 (approved):
LGTM! Left a few minor comments - take a look when you get a chance 👍

hryhoriiK97 (approved):

PR #780, opened 2026-09-08
fix: dont crash on liststylemarker being null

### LINKED ISSUE #773: crash on listStyleMarkerColor nil
Env:
React: 19.2.8
React native 0.86.3
Expo: 57.0
react-native-enriched-markdown: 1.0.2

Crash stack key snippet:

text
[CoreFoundation] *** Terminating app due to uncaught exception
'NSInvalidArgumentException',
reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]:
attempt to insert nil object from objects[1]'
Root cause location: The crash occurs in the native layer ListMarkerDrawer.m during ordered list marker rendering:

objc
// ListMarkerDrawer.m - Ordered list drawing
NSDictionary *mAttrs = @{
    NSFontAttributeName : font,
    NSForegroundColorAttributeName : [_config listStyleMarkerColor]   // ← objects[1] being nil will definitely crash
};
Unordered lists (drawBulletAtX) use [[_config listStyleMarkerColor] setFill], which tolerates nil values and does not crash;

Ordered lists use dictionary literals — once listStyleMarkerColor is nil, it will definitely crash.

The library fails to correctly retrieve the ordered list marker color from the JS styles (markerColor / list config), resulting in a nil value being inserted into the dictionary.

Root Cause
The ordered list native rendering logic in react-native-enriched-markdown@^1.0.2 has a defect: listStyleMarkerColor returns nil in some scenarios, and the native layer constructs the attributes dictionary using a literal without nil protection, triggering NSInvalidArgumentException.

As long as the content contains an ordered list (e.g., 1. xxx, 2. xxx, …), the app will crash immediately upon entering the page.

eszlamczyk:
### What/Why?

Closes #773.

Hardens the way we apply default for `listStyleMarker`. Now it fallbacks to default (and warns in dev), when the prop is invalid.

### Testing
Added test case to verify this

### PR Checklist

- [x] E2E tests are passing

### DIFF
--- a/packages/react-native-enriched-markdown/src/styleUtils.ts
+++ b/packages/react-native-enriched-markdown/src/styleUtils.ts
@@ -6,7 +6,16 @@ export const normalizeColor = (
 ): ColorValue | undefined => {
   if (!color) return undefined;
   if (Platform.OS === 'web') return color;
-  return processColor(color) ?? undefined;
+  const processed = processColor(color);
+  if (processed === null || processed === undefined) {
+    if (__DEV__) {
+      console.warn(
+        `[MarkdownStyle] Ignoring invalid color value "${color}"; falling back to the default.`
+      );
+    }
+    return undefined;
+  }
+  return processed;
 };
 
 export function mergeSubStyle<T extends Record<string, unknown>>(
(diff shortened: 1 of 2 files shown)


packages/react-native-enriched-markdown/src/styleUtils.ts
```diff
@@ -31,11 +40,13 @@ export function mergeSubStyle<T extends Record<string, unknown>>(
...
-    if (
-      key.toLowerCase().includes('color') &&
-      typeof result[key] === 'string'
-    ) {
-      result[key] = normalizeColor(result[key] as string);
+    if (key.toLowerCase().includes('color')) {
+      const value = result[key];
+      if (typeof value === 'string') {
+        result[key] = normalizeColor(value) ?? defaultValue;
+      } else if (value === undefined) {
```
hryhoriiK97: The `=== undefined` check misses `null` — `typeof null` is `'object'` so the string branch skips it, and `null === undefined` is `false` so this branch skips it too. I verified with a test: passing `{ markerColor: null as any }` leaves `null` in the result and the existing 3 tests still pass after the fix. Worth noting that even in TS, someone using a type assertion (`as any`) or receiving data from an untyped source can still pass `null` and hit the same crash.

```suggestion
      } else if (value == null) {
```

Loose equality catches both `null` and `undefined`.

hryhoriiK97 (approved):
LGTM! I've left one comment with a suggestion 👍

PR #783, opened 2026-09-08
feat(input): convert typed markdown block prefixes into blocks (markdownShortcuts)

### LINKED ISSUE #782: Feature: convert markdown block prefixes typed at line start (`# `, `- `, `1. `) into blocks in EnrichedMarkdownTextInput

Most markdown editors will convert headings and lists automatically as the user types - its how Obsidian, Notion, Noteplan, and others work.

When the user types a space after a recognised prefix at the start of a paragraph, apply the corresponding block and remove the prefix from the text:

- `#` … `######` + space → `toggleHeading(1..6)` on that paragraph
- `-`, `*`, `+` + space → `toggleUnorderedList()`
- `1.` (or `1)`) + space → `toggleOrderedList()`
- optionally `>` + space → blockquote, once the input supports it

Feeding these into the same block pipeline the ref commands already use would keep serialization, `onChangeState`, and the empty-line-anchor behaviour identical to the toolbar path. Backspace immediately after the conversion could undo it (the common convention), so a user who genuinely wants a literal `# ` can get one.

A prop to opt out (e.g. `markdownInputRules={false}`) would let apps that treat `#` as literal text — tags, channel names — keep today's behaviour.

- **A toolbar / `contextMenuItems`** — what we shipped. It works, but `contextMenuItems` and the Format submenu are only built once text is selected (`editMenuForTextInRange:` returns `nil` for `range.length == 0`; Android adds custom items only when `selectionStart < selectionEnd`), so an empty line can't be formatted at all — you must type first, then select. A keyboard-anchored toolbar avoids that but has to fight keyboard geometry (and hides behind a tab bar whenever a hardware keyboard keeps the soft one closed).
- **Implementing the rule app-side with `onKeyPress` + `insertText`/`setSelection`** — fragile: `onKeyPress` fires before the change is applied, Android's soft-keyboard key reporting can lag with autocomplete, and deleting the typed prefix from JS races the native edit session.
- **Escaping a literal leading `#` in the serializer** so the round-trip stops turning it into a heading — would make behaviour *consistent* but in the less useful direction; mentioning it because the current asymmetry is arguably a bug on its own.

**Additional context**

- `react-native-enriched-markdown@1.0.2`, iOS and Android; nightly `1.1.0-nightly-20260907` has no change here as far as I can see.
- Related: #339 (heading styling in the input, shipped), #497 (heading blocks on the block pipeline), #499 (`toggleHeading`), #392 (web input).
- Happy to test a nightly.

cvburgess:
### What/Why?

Closes #782.

Adds an **opt-in** `markdownShortcuts` prop to `EnrichedMarkdownTextInput`. When enabled, a space typed after `#`…`######`, `-`/`*`/`+`, or `1.`/`1)` at the start of a **plain paragraph** removes the prefix and turns the paragraph into the matching heading or list block — the Notion / Bear / Obsidian / Apple Notes convention.

Why: today a typed `# Heading` stays literal text, but the serializer emits it unescaped, so the same text becomes a heading the next time the markdown is loaded. Users read that as "formatting only applies after reload" (that's how #782 was found). The toolbar commands are the only live path to a block, and `contextMenuItems` / the Format menu are only built once text is selected, so an empty line can't be formatted at all without this.

How it's built — it reuses what's there rather than adding a new mechanism:

- The conversion runs where Enter's block continuation already mutates text mid-keystroke: iOS in `handleTextChanged` right after `processTextChangeWithContext:` (via the existing `replaceTextInRange:withText:formattingRanges:blockRanges:` path), Android as a `convertMarkdownShortcut` step in `EditPipeline.processTextChange` mirroring `continueBlocks` (delete inside `runAsATransaction`, re-adjust both stores, `setBlock`). Serialization, `onChangeState`, Return-continues-the-list and the Backspace-un-lists rule all behave exactly as if the block had been toggled from a toolbar.
- Matching is a tiny pure matcher per platform (`ENRMMarkdownShortcutMatcher`, `MarkdownShortcutMatcher.kt`). Ordered markers accept any number; the store renumbers ordinals.
- The trigger is the first space inside the inserted run, so it works for a single keystroke and for multi-character inserts (autocorrect, `inputText` in Maestro).
- Only fires on a plain paragraph — never inside an existing heading or list item.
- **Off by default** so apps that treat `#` or `-` as literal text (tags, dashes) see no change. Happy to flip the default if you'd rather match auto-link.

Docs: new "Markdown Shortcuts" section in `INPUT.md` (mirrors Auto-Link Detection) and an `API_REFERENCE.md` entry. The playground enables the prop so the new flows can exercise it.

### Testing

- `yarn typecheck`, `yarn lint` (0 errors; the one warning is pre-existing in `BlockRenderers.tsx`), `yarn test` (31 passing, including the Jest mock).
- `clang-format` is clean on the three iOS files I touched. Note `yarn lint-clang:ios` fails on pristine `main` today (a `gridView.menuProvider` line in the macOS code), so it can't be used as a gate here.
- **iOS, end to end:** compiled and exercised in a real consumer app (Dexter, RN 0.86.2, Xcode 26.6) by applying this diff to the installed 1.0.2 package via patch-package and rebuilding. Typing `# Heading test`, `- bullet item`, and `1. step one` on fresh paragraphs converted live into an H1, a bullet, and a numbered item with the prefixes removed; Return after the heading started a plain paragraph and two Returns exited the bullet list, as documented.
- **Android:** not compiled here — I have no Android SDK on this machine. The Kotlin follows `continueBlocks` line for line, but it needs CI (and ktlint) to confirm. I did not run their example app; the consumer-app build stood in for it.
- Two Maestro flows added under `flows/basic/block_elements/` (`markdown_shortcut_heading_test`, `markdown_shortcut_list_test`) following `list_toggle_test`. They end in the screenshot subflow, so baselines need generating with `yarn test:e2e:ios:update-screenshots` / `…android…` — I couldn't produce them without running your example app.

### PR Checklist

- [ ] Code compiles and runs on Android — not verified locally (no SDK); needs CI
- [ ] Ran example app to verify changes — verified in a consumer app instead (see Testing)
- [ ] E2E tests are passing — flows added, not run; screenshot baselines needed

### DIFF
--- a/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/editing/MarkdownShortcutMatcher.kt
+++ b/packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/editing/MarkdownShortcutMatcher.kt
@@ -1,0 +1,47 @@
+package com.swmansion.enriched.markdown.input.editing
+
+import com.swmansion.enriched.markdown.input.model.BlockType
+
+/**
+ * Recognizes the markdown block prefixes a user types at the start of a
+ * paragraph — `#`…`######`, `-`/`*`/`+`, `1.`/`1)` — so the input can turn the
+ * paragraph into the matching block and drop the prefix (Notion-style markdown
+ * shortcuts). Pure text matching; the caller owns the mutation.
+ */
+internal object MarkdownShortcutMatcher {
+  data class Match(
+    val type: BlockType,
+    val level: Int,
+  )
+
+  /** Anything longer is a number the user is writing, not a list marker. */
+  private const val MAX_ORDERED_MARKER_DIGITS = 9
+
+  /**
+   * Matches [prefix] (the paragraph text before the space the user just typed,
+   * with no leading whitespace). Ordered items match any number — the block
+   * store renumbers ordinals, so `3.` on a fresh line still starts at 1.
+   */
+  fun match(prefix: CharSequence): Match? {
+    if (prefix.isEmpty()) return null
+    val first = prefix[0]
+
+    if (first == '#') {
+      if (prefix.length > 6 || !prefix.all { it == '#' }) return null
+      val type = BlockType.forHeadingLevel(prefix.length) ?: return null
+      return Match(type, prefix.length)
+    }
+
+    if (prefix.length == 1 && (first == '-' || first == '*' || first == '+')) {
+      return Match(BlockType.UNORDERED_LIST_ITEM, 0)
+    }
+
+    val last = prefix.last()
+    if ((last == '.' || last == ')') && prefix.length >= 2 && prefix.length - 1 <= MAX_ORDERED_MARKER_DIGITS) {
+      if (prefix.subSequence(0, prefix.length - 1).all { it in '0'..'9' }) {
+        return Match(BlockType.ORDERED_LIST_ITEM, 0)
+      }
+    }
+    return null
+  }
+}
--- a/.maestro/enrichedMarkdownInput/flows/basic/block_elements/markdown_shortcut_heading_test.yaml
+++ b/.maestro/enrichedMarkdownInput/flows/basic/block_elements/markdown_shortcut_heading_test.yaml
@@ -1,0 +1,37 @@
+appId: swmansion.enriched.markdown.example
+tags:
+  - input
+  - heading
+  - block
+  - markdown-shortcuts
+---
+# Typing "# " at the start of a paragraph converts it into an H1 and drops the
+# prefix, so the visible text is "Title" styled as a heading — not the literal
+# "# Title" it used to stay until the markdown was reloaded.
+- launchApp:
+    clearState: true
+
+- runFlow:
+    file: '../../../../subflows/move_to_playground.yaml'
+
+- tapOn:
+    id: focus-button
+- waitForAnimationToEnd
+
+- inputText: '# Title'
+- waitForAnimationToEnd
+
+- pressKey: Enter
+- waitForAnimationToEnd
+
+# A heading does not continue onto the next line; this stays a paragraph.
+- inputText: 'Body text'
+- waitForAnimationToEnd
+
+- tapOn:
+    id: blur-button
+
+- runFlow:
+    file: '../../../subflows/capture_or_assert_screenshot.yaml'
+    env:
+      SCREENSHOT_NAME: 'markdown_shortcut_heading_input'
--- a/.maestro/enrichedMarkdownInput/flows/basic/block_elements/markdown_shortcut_list_test.yaml
+++ b/.maestro/enrichedMarkdownInput/flows/basic/block_elements/markdown_shortcut_list_test.yaml
@@ -1,0 +1,46 @@
+appId: swmansion.enriched.markdown.example
+tags:
+  - input
+  - unordered-list
+  - ordered-list
+  - block
+  - markdown-shortcuts
+---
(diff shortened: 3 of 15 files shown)


packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/editing/MarkdownShortcutMatcher.kt
```diff
@@ -0,0 +1,47 @@
...
+    if (prefix.isEmpty()) return null
+    val first = prefix[0]
+
+    if (first == '#') {
+      if (prefix.length > 6 || !prefix.all { it == '#' }) return null
+      val type = BlockType.forHeadingLevel(prefix.length) ?: return null
+      return Match(type, prefix.length)
+    }
+
+    if (prefix.length == 1 && (first == '-' || first == '*' || first == '+')) {
```
eszlamczyk, line 35: Nitpick:


```suggestion
    if (prefix.length == 1 && first in "-*+") {
```

eszlamczyk (changes requested):
Hi @cvburgess thanks for your pull request! The changes looks sound but I have some issues:

1. The API shape
With functionalities of `TextInput` getting extended (with for example blockquotes) some users might wat to enable only some of the shortcuts, so API like:
```ts
type MarkdownShortcutsConfig = {
    heading?: boolean;
    unorderedList?: boolean;
    orderedList?: boolean;
  };
  markdownShortcuts?: boolean | MarkdownShortcutsConfig;
```
(with the boolean optional) would be better. Following this we could cover all cases with:
- false/undefined = all off
- true = all on
- config object = custom subset

Then we can normalise false/true/undefined to full object and push it into the native side (so native doesn't require union types)

2. Cross-platform parity: the shortcut trigger is gated differently on Android vs iOS.

The two convert paths don't agree on when to fire:
- Deletion/replace: Android skips when deletedLength != 0 (EditPipeline.kt:157); pure insertions only. iOS never checks it (it's computed at EnrichedMarkdownTextInput.mm:1680 but not passed into applyMarkdownShortcut...), so iOS also converts on replacement edits; autocorrect commits, paste, or type-over-selection landing a #␣/-␣.
- Which line is checked: iOS additionally gates on the pre-edit cursor block (preEditBlockType != Paragraph), Android only on the resolved line (blockStartingAt(lineStart)). For a multi-line paste like "\n# ", Android converts and iOS doesn't.

Happy-path (single keystroke) is identical, so thats why the tests pass

Suggestion: pick the intended behavior, pass deletedLength through on iOS to match (or drop Android's guard)

cvburgess:
Sorry, I was on vacation - coming back to this now, hoping to have a commit soon!

packages/react-native-enriched-markdown/android/src/main/java/com/swmansion/enriched/markdown/input/editing/MarkdownShortcutMatcher.kt
cvburgess: Applied!

cvburgess:
Changes made @eszlamczyk - running the code locally to make sure it works as expected in addition to the tests

cvburgess:
Tested locally and it works as expected @eszlamczyk 

In future PRs I'd love to see if I can get typeahead support for quotes, code blocks, and other formatting to get even closer to parity with other editors.

eszlamczyk (approved):
Hi @cvburgess everything looks good, thanks for your contributions!

PR #803, opened 2026-09-11
feat(ios): add admonition support

maksg:
Render GitHub alerts (> [!NOTE], [!TIP], [!IMPORTANT], [!WARNING], [!CAUTION], behind Md4cFlags(admonitions: true)) as a blockquote variant: a bold title paragraph tinted per type, the matching octicon drawn by the decoration view in the column the title reserves, and accent bars tinted per nesting level. Themeable via Admonition(.note).foregroundStyle(...) and .background(...), defaulting to the GitHub palette shared with the React Native package with no fill. The alert round-trips through copy as Markdown (> [!NOTE] marker), source-faithful slicing, and copy as HTML (the web renderer's icon header).

The icon paths are byte-identical to the RN web source of truth, parsed by a Swift port of ENRMSVGPath cut down to the commands the octicons use; a test compares the copy against the web file when present.

Along the way, BlockquoteRenderer stops clobbering nested quotes' attributes and lays out paragraphs individually; quotes inside list items start on their own line and keep their own column instead of being restyled by the item; and bars and fills now run through paragraph spacing, which also closes the gap under headings in quotes.

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):

hryhoriiK97 (approved):
LGTM! One minor follow-up: `ListMarkerDrawer` now also draws admonition icons - worth renaming it (e.g. `ParagraphMarkerDrawer` or similar) in a follow-up PR to reflect the broader responsibility.

PR #808, opened 2026-09-11
feat(ios): add spoiler support

maksg:
Render `||spoiler||` text concealed behind a particle or solid overlay, revealed by tap, matching the React Native renderer.

- SpoilerRenderer tags the run; SpoilerConcealment makes the text transparent in a post-pass and stashes its color for the reveal, so wrappers rendered after the spoiler (strong, links) stay concealed
- SpoilerOverlayManager places one overlay per TextKit 2 segment and hit-tests taps; reveal flows through MarkdownRenderStore like the task-list toggle, keyed by spoiler ordinal so it survives theme re-renders
- Spoiler() theme element (.color, .background, .particleDensity, .particleSpeed, .solidBorderRadius) and .markdownSpoilerOverlay
- Links inside a concealed spoiler ignore taps and menus until revealed
- Copy as Markdown emits the `||` markers, before and after reveal
- Example app: spoiler theme entry and solid overlay in the playground

### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (approved):
LGTM! 🎖️ 

The internal design already has a clean factory seam for custom overlays: `SpoilerOverlayView` is a base class with a `prepareRevealAnimation()` hook, and `makeOverlay` in the manager is the single creation point. I think it's worth opening this up in a follow-up PR:

- Make `SpoilerOverlayView` `open`
- Extract a `SpoilerOverlayProvider` protocol around `makeOverlay(style:charRange:)`
- The existing `.particles` / `.solid` become built-in providers; power users subclass

This would let consumers build their own overlay effects (blur, shimmer, branded animations) without us having to add enum cases for every variation. The code is already structured for it — it's mostly a visibility change.

@maksg what do you think? Could you pick this up in a follow-up?

PR #815, opened 2026-09-15
feat(ios): let apps supply custom spoiler overlays

maksg:
The overlay concealing spoiler text could only be the built-in particle field or solid box. Open the seam so an app can draw its own effect.

- SpoilerOverlayView is open: animateReveal is the override point, and concealedText carries the segment's text for effects that show it through, since the text under an overlay renders transparent
- SpoilerOverlayProvider builds the views; .particles and .solid become built-in providers and markdownSpoilerOverlay takes any of them, replacing the MarkdownSpoilerOverlay enum
- Overlays rebuild on provider equality, so a provider with parameters keeps them in stored properties
- Segment frames now come with their character range, which is what slices the text each overlay gets
- The example playground cycles particles, solid, and a custom blur

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):
LGTM

Very cool change!

hryhoriiK97 (approved):

PR #818, opened 2026-09-15
fix(web): prevent native props from reaching the DOM

Lyczeq:
### What/Why?

On web, `testID` and `allowFontScaling` were forwarded to the root `<div>`, which caused React invalid DOM prop warnings.

Map `testID` to `data-testid` and consume `allowFontScaling` before forwarding valid HTML attributes. The same behavior applies to the parse-error fallback.

### Testing

- `yarn workspace react-native-enriched-markdown test --runInBand --no-watchman` (39 tests)
- `yarn typecheck`
- `yarn lint` (no errors; one existing inline-style warning)
- `yarn prepare`

### PR Checklist

- [ ] E2E tests are passing


packages/react-native-enriched-markdown/src/web/EnrichedMarkdownText.tsx
```diff
@@ -29,6 +29,8 @@ export const EnrichedMarkdownText = ({
   selectable = true,
   dir,
   selectionColor,
+  testID,
+  allowFontScaling: _allowFontScaling,
```
hryhoriiK97, line 33: Thanks for the fix, this looks good. FYI for anyone reading: other native-only props (`flavor`, `streamingConfig`, `contextMenuItems`, `numberOfLines`, …) still fall through `...rest` in shared cross-platform code, so we'll handle stripping the full set in a follow-up.

cc @eszlamczyk

packages/react-native-enriched-markdown/src/types/MarkdownTextProps.web.ts
```diff
@@ -12,6 +12,17 @@ export interface EnrichedMarkdownTextProps extends Omit<
...
+   * @platform web
+   */
+  testID?: string;
+  /**
+   * Accepted for React Native compatibility. Browsers handle font scaling
+   * through operating system accessibility settings.
+   * @platform web
+   */
+  allowFontScaling?: boolean;
   /**
```
hryhoriiK97, line 26: Nit: the native type tags this `@platform ios, android` and WEB.md lists it as ignored on web, so `@platform web` here reads as if it does something. Suggest matching the native tag and noting "ignored on web".

docs/WEB.md
```diff
@@ -30,6 +30,7 @@ All core `EnrichedMarkdownText` features are supported on web, including:
 | Prop | Description |
 |---|---|
 | `dir` | Sets the text direction on the root container (`'ltr'`, `'rtl'`, or `'auto'`). CSS logical properties in the renderers automatically flip layout for RTL. |
+| `testID` | Sets `data-testid` on the root container. |
```
hryhoriiK97, line 33: `testID` is a standard `ViewProps` prop on every platform, so it's not web-only. Maybe move it to the supported list as "`testID` maps to `data-testid`".

hryhoriiK97 (commented):
Thank you for the fix! I've left a few comments.

PR #819, opened 2026-09-15
fix(ios): stop selection handle drags from firing link presses

maksg:
Selecting the last word of a line parks the selection's end knob on top of the following line. Where a link sits under it, pressing the knob to drag it fired the link instead of moving the handle.

UIKit asks the delegate for a link's menu configuration as soon as the touch lands, before the range-adjustment gesture claims it, and discards the answer when the drag wins. The handleLinkLongPress fallback (no long-press handler means fire onLinkPress) turned that speculative ask into a real user-visible callback.

Record where each touch goes down and suppress the link callbacks when it started on a selection knob, derived from caretRect(for:) at both ends of the selection.

MarkdownTextView moves to its own file: the addition pushed MarkdownTextViewRepresentable.swift past SwiftLint's 600-line limit.

### PR Checklist

- [ ] E2E tests are passing


hryhoriiK97 (commented):
/

hryhoriiK97 (approved):

PR #821, opened 2026-09-15
feat(android): add spoiler support to enriched-markdown-android

jnowakow:
### What/Why?

Ports spoilers from `react-native-enriched-markdown` to the standalone `enriched-markdown-android`
package. Both overlays come across: drifting particles (default) and a solid rounded block, chosen
with `spoilerOverlay`. Tapping a spoiler reveals it, along with any spoilers adjacent to it.

`Spoiler` AST nodes already reached the renderer — `MD_FLAG_SPOILERS` is always on in the shared C++
core — but had no entry in the renderer map, so they fell through to `TextRenderer` and rendered as
**nothing**. `||text||` was silently dropped before this PR.

Two deliberate differences from the RN package:
- `SpoilerStyle.backgroundColor` is explicit (the particle overlay paints it over the concealed text
  before fading out). Left unset it walks the view tree, but under Compose the background usually
  lives on a `Modifier`, which inference cannot see.
- `solidBorderRadius` is stored in pixels like every other length here, rather than being
  density-multiplied at paint time.

Selection keeps the `||…||` markers, matching iOS (#808): copying concealed text yields
`||secret||`, not `secret`.

### Testing

`:ui:testDebugUnitTest :compose:testDebugUnitTest` — 233 tests, 0 failures (28 new spoiler cases).

### Test 1
1. Go to the "Text" screen
2. Scroll to "Spoiler Examples"
3. Tap a concealed span — the particles fade out and reveal the text

### Test 2
1. Long-press a revealed spoiler and copy it
2. Go to the "Playground" screen, tap "Set Raw Markdown", paste, tap "Set"
3. The pasted `||…||` renders concealed again

### PR Checklist

- [ ] Code compiles and runs on iOS (not applicable — iOS shipped in #808)
- [ ] Ran example app to verify changes (installed on an emulator; not yet verified on screen)
- [ ] E2E tests are passing


packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/renderer/SpanStyleCache.kt
```diff
@@ -31,6 +31,11 @@ class SpanStyleCache(
   val superscriptBaselineOffsetScale: Float = style.superscriptStyle.baselineOffsetScale
   val subscriptFontScale: Float = style.subscriptStyle.fontScale
   val subscriptBaselineOffsetScale: Float = style.subscriptStyle.baselineOffsetScale
+  val spoilerColor: Int = style.spoilerStyle.color
+  val spoilerBackgroundColor: Int? = style.spoilerStyle.backgroundColor
+  val spoilerParticleDensity: Float = style.spoilerStyle.particleDensity
+  val spoilerParticleSpeed: Float = style.spoilerStyle.particleSpeed
+  val spoilerSolidBorderRadius: Float = style.spoilerStyle.solidBorderRadius
```
jnowakow: Since there are two strategies for spoilers, wouldn't it be better to have data classes representing spoiler styles in a discriminative way? How RN does it?

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerDrawContext.kt
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * The color the particle overlay paints over the concealed text before fading it out, so it
+     * has to match whatever the text sits on.
+     *
+     * The styled value wins when it is set. Otherwise the view tree is walked upwards for the
+     * first opaque [ColorDrawable] background, and white stands in when there is none. Inference
+     * misses the common Compose case — a background declared as a `Modifier` on the `AndroidView`
+     * wrapper is not a view background — which is exactly why
+     * [com.swmansion.enriched.markdown.styles.SpoilerStyle.backgroundColor] exists.
+     */
```
jnowakow: Redundant comment

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/styles/SpoilerStyle.kt
```diff
@@ -0,0 +1,35 @@
+package com.swmansion.enriched.markdown.styles
+
+/**
+ * Styling of the overlay that conceals `||spoiler||` text until it is tapped.
+ *
+ * [particleDensity] and [particleSpeed] are unitless tuning knobs (they are divided by the base
+ * values the particle field was authored against), matching the React Native package. Unlike that
+ * package, [solidBorderRadius] is stored in **pixels** like every other length here, so the solid
+ * overlay does not re-apply the display density when it paints.
+ */
```
jnowakow: Unnecessary comments in this file

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/conversion/HTMLGenerator.kt
```diff
@@ -816,6 +827,21 @@ object HTMLGenerator {
...
+    if (isSpoiler) {
+      html
+        .append("<span data-spoiler=\"true\" style=\"background-color: ")
+        .append(styles.spoilerColor)
+        .append("; color: ")
+        .append(styles.spoilerColor)
+        .append("; border-radius: ")
+        .append(styles.spoilerBorderRadius)
+        .append("px;\">")
+    }
```
jnowakow: I think that RN ignores spoilers in HTMLGenerator. I think it should be ignored too.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlayDrawer.kt
```diff
@@ -0,0 +1,153 @@
...
+  private var cachedStyle: SpoilerStyle? = null
+
+  var spoilerOverlay: SpoilerOverlay
+    get() = currentMode
+    set(value) {
+      if (currentMode == value) return
+      strategy.stop()
+      currentMode = value
+      strategy = createStrategy(value)
+      cachedStyle?.let { strategy.applyStyle(it) }
```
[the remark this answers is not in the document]
jnowakow: 1cb8bdba

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlayDrawer.kt
```diff
@@ -0,0 +1,153 @@
...
+            ctx.paddingTop,
+          ) ?: continue
+        val key = SegmentKey(spanIdentity, line)
+        activeKeys.add(key)
+
+        strategy.drawSegment(canvas, ctx, key, rect)
+      }
+    }
+
+    strategy.pruneStaleSegments(activeKeys)
```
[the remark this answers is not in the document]
jnowakow: 1cb8bdba

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlayDrawer.kt
```diff
@@ -0,0 +1,153 @@
...
+    if (spans.isEmpty()) return null
+    return SpoilerDrawContext(
+      textView = textView,
+      layout = layout,
+      text = text,
+      spans = spans,
+      paddingLeft = textView.totalPaddingLeft.toFloat(),
+      paddingTop = textView.totalPaddingTop.toFloat(),
+      fontMetrics = layout.paint.fontMetrics,
+      backgroundColor = SpoilerDrawContext.resolveBackgroundColor(textView, cachedStyle?.backgroundColor),
```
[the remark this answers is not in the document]
jnowakow: 1cb8bdba

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
eszlamczyk, line 1: Just as a food on thought (not for this pr) but this file is getting bigger and bigger, and we still have a bunch of features to do. Maybe It would be worth to refactor each style patch into its own separate file? something like

```
  compose/style/patch/TextStylePatch.kt        (TextStylePatch + TextStyleScope)
  compose/style/patch/LinkStylePatch.kt        (LinkStylePatch + LinkStyleScope)
  compose/style/patch/SpoilerStylePatch.kt     (SpoilerStylePatch + 3 spoiler scopes)
...
```

Right now someone looking for spoiler styling has either grep the big file or know that its on line ~1200, that way he could just go to correct StylePatch file

eszlamczyk (approved):
LGTM!

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerDrawContext.kt
```diff
@@ -0,0 +1,38 @@
...
+        return styledBackgroundColor
+      }
+
+      var view: View? = textView
+      while (view != null) {
+        val color = (view.background as? ColorDrawable)?.color
+        if (color != null && Color.alpha(color) > 0) return color
+        view = view.parent as? View
+      }
+      return Color.WHITE
```
hryhoriiK97: Under Compose this walk usually ends here or at the window background. I probed three setups: no `ColorDrawable` anywhere → white; window background is a `ColorDrawable` → that color; any other drawable → white. The common Material 3 setup keeps a light XML theme and does dark mode in Compose, so the backdrop comes out light on dark surfaces unless every consumer sets `spoiler { backgroundColor }` (the example app had to, twice).

Minimum: resolve `android.R.attr.colorBackground` from `textView.context.theme` as the last fallback. Better: do what the iOS package does and draw the concealed text transparent (alpha from reveal progress in `SpoilerSpan.updateDrawState`), which removes the backdrop, the `backgroundColor` knob and this inference entirely.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spans/SpoilerSpan.kt
```diff
@@ -0,0 +1,33 @@
...
+  fun markRevealing() {
+    revealing = true
+  }
+
+  fun markRevealed() {
+    revealed = true
+    revealing = false
+  }
+
+  override fun updateDrawState(tp: TextPaint) {
```
hryhoriiK97, line 32: This overrides the size of any inline code inside the spoiler. Rendered `paragraph(spoiler(code("x")))` with code size 11 and paragraph 16: spans arrive as `CodeSpan, SpoilerSpan`, the spoiler runs last, and the code resolves to 16. Plain `paragraph(code("x"))` resolves to 11. Suggest not touching `textSize` here (the draw context already measures the band with the span's own metrics), or only setting it when the paint is still at the view default. A test next to `inlineStylesInsideASpoilerSurvive` asserting the resolved size would lock it in.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlay.kt
```diff
@@ -0,0 +1,17 @@
...
+
+enum class SpoilerOverlay(
+  internal val createStrategy: (SpoilerAnimator) -> SpoilerStrategy,
+) {
+  PARTICLES({ animator -> ParticleStrategy(animator) }),
+  SOLID({ _ -> SolidStrategy() }),
+  ;
+
+  companion object {
+    fun fromString(value: String?): SpoilerOverlay =
```
hryhoriiK97: Nothing in the standalone package parses strings; this is an RN leftover with no callers. Remove it or make it `internal`.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlay.kt
```diff
@@ -0,0 +1,17 @@
+package com.swmansion.enriched.markdown.spoiler
+
+enum class SpoilerOverlay(
```
hryhoriiK97, line 3: Follow-up PR, not this one: iOS exposes `SpoilerOverlayProvider` so apps can ship their own overlay, and the seam already exists here as `SpoilerStrategy`. Proposed shape:

```kotlin
sealed interface SpoilerOverlay {
  object Particles : SpoilerOverlay
  object Solid : SpoilerOverlay
}

interface CustomSpoilerOverlay : SpoilerOverlay {
  fun createRenderer(host: SpoilerOverlayHost): SpoilerOverlayRenderer
}
```

`SpoilerOverlayRenderer` is today's `SpoilerStrategy`, with `SpoilerAnimator` hidden behind a small host (`invalidate()`, `requestFrames()`) and an abstract base that fades by default, like iOS's `animateReveal`. The `compose` module can add a `DrawScope` adapter over the native canvas so custom overlays are written with Compose `Color` and `Brush`. Custom overlays stay canvas-drawn, since they paint inside `TextView.onDraw`, which is fine for the "different look" use case. The sealed interface also replaces the screaming-case `PARTICLES` / `SOLID`. Could you open an issue for it so it doesn't get lost?

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
```diff
@@ -1222,3 +1223,107 @@ class TableStyleScope {
...
+
+  /** Base drift speed of the particles. */
+  var speed: Float? = null
+}
+
+/** Tuning for the solid overlay (`spoilerOverlay = SpoilerOverlay.SOLID`). */
+@MarkdownStyleDsl
+class SpoilerSolidStyleScope {
+  /** Corner radius of the rectangles the solid overlay draws. */
+  var borderRadius: Dp? = null
```
hryhoriiK97: `codeBlock` and `table` call this `cornerRadius`. Let's not add another `borderRadius` to the DSL.

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
```diff
@@ -1222,3 +1223,107 @@ class TableStyleScope {
...
+}
+
+@MarkdownStyleDsl
+class SpoilerStyleScope {
```
hryhoriiK97: The block itself fits the DSL nicely: `particles { }` and `solid { }` mirror `blockquote { admonitions { } }`, and `color` / `backgroundColor` are the names every other scope uses. One nit: give the three new scopes an `internal constructor`, since nothing outside the DSL should build them (the older scopes have the same gap, fine to leave those).

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/EnrichedMarkdownText.kt
```diff
@@ -45,6 +50,7 @@ fun EnrichedMarkdownText(
   onLinkLongPress: ((String) -> Unit)? = null,
   onTaskListItemPress: ((TaskListItemPressEvent) -> Unit)? = null,
   enableTaskListItemToggle: Boolean = true,
+  spoilerOverlay: SpoilerOverlay = SpoilerOverlay.PARTICLES,
```
hryhoriiK97, line 53: Passing this as a composable parameter beside `selectable` and `flags` is the right call for this package, and the `typealias` re-export matches `Md4cFlags`. No change needed. If the follow-up lands the sealed interface, the default becomes `SpoilerOverlay.Particles` and nothing else here moves.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/LinkLongPressMovementMethod.kt
```diff
@@ -139,6 +156,64 @@ class LinkLongPressMovementMethod : ArrowKeyMovementMethod() {
...
+    for (span in spans) {
+      drawer.revealSpan(span) {
+        remaining[0]--
+        if (remaining[0] <= 0) widget.invalidate()
+      }
+    }
+    widget.invalidate()
+  }
+
+  private fun expandContiguousSpoilers(
```
hryhoriiK97, line 189: This reveals every touching spoiler together, matching RN. iOS reveals one at a time (`MarkdownRenderStore.revealSpoiler` passes a single ordinal). Either is fine, but the two standalone packages now disagree. Please state the choice in the PR description so iOS can align or document it.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/conversion/MarkdownExtractor.kt
```diff
@@ -362,6 +364,11 @@ object MarkdownExtractor {
     if (linkSpans.isNotEmpty()) {
       result = "[$result](${linkSpans[0].url})"
     }
+    // Outermost, matching the iOS package: copying a spoiler keeps it concealed in the markdown
+    // even though the selection itself reads the revealed text.
+    if (hasSpoiler) {
```
hryhoriiK97, line 369: Copy as Markdown keeps the markers, good. Plain Copy still yields the concealed text, since the characters are real. Same on iOS per its README, so not a blocker, but worth one README line so consumers know a concealed spoiler is not a secret.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/styles/SpoilerStyle.kt
```diff
@@ -0,0 +1,15 @@
+package com.swmansion.enriched.markdown.styles
+
+data class SpoilerStyle(
+  val color: Int = DEFAULT_COLOR,
+  val backgroundColor: Int? = null,
+  val particleDensity: Float = DEFAULT_PARTICLE_DENSITY,
+  val particleSpeed: Float = DEFAULT_PARTICLE_SPEED,
+  val solidBorderRadius: Float = 0f,
```
hryhoriiK97: Class default is 0, theme default is 4dp. No production path hits this today (`DefaultStyles` and `StyleConfigMerger` both pass an explicit style), so either drop the class defaults or add a comment that they are placeholders.

packages/enriched-markdown-android/README.md
```diff
@@ -132,6 +132,7 @@ The `markdownStyle` builder supports these blocks:
 | `inlineImage` | Inline images |
 | `thematicBreak` | Horizontal rules |
 | `table` | Tables |
+| `spoiler` | The overlay that conceals `\|\|spoiler\|\|` text |
```
hryhoriiK97, line 135: Please document the block's properties like the others: `color`, `backgroundColor` (with the backdrop caveat from the KDoc), `particles { density, speed }`, `solid { cornerRadius }`.

packages/enriched-markdown-android/README.md
```diff
@@ -296,6 +299,7 @@ Creates a style that tracks `MaterialTheme.colorScheme` changes. Use inside `Mat
 - Task lists (`- [ ]` / `- [x]`, tap to toggle — see `onTaskListItemPress`)
 - Links and images (block and inline)
 - Thematic breaks (`---`)
+- Spoilers (`||hidden||`, tap to reveal)
```
hryhoriiK97: Two behaviours worth stating here, both matching iOS: screen readers read concealed text as ordinary text, and adjacent spoilers reveal together (if that stays the choice).

hryhoriiK97 (commented):
Nice port, and the three places where you went beyond a straight copy (`finishReveal()`, per-span metrics, alpha-aware particles) are all real improvements. I ran the ui and compose tests on the head, all green, and diffed the spoiler package against the RN original.

Two things I'd like fixed before merge, both verified with a small probe:
- the particle backdrop falls back to white or the window color under Compose, so dark apps get light rectangles unless they set `backgroundColor` by hand;
- `SpoilerSpan` forces the block font size, so inline code inside a spoiler renders at 16 instead of its own 11.

The Compose surface fits the existing DSL well. One design item, custom overlays like iOS has, is worth a follow-up PR rather than this one. Everything else is docs and small naming alignments, inline below.

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
jnowakow: Agree, it will be much easier to read and navigate 👍 I'll open follow-up PR with this refactor

packages/enriched-markdown-android/README.md
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/README.md
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/styles/SpoilerStyle.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/conversion/MarkdownExtractor.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/EnrichedMarkdownText.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/compose/src/main/java/com/swmansion/enriched/markdown/compose/MarkdownStylePatches.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlay.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spans/SpoilerSpan.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerDrawContext.kt
jnowakow: https://github.com/software-mansion/enriched-markdown/pull/821/commits/a2bb2fafc2cd5d4cc21ca60a0bddb53b8d53bd5c

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/spoiler/SpoilerOverlay.kt
jnowakow: Issue for reference: https://github.com/software-mansion/enriched-markdown/issues/888

PR #825, opened 2026-09-16
feat(ios): add maxHeight, aspectRatio and resizeMode to block images

maksg:
### PR Checklist

- [ ] E2E tests are passing


maksg:
https://github.com/user-attachments/assets/72468a58-f98c-47ec-971f-2e3d16f41048

eszlamczyk (approved):

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Theme/Elements/Image.swift
```diff
@@ -14,6 +17,30 @@ public struct BlockImage: MarkdownThemeContent {
         return copy
     }
 
+    /// Fits the image into at most `value` points of height, keeping its
+    /// aspect ratio. Takes precedence over `height`; `0` clears it.
+    public func maxHeight(_ value: CGFloat) -> Self {
```
hryhoriiK97, line 18: Suggest replacing the three independent fields with one `sizing` value so the modifiers are mutually exclusive by construction, last one wins, like SwiftUI modifier order:

```swift
public enum ImageSizing: Equatable, Sendable {
    case height(CGFloat)
    case maxHeight(CGFloat)
    case aspectRatio(CGFloat)
}
```

`.height(_:)`, `.maxHeight(_:)` and `.aspectRatio(_:)` each set `sizing`. That removes the precedence table and the "pass `0` to clear" sentinel: a theme layer that sets any sizing replaces the lower layer's sizing outright. Please also add the SwiftUI overloads `aspectRatio(_ size: CGSize)` and `aspectRatio(_:contentMode:)`.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Theme/Elements/Image.swift
```diff
@@ -14,6 +17,30 @@ public struct BlockImage: MarkdownThemeContent {
...
+    /// e.g. `16 / 9`. Takes precedence over `height` and `maxHeight`; `0` clears it.
+    public func aspectRatio(_ value: CGFloat) -> Self {
+        var copy = self
+        copy.aspectRatio = value
+        return copy
+    }
+
+    /// How the image fills its box. Defaults to `.cover` for `maxHeight` and
+    /// `aspectRatio` boxes, and to fill-width drawing for a fixed `height`.
+    public func resizeMode(_ value: ImageResizeMode) -> Self {
```
hryhoriiK97: `resizeMode` is React Native vocabulary. SwiftUI and UIKit both call this `contentMode`, and SwiftUI's `Image.ResizingMode` means something different (tile/stretch), so `ImageResizeMode` reads as a near-collision. Suggest `.contentMode(_ value: ImageContentMode)`.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Theme/ImageStyle.swift
```diff
@@ -0,0 +1,72 @@
+import UIKit
+
+/// How a block image's bitmap fills the box laid out for it.
+///
+/// Leaving `ImageStyle.resizeMode` unset keeps the legacy drawing: the image is
+/// scaled to the box width, centered vertically, and any overflow is clipped.
+public enum ImageResizeMode: String, Equatable, Sendable {
```
hryhoriiK97: Case names are CSS `object-fit` values. On Apple platforms they map awkwardly, and `.center` actively misleads because UIKit's `.center` never scales, while this one scales down. Proposed:

```swift
public enum ImageContentMode: Equatable, Sendable, CaseIterable {
    case fit        // was contain
    case fill       // was cover
    case stretch
    case scaleDown  // was center
    case original
    case fitWidth   // the current nil "legacy" drawing, made explicit
}
```

Making the legacy drawing an explicit `fitWidth` case removes the optional from the public surface, lets the Playground cycle `allCases`, and makes the default theme readable as `.height(200)` plus `.fitWidth`. `CaseIterable` matches `AdmonitionType`. The `String` raw value is only used for the cache key, so it can stay or become internal.

hryhoriiK97 (commented):
The implementation is in good shape: sizing policy isolated in `ImageBoxSizing`, drawing in `ImageDrawing`, a minimal observer protocol for re-measure, and thorough tests. I ran the four affected suites on an iPhone 16 simulator (61 passed) and confirmed with a hosted probe that SwiftUI does shrink the view after a `maxHeight` image settles.

My main ask is on the public surface. The three-knob shape (`height` / `maxHeight` / `aspectRatio` with a precedence table and `0` as "clear") and the `resizeMode` vocabulary are ported from the React Native style object. Since the goal for this package is the best SwiftUI API rather than parity, I'd like to land this with a SwiftUI-shaped surface instead. Details inline on `Image.swift` and `ImageStyle.swift`. The internal `ImageBoxSizing` enum is already the right shape, so this is mostly renaming and moving it public.

Two small correctness items (stale bitmap race, and the zero sentinel, which the redesign removes) plus a few doc and alignment nits.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Theme/ImageStyle.swift
```diff
@@ -0,0 +1,72 @@
...
+    case cover
+    /// Fills the box exactly, ignoring the aspect ratio.
+    case stretch
+    /// Draws centered at its own size, scaled down only when it exceeds the box.
+    case center
+    /// Draws centered at its own size, never scaled. Overflow is cropped.
+    case original
+}
+
+public struct ImageStyle: Equatable, Sendable {
```
hryhoriiK97, line 33: With the above, `ImageStyle` becomes `sizing: ImageSizing?` and `contentMode: ImageContentMode?`. `ImageStyle.height` is existing public API, so consider keeping a `height` computed convenience on it to avoid a source break. Defaults stay as today: default theme `.height(200)` with `.fitWidth`, and `.fill` implied when a theme sets `maxHeight` or `aspectRatio` without a mode.

Separately, on file placement: every other style struct lives in `MarkdownStyleConfig.swift`. Either keep `ImageStyle` and `InlineImageStyle` there, or say in the PR that this starts a per-style split. Let's not leave one exception.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Attachments/ImageBoxSizing.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// A box fitted to the image's own proportions, capped at this height.
+    case maxHeight(CGFloat)
+    /// A box whose height is the available width over this ratio.
+    case aspectRatio(CGFloat)
+
+    /// Stands in wherever the real height cannot be known: no theme height at
+    /// all, or a ratio box before the container width is.
+    static let fallbackHeight: CGFloat = 200
+
+    init(style: ImageStyle) {
```
hryhoriiK97: `height` is passed straight through, so `.height(0)` yields a zero-height box while `.maxHeight(0)` means "unset". Inconsistent, and the README says `0` clears "a modifier" without qualification. With the `ImageSizing` enum this goes away entirely; if the knobs stay, treat all three the same and add a test for `height(0)`.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Attachments/MarkdownImageAttachment.swift
```diff
@@ -167,13 +220,12 @@ final class MarkdownImageAttachment: NSTextAttachment {
...
+                box: box,
                 borderRadius: self.cachedBorderRadius
             )
 
             if let processed {
-                Self.processedImageCache.setObject(processed, forKey: processedKey as NSString)
+                Self.processedImageCache.setObject(processed, forKey: key as NSString, cost: processed.byteCost)
             }
 
             DispatchQueue.main.async {
```
hryhoriiK97, line 233: Stale bitmap race. `lastProcessedBox` is set before the scaling is dispatched, and this completion assigns `loadedImage` unconditionally. Two quick box changes (rotation while a `maxHeight` image loads) can complete out of order on the concurrent queue, the older bitmap lands last, and the `box == lastProcessedBox` guard above then keeps it forever. The same shape existed with `lastProcessedKey`, but responsive boxes make this a normal path. Suggest:

```swift
DispatchQueue.main.async {
    guard self.lastProcessedBox == box else { return }
    ...
}
```

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Attachments/MarkdownImageAttachment.swift
```diff
@@ -137,24 +170,44 @@ final class MarkdownImageAttachment: NSTextAttachment {
         }
     }
 
+    /// Callbacks land on the main queue, except a synchronous hit on the
+    /// original-image cache, which arrives on the render queue while the
+    /// document is still being built — before any layout, hence the
+    /// `lastLaidOutBox` guard.
     private func handleLoadedImage(_ image: UIImage?) {
```
hryhoriiK97, line 176: This runs on the render queue on a synchronous original-cache hit and writes `originalImage` there. The `lastLaidOutBox` guard makes it safe today, but the RN native attachment simply hops to main when off-thread. Doing the same here would let us drop the comment-based reasoning. Non-blocking.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Views/MarkdownTextView.swift
```diff
@@ -295,6 +296,22 @@ final class MarkdownTextView: UITextView, SelectionHandleTouchReporting {
         setNeedsLayout()
     }
 
+    /// Attachments that resize after loading have no other way to reach this view.
+    private func adoptImageAttachments(in attributedText: NSAttributedString) {
```
hryhoriiK97, line 301: This walks every attribute run on each `setMarkdownAttributedText`, which streaming callers hit at frame rate (the comment above cites 30 re-assignments a second). It's cheap next to the `attributedText` assignment, but worth a line saying so. Non-blocking.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Views/MarkdownTextView.swift
```diff
@@ -295,6 +296,22 @@ final class MarkdownTextView: UITextView, SelectionHandleTouchReporting {
...
+        let full = NSRange(location: 0, length: attributedText.length)
+        let options: NSAttributedString.EnumerationOptions = .longestEffectiveRangeNotRequired
+        attributedText.enumerateAttribute(.attachment, in: full, options: options) { value, _, _ in
+            (value as? MarkdownImageAttachment)?.layoutObserver = self
+        }
+    }
+
+    /// The cached measurement is wrong now, not merely stale.
+    func attachmentDidInvalidateLayout() {
```
hryhoriiK97, line 310: The tests cover the UIView layer only. Since the whole point is SwiftUI re-measuring after `invalidateIntrinsicContentSize`, a test that hosts `MarkdownTextViewRepresentable` in a `UIHostingController` inside a window, completes a `DeferredImageDownloader`, and asserts the hosted frame shrinks by 50 would lock that in. I ran exactly that locally and it passes (153 → 103).

packages/enriched-markdown-ios/README.md
```diff
@@ -390,6 +390,48 @@ Images load from these sources:
...
+
+Block images fill the width available to them. Three `BlockImage` modifiers decide how tall the box they fill is:
+
+| Modifier | Box height |
+|----------|------------|
+| `.height(_:)` | Fixed. The default, at 200 points. |
+| `.maxHeight(_:)` | The image's own proportions at the current width, capped at this value. |
+| `.aspectRatio(_:)` | The width divided by this ratio, e.g. `16 / 9`. |
+
+Set more than one and `aspectRatio` wins over `maxHeight`, which wins over `height`. Passing `0` clears a modifier, so a theme layered over another can switch its responsive sizing back off.
```
hryhoriiK97: With the enum-based sizing this paragraph reduces to "the last sizing modifier applied wins, and a layer that sets one replaces the lower layer's". If the knobs stay, please state that `0` clears only `maxHeight` and `aspectRatio`.

packages/enriched-markdown-ios/README.md
```diff
@@ -390,6 +390,48 @@ Images load from these sources:
...
+
+`.resizeMode(_:)` decides how the image fills that box:
+
+| Mode | Drawing |
+|------|---------|
+| `.contain` | Scaled to fit inside the box, never cropped |
+| `.cover` | Scaled to fill the box, cropping what overflows |
+| `.stretch` | Fills the box exactly, ignoring the image's proportions |
+| `.center` | Centered at its own size, scaled down only when it exceeds the box |
```
hryhoriiK97: Please add the UIKit / SwiftUI equivalents next to each mode, and a one-line mapping to the RN style names (`contain`/`cover`/`center`/`none`) for people using both packages. Also note that with `.fit` (today's `.contain`) the rounded corners follow the drawn image, not the box.

apps/ios-example/EnrichedMarkdownExample/EnrichedMarkdownExample/Views/Playground/PlaygroundScreen.swift
```diff
@@ -2,6 +2,33 @@ import EnrichedMarkdown
...
+        switch self {
+        case .height: return ("Height 200", BlockImage().height(200))
+        case .maxHeight: return ("Max 150", BlockImage().maxHeight(150))
+        case .aspectRatio: return ("16:9", BlockImage().aspectRatio(16 / 9))
+        }
+    }
+}
+
+/// Legacy fill-width drawing first, then every explicit mode.
+private let imageResizeModeCycle: [ImageResizeMode?] = [
```
hryhoriiK97: Becomes `ImageContentMode.allCases` once the enum is `CaseIterable` and the legacy drawing is an explicit case.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Theme/ImageStyle.swift
maksg: MarkdownStyleConfig went over lint's line limit so that's why it was seperated. I will move the rest of them to it's own files in the next PR

hryhoriiK97 (approved):

Issue #827, opened 2026-09-16
[iOS] Inline code background is offset when code is taller than the paragraph line


MaryanPrydatko:
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a supported version of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

On iOS, when an inline code run is taller than the paragraph line (for example `code.fontSize: 22` with `paragraph.lineHeight: 20`, at the default text size), the inline code background is drawn in the wrong place. Every box in that paragraph is shifted up and to the left of its glyphs by a constant offset: the box starts partway through the preceding word, and its top border runs through the line above.

When the code run fits inside the line, the boxes are drawn correctly. So the trigger looks like code glyphs taller than `maximumLineHeight`, not font scaling. `ios/utils/CodeBackground.m` builds each rect from `boundingRectForGlyphRange:` plus the draw origin and only grows short rects to the reference height, so the mismatch probably comes from the line-height clamp or baseline offset applied to the paragraph. I have not traced it further.

This is separate from the inline code font scaling issue, but that issue makes it easier to hit: once inline code scales with Dynamic Type, a tight `lineHeight` reaches this state at larger text sizes.

### Library version

1.0.1 (unchanged on `main` at 124ce3b)

### Environment info

```shell
iOS Simulator, iPhone 17 Pro, iOS 26.5
react-native 0.86.2 (New Architecture), Expo SDK 57
```

### Steps to reproduce

1. Render a paragraph with a few inline code spans and a code font taller than the line:
   ```tsx
   <EnrichedMarkdownText
     markdown={'Call `forward` for you and records every operation so `loss.backward()` can compute gradients.'}
     markdownStyle={{
       paragraph: { fontSize: 15, lineHeight: 20 },
       code: { fontSize: 22, backgroundColor: '#1E1E1E', color: '#9CDCFE' },
     }}
   />
   ```
2. Run on iOS at the default text size.
3. The code backgrounds are offset up and left from the code text.

### Reproducible example repository

Reproduces in `apps/react-native-example` with the markdown style above.

eszlamczyk:
Hi @MaryanPrydatko thanks for the bug report! I'll look into this today.

PR #828, opened 2026-09-16
fix(ios): scale inline code font with Dynamic Type

### LINKED ISSUE #826: [iOS] Inline code ignores Dynamic Type / allowFontScaling

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a supported version of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

On iOS, inline code (`` `code` ``) does not scale with Dynamic Type / `allowFontScaling`, while the surrounding paragraph does.

- Below the default text size, the paragraph font and `lineHeight` shrink but inline code keeps its unscaled size. The code gets taller than the line, and its background covers the lines above and below.
- Above the default text size, body text grows and inline code stays small.

This affects every app on a non-default text size, not only apps that set `markdownStyle.code.fontSize`: with no `code.fontSize` set, the renderer falls back to the unscaled `blockStyle.fontSize`.

**Cause.** `ios/renderer/CodeRenderer.m` creates the code font with an explicit, unscaled size:

```objc
CGFloat codeFontSize = _config.codeFontSize > 0 ? _config.codeFontSize : blockStyle.fontSize;
// ...
codeFont = [RCTFont updateFont:nil withFamily:codeFontFamily size:@(codeFontSize) /* ... */ scaleMultiplier:1.0];
// or
codeFont = [UIFont monospacedSystemFontOfSize:codeFontSize weight:weight];
```

Block fonts go through `-[RenderContext cachedFontForSize:family:weight:]`, which applies `RCTFontSizeMultiplierWithMax(maxFontSizeMultiplier)` when `allowFontScaling` is on. `StrongRenderer`, `EmphasisRenderer` and `LinkRenderer` also pass `scaleMultiplier:1.0`, but with `size:nil` on top of the already scaled current font, so they inherit the scale. `CodeRenderer` is the only one that passes an explicit unscaled size.

A fix is ready in a PR.

### Library version

1.0.1 (the code is unchanged on `main` at 124ce3b)

### Environment info

```shell
iOS Simulator, iPhone 17 Pro, iOS 26.5
react-native 0.86.2 (New Architecture), Expo SDK 57
Reported first by a user on a physical iPhone.
```

### Steps to reproduce

1. Render inline code in a paragraph with a custom size and line height:
   ```tsx
   <EnrichedMarkdownText
     markdown={'Treat the model like a function: `output = my_model(input_data)`. More text follows here.'}
     markdownStyle={{
       paragraph: { fontSize: 15, lineHeight: 20 },
       code: { fontSize: 15, backgroundColor: '#1E1E1E', color: '#9CDCFE' },
     }}
   />
   ```
2. Set a small text size: `xcrun simctl ui booted content_size extra-small` (or Settings > Accessibility > Display & Text Size > Larger Text).
3. Relaunch the app. Body text shrinks, inline code does not, and the code background overlaps the neighboring lines.
4. Set `content_size extra-extra-large` and relaunch. Body text grows, inline code stays small.

Reproduces in this repo's `apps/react-native-example` with any inline code and a non-default Dynamic Type size. The fix PR adds a regression story for it.

MaryanPrydatko:
### What/Why?

Closes #826

`CodeRenderer` created the inline code font with an explicit, unscaled size (`markdownStyle.code.fontSize`, or the unscaled `blockStyle.fontSize` fallback) and `scaleMultiplier:1.0`. The surrounding text scales through `-[RenderContext cachedFontForSize:family:weight:]`, so at any non-default Dynamic Type size inline code had the wrong size relative to its paragraph. Below the default size the code outgrew the scaled line height, and its background covered the neighboring lines.

The fix applies the multiplier the block fonts already use, `allowFontScaling ? RCTFontSizeMultiplierWithMax(maxFontSizeMultiplier) : 1.0`, in both branches: the custom `fontFamily` path (as `scaleMultiplier`) and the `monospacedSystemFontOfSize` fallback. `allowFontScaling={false}` and `maxFontSizeMultiplier` keep working because the multiplier comes from the same context values. `BlockLineHeight` is left alone because `blockFont` is already scaled.

Related: #827 (a separate `CodeBackground` offset when a code run is taller than the paragraph line). This PR does not fix it. Once inline code scales, apps with a tight `lineHeight` can reach that state at large text sizes.

### Testing

- Verified in a consumer app (Expo SDK 57, react-native 0.86.2, iOS 26.5 simulator) with the same change applied as a patch on 1.0.1. `CodeRenderer.m` is unchanged between 1.0.1 and `main`.
  - Dynamic Type `extra-small`, `large` (default) and `extra-extra-large`, light and dark mode.
  - Before: inline code stayed at its unscaled size, overlapping the lines above and below at `extra-small`.
  - After: inline code scales with the paragraph at every size, and the default size is unchanged (multiplier 1.0).
- Added `Regression / Inline Code Font Scaling (#826)`: a paragraph with inline code next to a plain `<Text>` control, plus an `allowFontScaling={false}` row. Change the simulator text size and the inline code should grow and shrink with the body, except in the last row.
- `eslint` and `prettier` on the story and `clang-format` (repo version) on `CodeRenderer.m` pass, and so do the repo's lefthook pre-commit and commitlint hooks.
- I did not build `apps/react-native-example` locally.

### PR Checklist

- [ ] Code compiles and runs on Android (iOS-only change)
- [ ] Updated documentation/README if applicable (n/a)
- [ ] Ran example app to verify changes (verified in a consumer app, see Testing)
- [ ] E2E tests are passing


eszlamczyk (approved):
Hi @MaryanPrydatko thanks for the bug report and the solution, looks clean! Thank you for your contribution

PR #829, opened 2026-09-16
feat(ios): resolve writing direction per paragraph

maksg:
Renderers never set a paragraph's baseWritingDirection, so every decoration that reads it from the paragraph style (blockquote bars, list bullets and numbers, task checkboxes and their hit test, admonition icons) stayed on the app's leading side for Arabic or Hebrew content, while TextKit right-aligned the text itself. Copied HTML never carried a dir attribute either.

Add a WritingDirectionResolver post-pass that stamps each paragraph after a render, mirroring the React Native package's ENRMApplyWritingDirection- Mode: first-strong detection over the same Unicode ranges, code blocks skipped, neutral paragraphs falling back to the view's layout direction. It runs on the document and on every table cell, which is drawn from its own attributed string.

The mode comes from a new .markdownWritingDirection modifier with the same four values as the RN prop (first-strong by default), and the renderer entry points take it alongside the layout direction. The HTML copy flavor now reads its dir from the first copied paragraph.

Admonition titles deviate from the RN package deliberately: the title is the renderer's own text, so it follows the direction of the body it introduces rather than resolving from "Note". Resolving it separately left an LTR title row with its icon and bar on the left sitting above an RTL body barred on the right, which read as two boxes.

#### Screenshots
Before
Screenshot iPhone 17 Pro Max 16-09-2026 at 15 24 01

After
Screenshot iPhone 17 Pro Max 16-09-2026 at 16 05 36

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk (approved):

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Views/Environment+WritingDirection.swift
```diff
@@ -0,0 +1,39 @@
...
+/// React Native package's `writingDirection` prop.
+public enum MarkdownWritingDirection: Equatable, Sendable {
+    /// Each paragraph follows its first strong directional character;
+    /// paragraphs without one (digits, punctuation) follow the SwiftUI
+    /// layout direction. The default, matching Android.
+    case firstStrong
+    /// Leaves direction to TextKit. List markers, checkboxes, and
+    /// blockquote bars then follow the app's interface direction rather
+    /// than each paragraph's content.
+    case auto
```
hryhoriiK97: This is documented as "leaves direction to TextKit", and the value it produces is `NSWritingDirection.natural`. UIKit's word for it is `natural`; `auto` is the CSS and RN spelling. Since this package aims for a platform-native API, `.natural` reads right to an iOS developer, and the RN prop keeps `'auto'` on its side. Naming it now also avoids a second rename if the follow-up on `WritingDirectionResolver` lands.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Rendering/WritingDirectionResolver.swift
```diff
@@ -0,0 +1,96 @@
+import UIKit
+
+/// Stamps `baseWritingDirection` on every paragraph after a render, as the
+/// React Native package does: the block chrome (blockquote bars, list
+/// markers, checkboxes, admonition icons) and the task-list hit test read it
+/// from the paragraph style, where `.natural` would mean the app's interface
+/// direction rather than the paragraph's. Code blocks keep the
+/// `.leftToRight` their renderer sets.
+enum WritingDirectionResolver {
```
hryhoriiK97, line 9: Follow-up proposal, not for this PR. UIKit's header defines `NSWritingDirectionNatural` as "determines direction using the Unicode Bidi Algorithm rules P2 and P3", which is first-strong. So TextKit already resolves every `.natural` paragraph the way this resolver does, and the text was right before this PR; only the chrome and the HTML export were reading `.natural` as "app direction".

That means the text and the chrome are now resolved by two different algorithms, and they can disagree. I checked on the simulator: a paragraph starting with U+200F (which people insert to force RTL) is RTL to TextKit but `.natural` here because it is not a letter; Adlam letters are RTL to TextKit and LTR here because they sit outside the three ranges. In both cases the text flows one way and the bullet sits on the other side, which is the bug this PR fixes.

The platform-native fix: leave paragraphs `.natural` and have the drawers ask TextKit what it did. `NSTextLayoutManager.baseWritingDirection(at:)` returns the resolved direction (verified: RTL for both cases above), and the drawers already have the layout manager. Forced `.leftToRight` / `.rightToLeft` still stamp the paragraph style as now. The HTML generator is the one consumer without a layout; it can keep a first-strong scan or emit `dir="auto"`. With that, `.firstStrong` and `.auto` collapse into `.natural`, the enum mirrors `NSWritingDirection` one to one, and the Unicode table goes away. Happy to open an issue if you agree.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Views/Environment+WritingDirection.swift
```diff
@@ -0,0 +1,39 @@
+import SwiftUI
+
+/// How each paragraph's base writing direction is resolved. Mirrors the
+/// React Native package's `writingDirection` prop.
+public enum MarkdownWritingDirection: Equatable, Sendable {
```
hryhoriiK97: Nit: `CaseIterable` is cheap here and consistent with where the other public enums are heading.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Rendering/WritingDirectionResolver.swift
```diff
@@ -0,0 +1,96 @@
...
+            return
+        case .leftToRight:
+            direction = { _, _ in .leftToRight }
+        case .rightToLeft:
+            direction = { _, _ in .rightToLeft }
+        case .firstStrong:
+            let fallback: NSWritingDirection = layoutDirection == .rightToLeft ? .rightToLeft : .leftToRight
+            direction = { attrs, paragraphRange in
+                let detected = attrs[MarkdownAttribute.admonitionHeader] == nil
+                    ? firstStrongDirection(of: string.substring(with: paragraphRange))
```
hryhoriiK97: This allocates a string per paragraph on every render, including streaming renders. Scanning `output.string.unicodeScalars` over the converted range would avoid the copies. Not measurable at typical sizes, so fine as a follow-up, and moot if the TextKit-backed resolution lands.

packages/enriched-markdown-ios/README.md
```diff
@@ -357,6 +357,23 @@ extension View {
...
+```swift
+extension View {
+  func markdownWritingDirection(_ direction: MarkdownWritingDirection) -> some View  // default .firstStrong
+}
+```
+
+| Value | Behavior |
+|-------|----------|
+| `.firstStrong` (default) | Each paragraph follows its first strong directional character; paragraphs without one (digits, punctuation) follow the SwiftUI `layoutDirection`. Matches Android and the React Native package. |
+| `.auto` | Leaves direction to TextKit; list markers, checkboxes, and blockquote bars follow the app's interface direction rather than the paragraph's content. |
```
hryhoriiK97: Worth one more clause: under this mode the chrome falls back to `UIView.appearance().semanticContentAttribute` (see `TextLayoutHelpers.paragraphIsRTL`), so a subtree with `.environment(\.layoutDirection, .rightToLeft)` still gets LTR markers. Only `.firstStrong` honours the SwiftUI environment. Pre-existing, but this table is where someone will look.

packages/enriched-markdown-ios/README.md
```diff
@@ -357,6 +357,23 @@ extension View {
...
+| `.leftToRight` | Forces every paragraph left-to-right. |
+| `.rightToLeft` | Forces every paragraph right-to-left. |
+
+Code blocks always render left-to-right. See [Right-to-left text](#right-to-left-text) for what follows a paragraph's direction.
```
hryhoriiK97, line 375: `MarkdownRenderer.render` and `renderLaTeX` gained `writingDirection:` and `layoutDirection:` but only the modifier is documented. A line here, or under the LaTeX section where `renderLaTeX` is already described, would cover the UIKit entry points.

hryhoriiK97 (commented):
LGTM. One thing before merge: rename `.auto` to `.natural`, since that's what it maps to in UIKit. The rest inline is optional, including a proposal for a TextKit-backed follow-up.

packages/enriched-markdown-ios/Sources/EnrichedMarkdown/Rendering/WritingDirectionResolver.swift
hryhoriiK97: Tracked as a follow-up in #874, so this thread can be resolved without holding up the PR.

hryhoriiK97 (approved):

PR #830, opened 2026-09-16
feat(core): C++ markdown repair module for streaming

hryhoriiK97:
## What

Adds `packages/core/cpp/repair`, a C++ port of the incomplete-markdown repair that `react-native-streamdown` runs today in JavaScript (the `remend` package on a worklet runtime). It closes constructs left open at the end of a streamed prefix (`**bold`, `[link](http://`, `$$x`, `~~text`, and so on) so the parser renders the prefix the way the finished document will.

This is the first of a series of PRs. **Nothing here is compiled into an app build yet**: the podspec, SwiftPM manifest and Android CMake globs are untouched, so no shipped binary changes. Follow-ups wire it into the RN package, the native libraries and wasm, and then remove the JS dependency and the duplicated ObjC/Kotlin streaming filters.

## Why native

- We are adding streaming support to `enriched-markdown-ios` and `enriched-markdown-android`. Those libraries have no JS runtime, so a JavaScript repair step is not an option for them at all; the only place repair can live and serve every consumer is the shared C++ core. That is the primary reason this module goes into `packages/core` rather than into the RN package.
- For RN it removes the reason `react-native-streamdown` needs `react-native-worklets` Bundle Mode (Babel + Metro setup), which is its main adoption friction.
- The block-level streaming filter already exists twice (`StreamingMarkdownFilter.m` and `.kt`). Once repair lives in the core, both can be replaced by one implementation.

## Decisions

**Naming.** The module is `MarkdownRepair`; the upstream name is not used as an identifier anywhere. Apache-2.0 grants the code but explicitly not the trademark, and a neutral name keeps our divergences ours. Attribution lives in exactly two places: the header comment of `MarkdownRepair.hpp` and `LICENSE.remend`, which is what the license requires for a derivative work.

**No dependencies.** Nothing in the repo depends on the npm package. The reference's own test suite was run once at port time with a logging wrapper, and every call it makes is baked into `tests/repair/ReferenceCases.cpp` as C++ data (756 cases, plus our own for Unicode neighbours, option toggles and both link modes). doctest is fetched by CMake at configure time, pinned by tag and checksum, not vendored.

**Unicode.** The word-character test needs Unicode letter and number categories, which plain C++ cannot classify. `UnicodeTables.hpp` is a generated range table (770 ranges, ~6 KB) produced by `scripts/generate-unicode-tables.mjs` from V8's own `\p{L}\p{N}` classification, so it matches the reference exactly and records the Unicode version it came from. ICU, platform callbacks or libc would make behaviour depend on the host OS and are unavailable on wasm; this is the same pattern md4c and tree-sitter use for their own tables.

**Parity first, then deliberate divergences.** The port started as a byte-for-byte reproduction, verified against the reference's own recorded test suite. Review then surfaced behaviours that are wrong for our parser, and those are now fixed on purpose, each listed in the `MarkdownRepair.hpp` header and marked `// ours` in the recorded cases:

- Closers nest in reverse opening order, so `**bold `code` becomes `**bold `code`**` instead of `**bold `code**``.
- `**` and `__` close even when an italic is nested inside; the reference left them open.
- Emphasis before or inside an incomplete link closes correctly; the reference stopped at the placeholder.
- An unterminated HTML tag is stripped only at the start of a line. Inline HTML is disabled in our parser, so `if x<y then` is prose.
- A `[` that starts a task-list item, an admonition (`> [!NOTE`), a footnote or a reference link is not an incomplete link. Every checklist line used to flash a placeholder link.
- Trailing CRLF counts as a trailing newline.
- An opener before the last blank line, or on a heading line that has ended, is not closed: an inline span cannot cross a block boundary, so the closer would be a stray marker in a later block.
- New closers for md4c extensions the reference does not have: spoilers (always on in our parser), highlight, superscript and subscript, behind `RepairOptions` flags to be wired from the parser flags.

What is still mirrored on purpose: JS strings are UTF-16 and the reference inspects marker neighbours as single code units, so astral characters never count as word characters there; the port keeps that view for parity. The `handlers` option (arbitrary JS callbacks) is not ported.

**Structure.** Split by concern mirroring the reference's source layout (`RepairEmphasis.cpp`, `RepairHandlers.cpp`, `RepairInternal.*`) so a future upstream sync is a file-by-file diff. Handlers mutate the text in place through a `RepairContext` that builds the code-block, math and inline-code lookups once per run and drops them only on mutation. No global state.

**Performance.** ~143 µs per call on a 10 KB document and ~1.5 ms on 100 KB on an Apple Silicon Mac (`-O2`), linear scaling. An ASCII fast path in `isWordChar` and lookups shared across handlers keep it there. Well under a millisecond per token for typical answers, on the render queue.

## Tests

`yarn core:test` configures, builds and runs everything through ctest: 22 test cases, 935 assertions, clean with `-Wall -Wextra -Wpedantic`. Test areas live in `tests/<area>/` with one executable each, so a `parser/` area later is a one-line addition.

CI: new `core-tests` job gated on `packages/core/**`, required in `ci-success`. First run on Linux/GCC happens with this PR.

## Next PRs

1. Block-level streaming filter (table, code block, block math modes) ported into the same module, with the open-fence flag.
2. RN wiring: build globs, bridges, `streamingConfig.repair` prop, inert placeholder links, delete the ObjC and Kotlin filters.
3. Streaming API for the native iOS and Android libraries.
4. wasm export and web support.
5. streamdown drops `remend` and the worklet runtime.


packages/core/cpp/repair/MarkdownRepair.hpp
```diff
@@ -0,0 +1,69 @@
...
+//
+// Source layout, mirroring the reference's:
+//   MarkdownRepair.cpp   pipeline and public helpers        (index.ts, utils.ts)
+//   RepairEmphasis.cpp   bold / italic / bold-italic        (emphasis-handlers.ts)
+//   RepairHandlers.cpp   every other handler                (*-handler.ts)
+//   RepairInternal.*     UTF-8, JS character classes, scans (utils.ts, code-block-utils.ts)
+//
+// The block-level streaming filter (table / code block / block math modes,
+// today in the platform-specific StreamingMarkdownFilter implementations) is
+// not part of this module yet; see docs/MARKDOWN_REPAIR_PLAN.mdx, phase 2.
```
jnowakow: There's no such file as `dosc/MARKDOWN_REPAIR_PLAN.mdx`

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
```diff
@@ -0,0 +1,201 @@
...
+  auto underscore = handler(RepairHandlers::italicSingleUnderscore);
+  auto doubleUnderscore = handler(RepairHandlers::italicDoubleUnderscore);
+  CHECK(asterisk("*it") == "*it*");
+  CHECK(asterisk("hello*world") == "hello*world");
+  CHECK(asterisk("* item") == "* item");
+  CHECK(asterisk("2 * 3") == "2 * 3");
+  CHECK(underscore("_it") == "_it_");
+  CHECK(underscore("snake_case") == "snake_case");
+  CHECK(underscore("_it\n\n") == "_it_\n\n");
+  CHECK(underscore("**bold _und**") == "**bold _und_**");
```
jnowakow: Is this example correct? Looks like an bug to me that underscore formatting is applied inside already closed bold formatting 🤔

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
```diff
@@ -0,0 +1,201 @@
...
+  CHECK(inlineKatex("$x") == "$x$");
+  CHECK(inlineKatex("costs $5 and $6") == "costs $5 and $6");
+  CHECK(repair("$$ a * b") == "$$ a * b$$"); // asterisk inside math stays
+}
+
+TEST_CASE("html tags and comparison operators") {
+  auto htmlTags = handler(RepairHandlers::htmlTags);
+  auto comparison = handler(RepairHandlers::comparisonOperators);
+  CHECK(htmlTags("text <cus") == "text");
+  CHECK(htmlTags("text <b>bold</b") == "text <b>bold");
```
jnowakow: Is this correct that closing that was is removed? I would expect it to be the other way round: 

```suggestion
  CHECK(htmlTags("text <b>bold") == "text <b>bold</b");
```

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
```diff
@@ -0,0 +1,201 @@
...
+  repairInlineMarkdownInPlace(text, RepairOptions());
+  CHECK(text == repair("**bold [link](http://x"));
+  text = "";
+  repairInlineMarkdownInPlace(text, RepairOptions());
+  CHECK(text.empty());
+}
+
+TEST_CASE("full pipeline: mixed") {
+  CHECK(repair("This is **bold with *ital") == "This is **bold with *ital*");
+  CHECK(repair("Text **bold `code") == "Text **bold `code**`");
```
jnowakow: This doesn't look like a correct markdown. I haven't checked how reference implementation behaves in this example but I would expect correct fix to be: ```**bold `code`**```

jnowakow (commented):
Looks solid 💯 I only wonder if it works as expected in three places so please confirm if it is fine

packages/core/cpp/repair/MarkdownRepair.hpp
hryhoriiK97: Good catch, the plan doc lives outside the repo. Reference removed in 7a52d6cd.

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
hryhoriiK97: Correct behaviour, but the test was confusing because it fed the underscore handler the intermediate state. While streaming `**bold _und`, the bold handler appends `**` first, then the underscore opened inside the bold closes inside it: `**bold _und_**`. That is also the upstream test case. I replaced the handler-level check with the full-pipeline case so the intent is visible (7fc3f1df).

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
hryhoriiK97: You made me look at this properly. Our parser sets `MD_FLAG_NOHTMLSPANS`, so inline HTML is never rendered and the reference's stripping only risked eating prose like `if x<y then`. The handler now strips a tag only when it starts a line, which covers the one HTML block we honour (`<video>`) while streaming. Inline `<` is left alone. Tests added (7fc3f1df).

packages/core/cpp/tests/repair/RepairHandlerTests.cpp
hryhoriiK97: Agreed, `**bold `code`**` is the right output. This was the reference appending closers in handler order; we now insert them in reverse opening order so they nest, and the same fix makes bold close when an italic is nested inside and makes emphasis around an incomplete link close correctly. The affected recorded cases are marked `// ours` in `ReferenceCases.cpp`, and the divergences are listed in the `MarkdownRepair.hpp` header (7fc3f1df).

packages/core/cpp/repair/MarkdownRepair.hpp
```diff
@@ -0,0 +1,101 @@
...
+  bool bold = true;
+  bool boldItalic = true;
+  bool comparisonOperators = true;
+  bool htmlTags = true;
+  bool images = true;
+  bool inlineCode = true;
+  bool inlineKatex = false;  // opt-in upstream too: `$` is ambiguous with currency
+  bool italic = true;
+  bool katex = true;
+  bool links = true;
```
eszlamczyk: Nit: `katex`/`inlineKatex` are upstream's names but in our repository "KaTeX" is ambigious with web renderer. We use `latexMath` everywhere else. Not sure if this is worth changing in the whole code (maybe for future's sake). But at least leaving the comment that would break this ambiguity would be worth while.

maksg:
Overall looks awesome! 🎉
However I found some cases that are failing and could be worth fixing
```
// The italic handlers anchor the closer at the first `*`/`_` in the text instead of the
// open one (RepairEmphasis.cpp:295, :318), so with the earlier-block rule an italic in any
// previous paragraph disables italic repair for the rest of the stream. Bold is fine.
TEST_CASE("italic closer is anchored at the open marker") {
  CHECK(repair("This is *important*.\n\nNext *point") == "This is *important*.\n\nNext *point*"); // today: unchanged
  CHECK(repair("This is _important_.\n\nNext _point") == "This is _important_.\n\nNext _point_"); // today: unchanged
  CHECK(repair("# *Title*\nsome *text") == "# *Title*\nsome *text*");                             // today: unchanged
  CHECK(repair("_ x **y _z") == "_ x **y _z_**");                                                 // today: _ x **y _z**_
}

// An escaped `\[` is repaired as a link opener (RepairHandlers.cpp:208). LLMs stream
// `\[ … \]` for display math, so the placeholder shows up as literal text.
TEST_CASE("an escaped bracket is not a link") {
  CHECK(repair("Solve \\[ x^2 + y^2") == "Solve \\[ x^2 + y^2"); // today: Solve \[ x^2 + y^2](streamdown:incomplete-link)
  CHECK(repair("see \\[a](http://x") == "see \\[a](http://x");   // today: see \[a](streamdown:incomplete-link)
  RepairOptions o;
  o.linkMode = LinkMode::TextOnly;
  CHECK(repair("Solve \\[ x^2", o) == "Solve \\[ x^2"); // today: Solve \ x^2
}

// Both math handlers use `rfind` as the opener (RepairHandlers.cpp:365, :400), which can be
// half of a `$$`, an escaped `\$` or a `$$` inside inline code, so the earlier-block check
// and the nesting order work from the wrong index.
TEST_CASE("math closers are anchored at their opener") {
  RepairOptions o;
  o.inlineKatex = true;
  CHECK(repair("It costs $5.\n\nThe formula $$x^2$$", o) == "It costs $5.\n\nThe formula $$x^2$$"); // today: ...$$x^2$$$
  CHECK(repair("$a **b \\$5", o) == "$a **b \\$5**$"); // today: $a **b \$5$**
  CHECK(repair("$$a\n\n`$$` b") == "$$a\n\n`$$` b");   // today: $$a\n\n`$$` b$$
  CHECK(repair("$$a **b `$$`") == "$$a **b `$$`**$$"); // today: $$a **b `$$`$$**
}

// A CRLF blank line is not a block boundary (RepairInternal.cpp:438), although a trailing
// CRLF already is.
TEST_CASE("a CRLF blank line is a block boundary") {
  CHECK(repair("**a\r\n\r\nb") == "**a\r\n\r\nb");           // today: **a\r\n\r\nb**
  CHECK(repair("[link\r\n\r\nmore") == "[link\r\n\r\nmore"); // today: [link\r\n\r\nmore](streamdown:incomplete-link)
}

// countSingleUnderscores applies a flanking rule the reference does not have. Good change,
// but it is missing from the divergence list in MarkdownRepair.hpp. These pass today.
TEST_CASE("underscore flanking rule, pinned") {
  CHECK(repair("_a_ b_") == "_a_ b_"); // reference: _a_ b__
  CHECK(repair("_  b") == "_  b");     // reference: _  b_
}

// The task-list check accepts any word starting with x (RepairHandlers.cpp:225), so a link
// whose text starts with x is not repaired until its `](` arrives.
TEST_CASE("task-list check only matches a checkbox") {
  CHECK(repair("- [Xcode setup") == "- [Xcode setup](streamdown:incomplete-link)"); // today: unchanged
  CHECK(repair("- [x") == "- [x");                                                  // guard, passes today
}

// Optional follow-up: a fence or a heading after the opener ends its paragraph like a blank
// line does (RepairInternal.cpp:469). The reference has the same gap.
TEST_CASE("a fence or heading after the opener is a block boundary") {
  CHECK(repair("**a\n```\ncode") == "**a\n```\ncode");  // today: **a\n```\ncode**  the closer lands inside the code block
  CHECK(repair("**a\n# heading") == "**a\n# heading"); // today: **a\n# heading**
}

```

I actually prepared proposed fixes for those cases on branch `review/pr-830-proposed-tests` in case you want to take a look at them

hryhoriiK97:
@maksg thanks, great catches, and thank you for the branch! Cherry-picked your fixes with your authorship (ef03d337). On top: folded the tests into `RepairDivergenceTests.cpp`, listed the missing divergences in the `MarkdownRepair.hpp` comment (including the underscore flanking rule), and two small tidy-ups in your code (counters return a struct instead of an out-pointer; block boundary cached per run). Same behaviour, same speed, 954 assertions green.

packages/core/cpp/repair/RepairEmphasis.cpp
```diff
@@ -0,0 +1,334 @@
...
+  size_t &count = result.count;
+  scanOutsideFences(text, [&](size_t &i) {
+    if (text[i] != '_') {
+      return false;
+    }
+    const uint32_t prev = codePointBefore(text, i);
+    const uint32_t next = codePointAt(text, i + 1);
+    // shouldSkipUnderscore()
+    const bool skip = prev == '\\' || math.inside(i) || isWithinLinkOrImageUrl(text, i) || isWithinHtmlTag(text, i) ||
+                      prev == '_' || next == '_' || (isWordCharUnit(prev) && isWordCharUnit(next));
```
eszlamczyk: We potentially could squeeze a bit of performance here with reordering of the conditions, that would trigger the early return for cases like `snake_case_text`: 
```suggestion
    const bool skip = prev == '\\' || prev == '_' || next == '_' || (isWordCharUnit(prev) &&
  isWordCharUnit(next)) ||
                    math.inside(i) || isWithinLinkOrImageUrl(text, i) || isWithinHtmlTag(text, i);
```

jnowakow (approved):
💯

packages/core/cpp/repair/MarkdownRepair.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+  }
+  return inFence;
+}
+
+bool isWithinMathBlock(std::string_view text, size_t position) {
+  return MathLookup(text).inside(position);
+}
+
+bool isWithinLinkOrImageUrl(std::string_view text, size_t position) {
```
eszlamczyk, line 47: This function is problematic, everytime we call this we do a backward search (for each character we check if is inside a `(`. This is costly. I'd rather use something that does lookup while we go forward (storing position of last `(` then the check is O(1) instead of O(n)). SOmething that MathLookup and CodeLookup already seem to be doing

packages/core/cpp/tests/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+# or simply `yarn core:test` from the repo root.
+#
+# Layout: one subdirectory per area of the core (repair/, later parser/,
+# highlight/ ...), each compiled into its own test executable by
+# add_core_test() below, all sharing TestMain.cpp.
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Debug)
```
eszlamczyk: Defaulting to default might mean that CI only ever does `-O0` optimisation, while this ships inside optimized release builds on both platforms. 

Two consequences:
1. The configuration that actually ships is never tested
2. Possible performance assertions that might be worth to add would not give us the real numbers (tested on `64KB` document I got 4.8x faster times in production settings.

I'd suggest using `RelWithDebInfo` as the default, with a CI matrix that also builds `Debug`, so assert()-based perconditions still get exercised.

eszlamczyk (commented):
Nothing consumes this module yet, so this is the cheap moment to fix it.

The backward walks here are re-deriving the same line-local state for every delimiter they look at — `isWithinLinkOrImageUrl` and `isWithinHtmlTag` per `_`, and `lineBefore` (via `isBracketNotALink`) per `[`. That's O(delimiters × line length). Document length is fine, but streamed paragraphs are unwrapped, so line length grows with the paragraph: 8 paragraphs of 8000 chars is ~20ms per call, re-paid on every token.

All three answers can be carried forward in a single pass instead. For the HTML tag and the line start, a running flag/index is enough. The link URL needs slightly more care: the current function requires both a `](` behind and a closing `)` ahead on the same line, so a flag that flips on `](` alone isn't equivalent — `isWithinLinkOrImageUrl("[a](url _x", 8)` is `false` today and would become `true`. Worth flagging that neither recorded case covers an unclosed URL, so that change would pass CI silently.

`CompleteInlineCodeLookup` already does exactly the right thing for this shape — remember the opener, back-fill the range when the closer arrives. Same approach here gives an exact answer in one forward pass, no backward fallback needed. Clearing state on `\n` and `)` matches the current bail conditions.

Since `isWithinLinkOrImageUrl` is public and has parity cases, I'd keep it as-is and add the lookup as an internal fast path — then the corpus doubles as an oracle that the two agree.



Additionaly, maybe it would be worth to add performance assertions to the tests as mentioned in the comment in `CMakeLists.txt`

eszlamczyk:
I ran some benchmarks on this and prototyped the fix to see what it'd buy us. Numbers are desktop x86 at `-O2`, per **single repair call** - that's the unit that matters, since we call this once per streamed token and it has to fit inside a frame (16.67ms at 60fps, and repair only gets a slice of that - parsing, AST diffing and layout come out of the same budget).

Documents are realistic LLM generated prose: unwrapped paragraphs separated by a blank line. The three columns are the same byte count formatted differently.

### Baseline (this PR as-is)

| doc size | 600-char paras | 2000-char paras | one unbroken para |
|---|---|---|---|
| 4 KB | 0.17 ms | 0.36 ms | 0.66 ms |
| 8 KB | 0.32 | 0.69 | 2.40 |
| 16 KB | 0.62 | 1.38 | 9.21 |
| 32 KB | 1.26 | 2.84 | **36.38** |
| 64 KB | 2.47 | 5.56 | **145.37** |
| 128 KB | 4.97 | 11.16 | **577.15** |

The last column crosses a frame at ~20 KB. At 32 KB we'd drop 2 frames per token, at 64 KB about nine. Note it's the same content in all three columns — cost is driven by paragraph length, not document size.

### With the prototyped fix (hoisted cheap predicates + forward-state lookup replacing the backward walks)

| doc size | 600-char paras | 2000-char paras | one unbroken para |
|---|---|---|---|
| 4 KB | 0.10 ms | 0.09 ms | 0.09 ms |
| 8 KB | 0.19 | 0.18 | 0.18 |
| 16 KB | 0.37 | 0.36 | 0.37 |
| 32 KB | 0.74 | 0.73 | 0.72 |
| 64 KB | 1.50 | 1.50 | 1.45 |
| 128 KB | 3.10 | 3.07 | 3.06 |

The paragraph-length dimension disappears — all three columns collapse onto one line. That's the part I care about more than the 189× worst case: cost becomes predictable from document size alone, so we can actually set a budget. 16.67ms isn't reached until ~700 KB.

I checked the lookup is exactly equivalent to the current backward walks: 7.5M position-by-position comparisons over random inputs, 0 mismatches, plus the 756-case repair corpus and all 29 recorded helper calls green.

### What this doesn't fix

Per-call now fits a frame, but there's one call per token and per-call cost is still linear in document size, so the aggregate over a stream stays O(n²):

> 128 KB streamed at ~4 chars/token ≈ 32,000 calls × ~1.55 ms average ≈ **~50 seconds of CPU** over the response.

Every frame is fine, we just burn most of a core for the whole stream. Scoping the work to the trailing block is what fixes that:

| | whole doc | last block only (600 B) |
|---|---|---|
| 32 KB doc | 0.77 ms | 0.015 ms |
| 64 KB doc | 1.55 ms | 0.015 ms |
| 128 KB doc | 3.18 ms | 0.015 ms |

Constant in document size. A real implementation still needs one cheap prefix pass for fence/code state — measured at 0.178 ms over 128 KB, against the ~30 full passes we make today — so realistically ~0.2 ms per call at 128 KB instead of 3.1 ms, and the whole-stream total drops to ~6 s and scales linearly.

### Caveats

- Desktop numbers. A mid-range phone is materially slower, so we want several times headroom and a re-measure on device before committing to any threshold.
- The trailing-block figure is a proxy (repairing the tail standalone) — it's the shape of the win, not a committed number. Unlike the other two, I haven't validated that change against the corpus.

hryhoriiK97:
@eszlamczyk thanks, great review. All four points are in: `LineContextLookup` replaces the backward walks, with the public helper kept as-is and an oracle test that the two agree on every position of every recorded case plus the unclosed-URL and nested-paren shapes; cheap predicates first; `katex`/`inlineKatex` renamed to `displayMath`/`inlineMath`; tests default to `RelWithDebInfo` with a Debug CI leg, and linearity/budget assertions in optimised builds.

Reproduced your tables here: the same quadratic column before (164 ms at 128 KB unbroken, this machine is ~3× faster than yours) and the same collapse after (1.4 to 1.6 ms at 128 KB for all three shapes).

On the O(n²) aggregate: agreed, and trailing-block scoping is the right fix, but it goes in the next PR as an incremental session. Counting parity over the tail alone changes results (`**a\n\n**b`), and the tilde/comparison escapes rewrite earlier paragraphs, so they need the previous repaired prefix kept rather than recomputed. It shares the fence-aware prefix pass the block filter needs, and gets an on-device measurement before any threshold.

eszlamczyk (approved):
One thing from my previus review that didn't make it: the third backward walk, `lineBefore` via `isBracketNotALink`, called per `[` in `findFirstIncompleteBracet`. Still quadratic in line length, but only in `LinkMode::TextOnly` path. Not the same thing as the O(n²) aggregate you deferred, and trailing-block scoping won't fix it - a one-paragraph doc is the trailing block. Could you add it to the Next PRs list as its own item? Same shape as the fix you just landed, and the new linearity test needs a bracket-heavy TextOnly case or it won't catch it either.

hryhoriiK97:
> One thing from my previus review that didn't make it: the third backward walk, `lineBefore` via `isBracketNotALink`, called per `[` in `findFirstIncompleteBracet`. Still quadratic in line length, but only in `LinkMode::TextOnly` path. Not the same thing as the O(n²) aggregate you deferred, and trailing-block scoping won't fix it - a one-paragraph doc is the trailing block. Could you add it to the Next PRs list as its own item? Same shape as the fix you just landed, and the new linearity test needs a bracket-heavy TextOnly case or it won't catch it either.

Done 👍 [268ea71](https://github.com/software-mansion/enriched-markdown/pull/830/commits/268ea71df8e7e7b5fe84438bd388336530185921)

maksg (approved):

PR #832, opened 2026-09-16
fix(ios): grow line to fit tall inline code so its background aligns

### LINKED ISSUE #827: [iOS] Inline code background is offset when code is taller than the paragraph line

eszlamczyk:
### What/Why?

Fixes #827: on iOS an inline `code` run taller than the paragraph `lineHeight` (e.g. `code.fontSize: 22`, `lineHeight: 20`) drew its background box offset up/left, with the box top running through the line above.

Change: `applyLineHeight` now sets a **floor instead of a clamp** - `minimumLineHeight = lineHeight` but `maximumLineHeight = 0` - so a line grows to fit its tallest run instead of clipping it. This also folds the old math/block-image special-case into one rule and applies the same fix to list items.

This intentionally **diverges from React Native**, which hard-clamps `maximumLineHeight = lineHeight`. RN must do that because `<Text>` has to honor an exact line height for Yoga measurement; we don't - we measure the real laid-out height (`usedRectForTextContainer`), so a grown line is measured and reserved correctly. It also brings iOS in line with Android, whose `LineHeightSpan` already floors rather than clamps (#643).

### Testing

Storybook (`apps/react-native-example`), iOS, default text size:
```
Call `forward` for you and records every operation so `loss.backward()` can compute gradients.
```
`paragraph: { fontSize: 15, lineHeight: 20 }`, `code: { fontSize: 22, backgroundColor: '#1E1E1E', color: '#9CDCFE' }`. Regression-checked paragraphs, lists, blockquotes, and headings (all shared the clamp).

| Before | After |
| - | - |
| image | image |

### PR Checklist

- [ ] E2E tests are passing


eszlamczyk:
Note: Maestro tests will be provided in separate issue to not clutter this issue

packages/react-native-enriched-markdown/ios/utils/ParagraphStyleUtils.m
```diff
@@ -264,47 +264,19 @@ void applyBlockSpacingAfter(NSMutableAttributedString *output, CGFloat marginBot
...
-                  }];
-#endif
-
-  BOOL hasBlockImage = ENRMRangeContainsBlockImage(output, range);
-
   NSMutableParagraphStyle *style = getOrCreateParagraphStyle(output, range.location);
 
   style.minimumLineHeight = lineHeight;
-  style.maximumLineHeight = (hasMath || hasBlockImage) ? 0 : lineHeight;
+  style.maximumLineHeight = 0;
```
hryhoriiK97, line 279: Code blocks use `applyLineHeight` too (`ENRMCodeBlockContent.mm:72`), but `ENRMCodeBlockContainerView` still reserves `lineCount * lineHeight` (line 95). With the floor a line can grow past that (e.g. an emoji in a Menlo 14 line at the default `lineHeight: 20`), so the last lines could get clipped or overlap the next block. Either keep the clamp for code blocks or measure with `usedRectForTextContainer`, and update the comment at `ENRMCodeBlockContainerView.m:70-72` ("every line fragment is exactly that tall" is no longer true).

packages/react-native-enriched-markdown/ios/utils/ParagraphStyleUtils.m
```diff
@@ -264,47 +264,19 @@ void applyBlockSpacingAfter(NSMutableAttributedString *output, CGFloat marginBot
...
-                  usingBlock:^(id value, __unused NSRange attrRange, BOOL *stop) {
-                    if ([value isKindOfClass:[ENRMImageAttachment class]] && !((ENRMImageAttachment *)value).isInline) {
-                      found = YES;
-                      *stop = YES;
-                    }
-                  }];
-  return found;
-}
-
+// Floor, not clamp: minimumLineHeight keeps short lines at lineHeight, while maximumLineHeight = 0 lets
```
hryhoriiK97, line 267: This also changes every block with `lineHeight` below the font's natural height (a tight paragraph now renders taller and no longer matches RN `<Text>`), and tall fallback scripts like Thai now break uniform spacing. Fine to accept, but please call it out as a behaviour change in the description/changelog.

packages/react-native-enriched-markdown/ios/renderer/ListItemRenderer.m
```diff
@@ -132,7 +132,7 @@ - (void)renderNodeContent:(MarkdownASTNode *)node
     style.headIndent = totalIndent;
     if (lineHeightConfig > 0) {
       style.minimumLineHeight = lineHeightConfig;
-      style.maximumLineHeight = ENRMRangeContainsBlockImage(output, range) ? 0 : lineHeightConfig;
+      style.maximumLineHeight = 0;
```
hryhoriiK97, line 135: `style` is freshly allocated, so `maximumLineHeight = 0` is a no-op. Drop it, or route through `applyLineHeight` so the rule lives in one place.

hryhoriiK97 (commented):
Thank you for the fix! I've left a few comments inline, the code block height one is the main blocker.

Small nit outside the diff: the comment at `ParagraphStyleUtils.m:294` ("Math paragraphs leave maximumLineHeight at 0") is stale now, and `MAX(maximumLineHeight, minimumLineHeight)` is effectively always the minimum, so it can be simplified.

Also, pre-existing: `applyBaselineOffset` bails for the whole paragraph when one run is taller than `lineHeight`, so all lines lose centering. Might be worth a per-line fix now that grown lines are the intended path.

cc @maksg — the Swift package still has the same clamp (`ParagraphStyleHelpers.swift:73`, list and table renderers too), so #827 should reproduce there. Does it make sense to port this fix there as well?

eszlamczyk:
Addressed the inline comments in the latest commits. For the pre-existing `applyBaselineOffset` whole-block centering point, I filed a separate follow-up: #869 (a correct fix is layout-dependent, so I kept it out of this PR).

PR #835, opened 2026-09-17
perf(android): hand the rendered buffer to the text view uncopied

jnowakow:
### What/Why?

Android copied every rendered buffer twice (`Renderer` froze it into a `SpannableString`, then `TextView.setText` copied it again). Each copy is O(spans²), about 28ms for a 10k-span document.

- The renderer now returns its `SpannableStringBuilder`, and a `Spannable.Factory` lets the text view use it without copying.
- The builder is wrapped in a plain `Spannable` so `Layout` keeps paragraph spans in insertion order. Without the wrapper, nested task checkboxes are drawn one indent too far right.
- Adds a `benchmark` module that measures display time on a device (full, render, layout and draw).

Results are in the table below: 
image

### Testing

Unit tests, including a new one for where nested checkboxes are painted. 

### PR Checklist

- [ ] E2E tests are passing


packages/enriched-markdown-android/display-benchmark/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  // Benchmarks must run non-debuggable.
+  testBuildType = "release"
+
+  buildTypes {
+    release {
+      minifyEnabled = false
+    }
+  }
```
jnowakow: Why there is no minification?

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextViewSetup.kt
```diff
@@ -2,13 +2,102 @@ package com.swmansion.enriched.markdown.utils.text.view
...
+ *
+ * `Layout.getParagraphSpans` special-cases `SpannableStringBuilder` and reads its
+ * paragraph spans in position order instead of insertion order. Layout paints
+ * `LeadingMarginSpan`s in that order, advancing x by each one's margin, and the
+ * list spans place their markers assuming the innermost (first inserted) span
+ * is painted first - in position order a nested checkbox lands an indent too far
+ * right, over its own text and outside the tappable margin. `LineHeightSpan`s
+ * are chained in that same order. Behind this wrapper Layout falls back to the
+ * public `getSpans`, which keeps insertion order, as a `SpannableString` would.
+ */
```
jnowakow: Is this comment really useful?

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextViewSetup.kt
```diff
@@ -2,13 +2,102 @@ package com.swmansion.enriched.markdown.utils.text.view
...
+ * The default factory copies the text into a fresh `SpannableString` on every
+ * `setText`, and `SpannableString.setSpan` scans the spans it already holds
+ * before appending, so that copy costs O(spans^2) - tens of milliseconds of main
+ * thread time for a long document. `Renderer` builds a private buffer per
+ * segment and nothing else holds a reference to it, so the view can adopt it
+ * without a defensive copy.
+ *
+ * A `SpannableStringBuilder` is wrapped in [InsertionOrderedSpannable] rather
+ * than adopted directly, see there for why.
+ */
```
jnowakow: Is this comment really useful?

packages/enriched-markdown-android/ui/src/test/java/com/swmansion/enriched/markdown/EnrichedMarkdownTest.kt
```diff
@@ -87,6 +87,22 @@ class EnrichedMarkdownTest {
...
+    val container = EnrichedMarkdown(context)
+    container.applyRenderedSegments(segments)
+    val child = container.getChildAt(0) as EnrichedMarkdownInternalText
+
+    // Freezing the buffer into a SpannableString costs a span copy quadratic in
+    // span count, so the view is set up to write through to the rendered one.
+    val marker = Any()
+    (child.text as Spannable).setSpan(marker, 0, 1, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+    assertEquals(0, rendered.styledText.getSpanStart(marker))
+  }
```
jnowakow: How does this test work?

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/segments/RenderedSegment.kt
```diff
@@ -11,7 +11,7 @@ sealed interface RenderedSegment {
   val signature: Long
 
   data class Text(
-    val styledText: SpannableString,
+    val styledText: Spannable,
     val imageSpans: List<ImageSpan>,
```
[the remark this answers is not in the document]
jnowakow: Clients relay on `compose`, not `ui` so the change is not breaking the API. Moreover, it will be published as a new version of the library so existing clients won't be affected.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/renderer/Renderer.kt
```diff
@@ -38,14 +37,22 @@ class Renderer {
     document: MarkdownASTNode,
     onLinkPress: ((String) -> Unit)? = null,
     onLinkLongPress: ((String) -> Unit)? = null,
-  ): SpannableString = renderContent(document.children, onLinkPress, onLinkLongPress)
+  ): SpannableStringBuilder = renderContent(document.children, onLinkPress, onLinkLongPress)
```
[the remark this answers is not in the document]
jnowakow: Clients relay on `compose`, not `ui` so the change is not breaking the API. Moreover, it will be published as a new version of the library so existing clients won't be affected.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/segments/RenderedSegment.kt
```diff
@@ -11,7 +11,7 @@ sealed interface RenderedSegment {
   val signature: Long
 
   data class Text(
-    val styledText: SpannableString,
+    val styledText: Spannable,
```
hryhoriiK97, line 26: Worth a one-line note here that this buffer becomes the view's live text: once displayed it carries the TextView's ChangeWatcher, the Editor's SpanController and the selection marks as spans (I confirmed this in a Robolectric run). Nothing reads segments after they land today, but anything that later caches, diffs or counts spans on a segment will see those extras.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextViewSetup.kt
```diff
@@ -2,13 +2,86 @@ package com.swmansion.enriched.markdown.utils.text.view
...
+      else -> SpannableString(source)
+    }
+}
+
+/**
+ * Hides that [buffer] is a `SpannableStringBuilder`. Layout reads a builder's paragraph
+ * spans in position order rather than insertion order, which paints nested list margins
+ * outer-first and pushes nested checkboxes an indent right.
+ */
+private class InsertionOrderedSpannable(
```
hryhoriiK97, line 39: Nit, doc only. Because this hides the builder type, every paragraph-span query from `Layout` now goes through the builder's sorted `getSpans` instead of the unsorted fast path the framework reserves for builders. Your `layout` benchmark already covers the cost, so no change needed, but a sentence in this KDoc would stop someone from "optimising" the wrapper away later.

packages/enriched-markdown-android/settings.gradle
```diff
@@ -25,3 +26,6 @@ rootProject.name = "enriched-markdown-android"
 include ":parser"
 include ":ui"
 include ":compose"
+// Measurement only, never published: the root subprojects block applies the publish
+// script to parser/ui/compose alone, and this module is absent from nmcpAggregation.
+include ":display-benchmark"
```
hryhoriiK97, line 31: The new module falls outside the root `build.gradle` subprojects filter (`parser`, `ui`, `compose`), so its Kotlin is never linted: `:display-benchmark:tasks --all` lists no ktlint tasks, `:ui` lists 56. Consider applying ktlint to it there while still keeping it out of the publish script and `nmcpAggregation`.

packages/enriched-markdown-android/ui/src/test/java/com/swmansion/enriched/markdown/TaskListInteractionTest.kt
```diff
@@ -445,7 +490,7 @@ class TaskListInteractionTest {
     return view
   }
 
-  private fun laidOutTextView(spannable: SpannableString): TextView =
+  private fun laidOutTextView(spannable: Spannable): TextView =
```
hryhoriiK97, line 496: This helper builds a plain `TextView`, whose default factory copies the builder into a `SpannableString`, so the 13 tests going through it exercise the old copied path rather than the no-copy wrapper. Only the 9 container-based tests cover the new path. Not blocking, but switching this helper to `EnrichedMarkdownInternalText` would make the whole task-list suite guard the change.

packages/enriched-markdown-android/display-benchmark/src/androidTest/java/com/swmansion/enriched/markdown/benchmark/DisplayBenchmark.kt
```diff
@@ -0,0 +1,322 @@
...
+   *
+   * `EnrichedMarkdown` has no entry point for a parsed document: it parses and renders
+   * on a background thread and posts the result. This runs the same steps for a new
+   * view synchronously, minus the parse: a fresh [Renderer] configured with the view's
+   * style renders the AST to spans, which become the view's text.
+   *
+   * Setting the text directly skips the view's link movement method and its image and
+   * table span registration; the documents contain no images or tables.
+   */
+  private fun createView(
```
hryhoriiK97, line 69: Follow-up idea, not for this PR. This is the only library-specific step, and the fixtures are already restricted to CommonMark that Markwon renders out of the box. Putting `parse` and `createView` behind a small `Subject` interface and parameterizing on `(subject, document)` would let us benchmark other libraries side by side, with `full` and `draw` comparable for any of them and `render`/`layout` for the spannable-based ones.

hryhoriiK97 (approved):
Thank you for the improvement, great job! Verified against the framework sources that hiding the builder type is the right fix for paragraph span order, and that re-setting an attached span keeps its slot. Unit tests pass, benchmark APK builds. Reproduced the benchmark on an emulator: same span counts and the same trend, with the largest gain in `layout` on complex_large.

packages/enriched-markdown-android/display-benchmark/src/androidTest/java/com/swmansion/enriched/markdown/benchmark/DisplayBenchmark.kt
jnowakow: It's good idea! I've added it to my To Do list ☑️

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/segments/RenderedSegment.kt
jnowakow: Added in 9e2bee6a01092921eab88f6d9deb1e077f00e139

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/utils/text/view/TextViewSetup.kt
jnowakow: Resolved in 9e2bee6a01092921eab88f6d9deb1e077f00e139

packages/enriched-markdown-android/ui/src/test/java/com/swmansion/enriched/markdown/TaskListInteractionTest.kt
jnowakow: That's right, fixed in 9e2bee6a01092921eab88f6d9deb1e077f00e139

packages/enriched-markdown-android/settings.gradle
jnowakow: Addressed in 9e2bee6a01092921eab88f6d9deb1e077f00e139

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/renderer/Renderer.kt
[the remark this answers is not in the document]
jnowakow: It will be released in new library version so the existing clients won't be affected. Moreover clients depend on `compose` - its API didn't changed.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/segments/RenderedSegment.kt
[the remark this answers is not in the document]
jnowakow: It will be released in new library version so the existing clients won't be affected. Moreover clients depend on `compose` - its API didn't changed.

eszlamczyk (approved):

PR #843, opened 2026-09-18
fix(android): clamp a11y items to laid-out text when truncated

### LINKED ISSUE #842: [Android] IndexOutOfBoundsException in MarkdownAccessibilityHelper when text is truncated (numberOfLines)
## Crash

When `EnrichedMarkdownText` is truncated via `numberOfLines`, the first accessibility traversal (TalkBack, or any service walking the node tree — e.g. Maestro) crashes the app:

```
java.lang.IndexOutOfBoundsException: offset(26) should be less than line limit(6)
	at android.text.TextLine.measure(TextLine.java:389)
	at android.text.Layout.getHorizontal(Layout.java:1246)
	at android.text.Layout.getHorizontal(Layout.java:1222)
	at android.text.Layout.getPrimaryHorizontal(Layout.java:1192)
	at android.text.Layout.getPrimaryHorizontal(Layout.java:1181)
	at com.swmansion.enriched.markdown.accessibility.MarkdownAccessibilityHelper.singleLineBounds(MarkdownAccessibilityHelper.kt:423)
	at com.swmansion.enriched.markdown.accessibility.MarkdownAccessibilityHelper.boundsForItem(MarkdownAccessibilityHelper.kt:410)
	at com.swmansion.enriched.markdown.accessibility.MarkdownAccessibilityHelper.onPopulateNodeForVirtualView(MarkdownAccessibilityHelper.kt:317)
```

## Repro

1. Render `EnrichedMarkdownText` with `numberOfLines={4}` (commonmark) and text long enough to be truncated.
2. Traverse the accessibility tree (enable TalkBack, or dump the hierarchy).
3. App crashes in `getPrimaryHorizontal()`.

The exact offset/limit vary with content; any summary longer than the visible lines triggers it.

## Root cause

`numberOfLines` applies `maxLines` + ellipsize to the layout, so the `Layout` holds fewer characters than the full text. But `MarkdownAccessibilityHelper.buildItems()` builds virtual nodes from the **full** text, and `boundsForItem()` passes those raw offsets to `layout.getPrimaryHorizontal()`, which throws for offsets past the laid-out line end. The helper predates `numberOfLines` support (#761) and was never updated for truncated layouts.

The standalone Android package (`packages/enriched-markdown-android`) has the same latent bug in its copy of the helper.

## Environment

- `react-native-enriched-markdown@1.1.0-nightly-20260912-4c1270f48` (also present on `main` and in `1.1.0-nightly-20260915`)
- Android API 33 emulator, also affects real devices with TalkBack

Happy to put up a PR clamping items to the laid-out range — just flagging here first per CONTRIBUTING.

ryanlntn:
### What/Why?

Fixes #842.

When the text is truncated (`numberOfLines` + ellipsize), the `Layout` holds fewer characters than the full text, but `MarkdownAccessibilityHelper.buildItems()` built virtual nodes from the full text. The first accessibility traversal then crashed in `boundsForItem()` → `getPrimaryHorizontal()` with `IndexOutOfBoundsException: offset(N) should be less than line limit(M)`.

This change caps the paragraph walk, semantic-span ranges, and the empty-fallback item at the last laid-out offset (`layout.getLineEnd(lineCount - 1)`), so only the visible portion is exposed, and defensively clamps offsets in `boundsForItem()` in case the layout changed since the last rebuild. Applied to both the React Native package and the standalone Android package, which share the same helper logic (standalone additionally skips admonition headers anchored past the visible text).

No behavior change when text isn't truncated: the visible length equals the full length.

### Testing

- Reproduced the crash downstream on an API 33 emulator (TalkBack-style traversal via Maestro): 3 crashes across e2e shards before the fix, 0 after, with the previously-failing flows green.
- Not ADDED upstream Maestro coverage: asserting a *lack* of crash during traversal would need a truncated-text + hierarchy-dump flow in the example app — happy to add one if maintainers want it.

### PR Checklist

- [ ] E2E tests are passing


packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
```diff
@@ -191,12 +205,13 @@ class MarkdownAccessibilityHelper(
...
     items: MutableList<AccessibilityItem>,
     spanned: Spanned,
+    endLimit: Int,
   ) {
     val layout = textView.layout ?: return
 
     for (header in spanned.getSpans(0, spanned.length, AdmonitionHeaderSpan::class.java)) {
       val start = spanned.getSpanStart(header)
-      if (start < 0) continue
+      if (start < 0 || start > endLimit) continue
```
hryhoriiK97: Should be `>=`. The paragraph walk uses `paraStart < endLimit`, so offset `endLimit` is not laid out. With `>` a header sitting exactly at `endLimit` still gets a node, positioned on the last visible line.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
```diff
@@ -169,9 +173,12 @@ class MarkdownAccessibilityHelper(
       paraStart = paraEnd
     }
 
-    addAdmonitionHeaderItems(result, spanned)
+    addAdmonitionHeaderItems(result, spanned, endLimit)
 
-    if (result.isEmpty()) return listOf(AccessibilityItem(0, text.trim(), 0, spanned.length))
+    if (result.isEmpty()) {
```
hryhoriiK97, line 179: This fallback can return an item with empty text. Blockquotes insert a leading `\n` spacer, so a doc starting with a quote under `numberOfLines=1` gives `endLimit == 1` and no paragraphs. The empty item then turns off host focus and TalkBack lands on an unlabeled node. Suggest `emptyList()` when the trimmed text is empty.

Same in the RN copy.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
```diff
@@ -181,6 +188,13 @@ class MarkdownAccessibilityHelper(
       .mapIndexed { index, item -> item.copy(id = index) }
   }
 
+  /** End offset of the laid-out (visible) text; less than the full length when truncated. */
+  private fun visibleTextLength(fullLength: Int): Int {
+    val layout = textView.layout ?: return fullLength
+    if (layout.lineCount == 0) return fullLength
+    return minOf(fullLength, maxOf(0, layout.getLineEnd(layout.lineCount - 1)))
```
hryhoriiK97: With a tail ellipsis, StaticLayout merges the overflowed lines into the last line, so `getLineEnd(last)` covers text hidden under the ellipsis. Links there stay focusable and clickable. Cap at `getLineStart(last) + getEllipsisStart(last)` when `getEllipsisCount(last) > 0`.

Same in the RN copy.

hryhoriiK97: Truncation is only detected when the layout ellipsizes. With `clip`, or head/middle over multiple lines, all lines are laid out and only the view height is capped, so `endLimit` is the full length. Also clamp the line index to `min(lineCount, textView.maxLines) - 1`. A shared helper with `boundsForItem` would keep both call sites in sync.

Same in the RN copy.

hryhoriiK97 (changes requested):
@ryanlntn, thanks for the fix! I've left a few comments.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
ryanlntn: Good catch — you're right, `endLimit` itself isn't laid out. I'll switch this to `>=` to match the `paraStart < endLimit` walk.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
ryanlntn: Great edge case, thanks — a doc starting with a blockquote spacer under `numberOfLines=1` would indeed produce an empty item and steal host focus. I'll return `emptyList()` when the trimmed visible text is empty, in both copies.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
ryanlntn: Ah right — `getLineEnd(last)` includes text hidden under the ellipsis. I'll cap at `getLineStart(last) + getEllipsisStart(last)` when `getEllipsisCount(last) > 0`, in both copies.

ryanlntn: Good point — `clip` (and head/middle over multiple lines) won't shrink the layout, so `endLimit` would stay full length. I'll also clamp the last line to `min(lineCount, textView.maxLines) - 1` and share the helper with `boundsForItem()` so they stay in sync. Both copies.

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
```diff
@@ -181,6 +190,27 @@ class MarkdownAccessibilityHelper(
...
   }
 
+  /** Last line that is visible; less than lineCount - 1 when height-capped (e.g. `clip`). */
+  private fun lastVisibleLineIndex(layout: Layout): Int {
+    val maxLines = textView.maxLines.coerceAtLeast(1)
+    return (minOf(layout.lineCount, maxLines) - 1).coerceIn(0, layout.lineCount - 1)
+  }
+
+  /** Ellipsis-aware end offset for a visible line. */
+  private fun visibleLineEnd(layout: Layout, line: Int): Int =
```
hryhoriiK97, line 200: @ryanlntn thank you for addressing the comments! I found another issue with the ellipsis cap, and the lint failure is a small formatting nit.

**Head and middle truncation drop every visible node**

`visibleLineEnd` caps at `getLineStart(line) + getEllipsisStart(line)` whenever the line is ellipsized. That's only right for tail. Per `StaticLayout.calculateEllipsis`, `START` sets `ellipsisStart = 0` with the hidden prefix as the count, and `MIDDLE` sets `ellipsisStart` to the left cut. Both modes are reachable through `ellipsizeMode` ("head" / "middle") with `numberOfLines={1}`.

I ran the PR head under Robolectric native graphics (SDK 35): one paragraph ending in a link, `maxLines = 1`, 300px wide.

| mode | layout ellipsis (start, count) | nodes exposed |
|---|---|---|
| tail | 32, 72 | 1 node, visible prefix only ✅ |
| clip | none, 4 lines | 1 node, first line ✅ |
| head | 0, 69 | 0 nodes, host becomes focusable, link lost ❌ |
| middle | 16, 68 | 1 node with the first 16 chars, link lost ❌ |

The hidden range only reaches the end of the line for tail, so we can key off that instead of the mode:

```kotlin
/** Ellipsis-aware end offset for a visible line. */
private fun visibleLineEnd(
  layout: Layout,
  line: Int,
): Int {
  val lineEnd = layout.getLineEnd(line)
  val count = layout.getEllipsisCount(line)
  if (count == 0) return lineEnd
  val hiddenStart = layout.getLineStart(line) + layout.getEllipsisStart(line)
  return if (hiddenStart + count >= lineEnd) hiddenStart else lineEnd
}
```

With this, head and middle expose the paragraph and the link again, tail and clip are unchanged. The head/middle paragraph node still includes the hidden text, same as on `main`, which seems like the right tradeoff for now. Same in the RN copy.

**`android-kotlin-lint` failure**

The two-parameter one-liner `visibleLineEnd(layout: Layout, line: Int)` trips `standard:function-signature`, which forces one parameter per line at two or more. The snippet above is already in that shape and passes `ktlintCheck` locally. Worth reformatting the RN copy the same way even though it isn't linted.

hryhoriiK97 (changes requested):

packages/enriched-markdown-android/ui/src/main/java/com/swmansion/enriched/markdown/accessibility/MarkdownAccessibilityHelper.kt
ryanlntn: Thanks for running this through Robolectric — that's a clean repro. Adopted your `visibleLineEnd` as-is (keying off whether the hidden range reaches the line end, so head/middle keep the paragraph and link while tail/clip are unchanged) in both copies, and reformatted to one param per line. `:ui:ktlintCheck` passes locally. Pushed in 835bdbac.

hryhoriiK97 (approved):
Thanks for the PR, @ryanlntn ! Verified all the changes, LGTM 🚀 Your fix will be included in tomorrow's nightly release 😃

